PR #11, opened 2016-10-15
Add support for downloading and reading the diff from the github API

orta:
So I had to add a new dependency, I know, I know, but I'm not sure that reading diffs is the domain of   Danger. 

However, this will grab a PR, and pull out the `modified_files`, `added_file` directly from the provided diff in the API - in theory, I could get a working build of Danger with a bit of work now.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -16,7 +16,7 @@
     "quotes": [1, "double", "avoid-escape"],
     "flowtype/define-flow-type": 1,
     "flowtype/require-parameter-type": 1,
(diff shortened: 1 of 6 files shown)


.eslintrc.json
```diff
@@ -16,7 +16,7 @@
     "quotes": [1, "double", "avoid-escape"],
     "flowtype/define-flow-type": 1,
     "flowtype/require-parameter-type": 1,
-    "flowtype/require-return-type": [ 1,"always", { "annotateUndefined": "never" } ],
+    "flowtype/require-return-type": [ 0,"always", { "annotateUndefined": "never" } ],
```
orta: this gets really annoying

PR #14, opened 2016-10-16
Initial work on the CHANGELOG test, and adds a fail command

orta:
In theory, build should fail - changelog is in the new files - not modified. If it fails e should be good to go


orta:
screen shot 2016-10-16 at 18 00 13

PR #16, opened 2016-10-26
introduce the eslint `prefer-const` rule

romankl:
Some files already use `const`, but some use `let` nearly exclusive. Eslint offers the `prefer const` which detects possible constant values. This PR introduces the rule and fixes all reported violations.

I've created two separated commits for:
- introducing the rule
- fixing the issues

Should I squash them together?

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,6 +21,10 @@
             "ClassDeclaration": false
         }
     }],
+    "prefer-const": ["error", {
+        "destructuring": "any",
+        "ignoreReadBeforeAssign": false
+    }],
     "space-before-function-paren": ["error", "never"],
     "quotes": [1, "double", "avoid-escape"],
     "flowtype/define-flow-type": 1,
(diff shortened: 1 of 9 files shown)


.eslintrc.json
```diff
@@ -21,6 +21,10 @@
             "ClassDeclaration": false
         }
     }],
+    "prefer-const": ["error", {
+        "destructuring": "any",
+        "ignoreReadBeforeAssign": false
+    }],
```
orta: yep definitely, this comes up all the time in code review for me https://github.com/artsy/emission/pull/380#pullrequestreview-5433438

orta:
Hah - it was actually danger failing - "No Changelog changes!" 

We still don't have any feedback yet, was planning on starting to look at that today. I'm OK with this being red for now 👍 

I also don't care about git history - so this is great

orta:
Thanks @romankl - would you like to join the Danger org? we conform to the [Moya Community Continuity](https://github.com/Moya/contributors/issues/6) - you're welcome to participate in PRs and help out at a level that works for you.

romankl:
@orta hey, whoaa that would be an honor - thank you! 👍  (right now it was just a small PR)

orta:
sent - no problem, gives you more freedom, and eventually we'll add a Danger rule to ask, https://github.com/danger/danger/blob/master/Dangerfile#L14-L22

Issue #23, opened 2016-10-27
Do not create duplicate messages when failing


orta:
Currently for every fail there will be a new comment. We should edit an existing message, so that you don't get a lot of emails.

orta:
Done in #27

PR #27, opened 2016-10-27
Update or delete comments

orta:
Should address #23


orta:
Awww yeah

That's the one. Went through a full cycle, new post -> editing -> deleting.

PR #28, opened 2016-10-28
Update README for Early Adopters

orta:
I'm about to recommend it as being usable in prod.


orta:
Hrm - things aren't looking good :D

https://github.com/artsy/emission/pull/385

PR #30, opened 2016-10-30
Add some test structure for the github class + flow typed definition

orta:
Wanted to see what it looks like to add some tests to check the resolves URLs for the GitHub API stuff 👍


orta:
The flow typed definition is now being built from the real code 👍

orta:
looks like the export file needs to go somewhere outside of `source/`

orta:
Alright, lets get this in.

PR #36, opened 2016-11-23
Allows fakeci to run via DANGER_TEST_PR and DANGER_TEST_REPO env vars

nsfmc:
This is a pretty good way to test danger and also a lightweight way to
evaluate if "danger is right for you™". Example in README gives a pretty
good overview.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ Danger on Node, wonder what's going on? see [VISION.md](VISION.md)
 
 *Welcome!*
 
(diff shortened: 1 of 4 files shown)


nsfmc:
i'm not sure how i feel about the env var names or if this behavior should be spun off into its own CI, this felt like the right spot, though. happy to hear any feedback

orta:
This feels good to me for now, ideally this will need moving into a `danger local` command which can fake it all appropriately ( which is how we do it in Ruby ) but I have no problem with this 👍

orta:
I should see what's up with Circle

orta:
Yeah. Let's get this in 👍

PR #37, opened 2016-11-24
emit console error if `DANGER_GITHUB_API_TOKEN` is not set.

nsfmc:
Without this, danger will appear to fail because it is unable to
execute `.find()` within the PR's comments, this at least 
explains the root cause.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,13 +2,15 @@
 
 //  Add your own contribution below
 
+* Improved error messaging around not including a `DANGER_GITHUB_API_TOKEN` in the ENV - nsfmc / orta
+
 ### 0.6.3
 
 * Does not break commonmark on GitHub - orta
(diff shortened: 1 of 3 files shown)


nsfmc:
i'm curious whether you think this should `throw` the error and stop `danger` before it can continue any further.

orta:
IMO,  throw - can't do anything

orta:
Writing tests is so dreamy with my new VS Code setup 💟 

screen shot 2016-11-28 at 11 13 55

OK - this looks obvious to me what the problem is, let's get this in

screen shot 2016-11-28 at 11 17 47

orta:
Alright - we're green

nsfmc:
rad, sorry for the late replies here, had a pretty busy weekend.

PR #41, opened 2016-12-14
Adds support for shipping the danger annotations during publishing

orta:
This _should_ provide typings for danger, without flow-install, I believe.

It's based on the technique used by [GraphQL team](https://github.com/graphql/graphql-js/blob/master/package.json#L31) which is to just ship their actual source (with normal flow annotations) alongside the babel derived sources for consuming in node. Clever.

So in this case, instead of the file generated to ship to flow-typed  which typically looks like

screen shot 2016-12-14 at 9 23 51 am

it is passed along into the npm module.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,10 @@
 
 //  Add your own contribution below
 
+### 0.6.6
+
+* Ship flow annotations with the npm module - orta
+
 ### 0.6.5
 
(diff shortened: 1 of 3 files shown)


orta:
Also added [in-publish](https://github.com/iarna/in-publish) to make the build process for deployment only happens during deployment.

PR #57, opened 2016-12-22
Version bump

orta:
With all the Jest under the hood changes


orta:
weird, I 
can't leave a reaction on that comment

PR #59, opened 2016-12-27
Set exit code to 1 when error running Dangerfile

### LINKED ISSUE #50: Broken Dangerfile did not fail the CI
I vote this was not "All Good."

screen shot 2016-12-14 at 5 11 30 pm

macklinu:
Resolves #50 

@orta I wasn't sure how to test this. `danger-run.js` is more of a script than something that can be imported and unit tested individually. Any thoughts on adding integration tests for a scenario like this? Not sure what that would look like (maybe use something like [bats](https://github.com/sstephenson/bats)). Might also be overkill here and wondering what your thoughts are.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,8 @@
 
 //  Add your own contribution below
 
+* Set exit code to 1 when running `danger` throws an error - macklinu
+
 ### 0.7.0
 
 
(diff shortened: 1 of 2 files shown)


orta:
Yeah, this one is tough. 

We found this close to [unfeasible](https://github.com/danger/danger/blob/master/spec/lib/danger/commands/runner_spec.rb) on Danger Ruby. 

I think the best strategy for this kind of thing is to make this script configure an object then call it (which is what the executor is for) and try to have as little logic in here as possible

orta:
So I'm alright with not having tests

PR #60, opened 2016-12-27
Add .editorconfig

macklinu:
http://editorconfig.org/

I found during development that ESLint would want me to remove trailing whitespace and use a 2 space indent, and I'm used to an `.editorconfig` helping enforce that when a file is saved (that way you don't have to think about it). Thought I'd bring it in here to help support the code style across IDEs (I'm trying out VSCode for now, but EditorConfig is supported by many IDEs). 📔

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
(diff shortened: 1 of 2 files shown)


orta:
Sure thing 👍

PR #61, opened 2016-12-27
Add Jenkins CI source

### LINKED ISSUE #56: Add Jenkins CI source
The [Jenkins CI source](https://github.com/danger/danger/blob/master/lib/danger/ci_source/jenkins.rb) in Danger.rb, for reference

macklinu:
Resolves #56

### DIFF
--- a/source/ci_source/Jenkins.js
+++ b/source/ci_source/Jenkins.js
@@ -1,0 +1,39 @@
+// @flow
+"use strict"
+
+import type { Env } from "./ci_source"
+import { ensureEnvKeysExist, ensureEnvKeysAreInt } from "./ci_source_helpers"
+
+export default class Jenkins {
+  env: Env
+
+  constructor(env: Env) {
+    this.env = env
+  }
+
+  get name(): string {
+    return "Jenkins"
+  }
+
+  get isCI(): boolean {
+    return ensureEnvKeysExist(this.env, ["JENKINS_URL"])
+  }
+
+  get isPR(): boolean {
+    const mustHave = ["JENKINS_URL", "ghprbPullId", "ghprbGhRepository"]
+    const mustBeInts = ["ghprbPullId"]
+    return ensureEnvKeysExist(this.env, mustHave) && ensureEnvKeysAreInt(this.env, mustBeInts)
+  }
+
+  get pullRequestID(): string {
+    return this.env.ghprbPullId
+  }
+
+  get repoSlug(): string {
+    return this.env.ghprbGhRepository
+  }
+
+  get supportedPlatforms(): string[] {
+    return ["github"]
+  }
+}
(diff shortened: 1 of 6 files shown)


source/ci_source/Jenkins.js
```diff
@@ -0,0 +1,39 @@
...
+  constructor(env: Env) {
+    this.env = env
+  }
+
+  get name(): string {
+    return "Jenkins"
+  }
+
+  get isCI(): boolean {
+    return ensureEnvKeysExist(this.env, ["JENKINS_URL"])
```
macklinu: Based on https://github.com/danger/danger/blob/v4.0.3/lib/danger/ci_source/jenkins.rb#L42

source/ci_source/ci_source.js
```diff
@@ -26,14 +26,12 @@ export interface CISource {
 
     /** What unique id can be found for the code review platform's PR */
     +pullRequestID: string,
-
-    /** What is the project name */
-    +name: string
```
macklinu: Dupe, removed.

source/ci_source/_tests/_travis.test.js
```diff
@@ -39,7 +39,7 @@ describe(".isPR", () => {
     env[key] = null
 
     test(`does not validate when ${key} is missing`, () => {
-      const travis = new Travis({})
+      const travis = new Travis(env)
```
macklinu: The Travis CI source test wasn't validating the `env` where `env[key]` was null. Updated this test to do so (which I think it intends to do).

source/ci_source/ci_source_helpers.js
```diff
@@ -11,7 +11,7 @@ import type { Env } from "./ci_source"
 */
 export function ensureEnvKeysExist(env: Env, keys: string[]): boolean {
   const hasKeys = keys.map((key: string): boolean => {
-    return env.hasOwnProperty(key) && env[key].length > 0
+    return env.hasOwnProperty(key) && env[key] != null && env[key].length > 0
```
macklinu: The Travis (and now Jenkins) CI source tests set an `env[key]` to null in the tests (our way of saying the key exists but there is no value for that key). This function was attempting to access `.length` on null, which throws a `TypeError`, so I believe this null/undefined check is useful here in case the `env` object has a supplied property but its value is falsy.

source/ci_source/_tests/_travis.test.js
orta: 👍

source/ci_source/ci_source.js
orta: nice one

source/ci_source/ci_source_helpers.js
orta: Yep 👍

orta:
Will merge on 🍏

PR #66, opened 2016-12-30
Enhance windows OS compatibility

kwonoj:
Hi 👋 

I was trying to suggest proposal on codebase, while trying those came to think it'd be better to separate out effort to enhance Windows OS compatibility first.

This PR includes few changes related to those,
- unit test fixture setup / assertions are now os independent
- adding some default values for configuration of `jest-runtime` to avoid specific behavior of jest runtime, specifically jest internally picks up crawl function based on os, and it doesn't resolve at all and causing danger runner stall when missing some of default config values. 

I have verified windows side locally, and *nix regressions are verified via https://travis-ci.org/kwonoj/danger-js/builds/187634353.


source/platforms/_tests/GitHub.test.js
```diff
@@ -81,10 +83,11 @@ describe("with fixtured data", () => {
       expect(gitDSL.deleted_files).toEqual(["lib/components/artist/related_artists/index.js", "lib/components/artist/related_artists/related_artist.js", "lib/components/gene/about_gene.js"])
     })
 
-    // sorry windows users - I guess this test will fail for you
     it("shows the diff for a specific file", async () => {
```
orta: :D

orta:
brilliant - for danger-rb we run tests on appveyor to ensure we don't break windows support. Let's get this is, and I'll add support for it now! Thanks

orta:
also, I've invited you to the org, as a part of [Moya Community Continuity](https://github.com/Moya/contributors/issues/6) - you're welcome to participate in PRs and help out at a level that works for you too

kwonoj:
@orta appreciate it!

PR #67, opened 2016-12-30
Add appveyor file

orta:
Thanks @kwonoj for doing the ground work here - this should try and ensure we don't break windows support in the future :1st_place_medal:


orta:
👌

PR #69, opened 2016-12-30
Remove danger imports in dangerfiles

orta:
When you include the `import { danger, warn, fail, message } from 'danger'` for Flow typings, then we need to make sure it gets wiped out. Will also release with this.

### DIFF
--- a/source/dsl/GitHubDSL.js
+++ b/source/dsl/GitHubDSL.js
@@ -187,6 +187,60 @@ export interface GitHubPRDSL {
    * The User who submitted the PR
    * @type {GitHubUser}
    */
-  user: GitHubUser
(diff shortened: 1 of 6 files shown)


source/dsl/GitHubDSL.js
```diff
@@ -187,6 +187,60 @@ export interface GitHubPRDSL {
...
+   * The number of deleted lines in the PR
+   * @type {number}
+   */
+  deletions: number,
+
+  /**
+   * The number of changed files in the PR
+   * @type {number}
+   */
+  changed_files: number,
```
orta: used in https://github.com/styled-components/styled-components/pull/169

PR #71, opened 2016-12-31
Allow to specify external CI source provider

kwonoj:
This PR adds new feature into `danger run` command via `-c` option, allow specify external ci source provider not available in default providers. There are numbers of CI servers yet included in package (i.e teamcity), also some highly-customizable CI like `surf` (https://github.com/surf-build/surf) may require own level of source detection mechanism. 

In addition to those, added initial typescript definitions for `CISource` interface. Personally prefer to write whole codebase as TS instead of manually writing ts definition, but at least this'll allow any TS consumer can import interfaces.


source/ci_source/ci_source.js
```diff
@@ -28,34 +28,61 @@ export interface CISource {
...
-    return circle
-  } else if (semaphore.isCI) {
-    return semaphore
-  } else if (jenkins.isCI) {
-    return jenkins
-  } else if (fake.isCI) {
-    return fake
+  const availableProviders = [...providers]
+    .map(Provider => new Provider(env))
+    .filter(x => x.isCI)
```
orta: 👏 - I knew there was something elegant like this. Nice work

orta:
With respect to porting to TypeScript, I'm totally fine with doing so ( [Peril](https://github.com/danger/peril) uses TS for example. )

PR #54, opened 2016-12-20
Initial proof of concept for jest-runtime based Dangerfile parsing

orta:
Uses jest-runtime to power the babel parsing and module separation for running a Dangerfile.

TODO:

 - [x] tests
 - [x] check out that relative imports work fine
 - [x] ensure type annotations are set up right for working with jest-runtime

PR #73, opened 2016-12-31
Add some debug info to the Dangerfile 

orta:
when you include the word verbose in the message body.

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -1,6 +1,6 @@
 // @flow
 
-// import { danger, warn } from "danger"
+import { danger, warn, markdown } from "danger"
 import fs from "fs"
 
 // Request a CHANGELOG entry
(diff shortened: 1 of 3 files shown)


dangerfile.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-// import { danger, warn } from "danger"
+import { danger, warn, markdown } from "danger"
```
macklinu: This was commented out, but now it's not. I remember in #54 (https://github.com/danger/danger-js/commit/0ba01b15e56093c5ce11ecccccef738b3ef66cf0#diff-3abee5fb96982a8bed436d8b59267dd8R76) that the danger DSL was added to the global namespace for convenience, at least as I understood it. Is this still the case?

orta: It is the case, yes, but its fakey - this is in there for tooling -> https://github.com/danger/danger-js/blob/master/source/runner/DangerfileRunner.js#L90

This means that flow/TS will give you the right autocompletions + documentation. The runner will just comment it out before evaling the DAngerfile

orta:
Had a debate about having this be a part of danger, which can be triggered by a body message instead  of just adding it to this specific Dangerfile - think I'm gonna do that

dangerfile.js
macklinu: Ah makes sense regarding the Flow/TS types. 👍  thanks

macklinu:
@orta what are next steps for this PR? Are you suggesting adding this as a feature of Danger-js itself?

orta:
Yeah, I was debating this as something which you could say inside the GitHub PR body like `[danger debug]` which would output the above, wanted to keep it open till I or someone else made up their mind about it 💃

orta:
OK, I'm closing this. Happy for someone to pick up but TBH I think `danger pr` is a better call

PR #74, opened 2017-01-02
correct example syntax in VISION.md

bgoscinski:


orta:
Great - thanks!

I did this before a Dangerfile would be transpiled in babel, a modern one can be much more elegant now 🎉 - will try and give an update to that tomorrow

Issue #18, opened 2016-10-27
Investigate an easy way to execute commandline tools


orta:
I'll want something that takes flow & ESLint's JSON output and allows us to show errors for that. In ruby this is as simple as 

``` ruby
`echo "OK"`
```

I don't know what that should look like in JS, perhaps we'll need something like?

``` ruby
danger.run("flow check --json", output: "json", (result: any) => {
  if (!result.passed) {
    fail("Boo! Better fix them types")
  }
})
```

Given the prevalence of JSON output, we can optionally handle the transition from text for users too

orta:
This will NO-OP when running inside hosted Danger

orta:
OK, I"ve one back to this. I've found a good idea on what to base it on, [VS Code's problem matchers](https://code.visualstudio.com/Docs/editor/tasks#_processing-task-output-with-problem-matchers)

It's support an arg like:

```js
{
    // The problem is owned by the cpp language service.
    "owner": "cpp",
    // The file name for reported problems is relative to the opened folder.
    "fileLocation": ["relative", "${workspaceRoot}"],
    // The actual pattern to match problems in the output.
    "pattern": {
        // The regular expression. Example to match: helloWorld.c:5:3: warning: implicit declaration of function ‘prinft’ [-Wimplicit-function-declaration]
        "regexp": "^(.*):(\\d+):(\\d+):\\s+(warning|error):\\s+(.*)$",
        // The first match group matches the file name which is relative.
        "file": 1,
        // The second match group matches the line on which the problem occurred.
        "line": 2,
        // The third match group matches the column at which the problem occurred.
        "column": 3,
        // The fourth match group matches the problem's severity. Can be ignored. Then all problems are captured as errors.
        "severity": 4,
        // The fifth match group matches the message.
        "message": 5
    }
}
```

``` js
danger.run({
  command: "eslint",
  errorPatterns: [
        {
            regexp: "^([^\\s].*)$",
            file: 1
        },
        {
            regexp": "^\\s+(\\d+):(\\d+)\\s+(error|warning|info)\\s+(.*)\\s\\s+(.*)$",
            line: 1,
            column: 2,
            severity: 3,
            message: 4,
            code: 5,
            loop: true
        }
    ]
})
```

We may be able to take the code directly from vscode, don't think they separate it into modules much so it'd need a license attaching and bringing inline but it seems like the right abstraction.

orta:
[...]

PR #78, opened 2017-01-04
Migrate codebase into TypeScript from Flowtype

### LINKED ISSUE #76: Port codebase to TypeScript
Via https://github.com/danger/danger-js/pull/71#issuecomment-269869256

Is this something we'd like to move forward with? I personally enjoy writing TS. Haven't converted Flow to TS before, so I'm not sure if it's something we could do incrementally or maybe doing it all in one shot isn't a difficult task. 

Any thoughts? /cc @orta @kwonoj

kwonoj:
Apologize for huge PR. I was initially thought gradual migration via `allowJS` would work, turned out it was not possible since pretty much most of codes are having flow based typedef, TS compiler can't consume it as is.

This PR updates codebase into TypeScript from flowtype, includes necessary build pipeline alignment as well. Main focus for PR is build sanity, tried to avoid add new changes as much as possible only except necessary one. Most of compiler config / lint configs are inspired by my daily work.

- closes #76.


orta:
This is great, I was just working on migrating in the vscode problem matcher https://github.com/danger/danger-js/issues/18 - which meant porting a bunch of TS to JS+Flow. 

Will clone this locally and dig around a bit. 👍

macklinu:
I poked around at converting last night and came across the same issue @kwonoj, that `allowJs` wouldn't work because of Flow. I'll give this a run through tonight/tomorrow - thanks for getting this going!

orta:
hah, I also love that this means I'll also have to migrate this upcoming PR to tslint too - https://github.com/eslint/eslint/issues/7812 :D

orta:
I only care about two lint-y things:

* double quotes instead of single, I try keep it consistent among all projects as I touch so many languages ( and it's the [default for TS](https://github.com/alloy/relational-theory/pull/12#issuecomment-270187669) )

* I'd prefer no semi-colons, I think it's rare in modern JS/TS that you actually need it and I'd rather occasionally hit whatever edge-cases there are than have to remember to do something I don't need to do

do you mind if we change those settings?

kwonoj:
@orta no worries, it'll be easy - let me update quickly.

source/dsl/DangerDSL.ts
```diff
@@ -14,23 +11,21 @@ export interface DangerDSLType {
    *  Currently, this is just the raw file paths that have been
    *  added, removed or modified.
    */
-  git: GitDSL;
+  readonly git: Readonly<GitDSL>;
```
orta: this is really cool

kwonoj:
@orta done, updated codes & lint config both.

source/ambient.d.ts
```diff
@@ -0,0 +1,8 @@
+declare module "node-fetch";
+declare module "parse-diff";
+declare module "lodash.includes";
+declare module "lodash.find";
+declare module "jest-runtime";
+declare module "jest-environment-node";
+
+declare module "*/package.json";
```
orta: what does this do?

source/runner/Executor.ts
```diff
@@ -1,23 +1,16 @@
...
-import githubResultsTemplate from "./templates/github-issue-template"
+import { DangerResults } from "../dsl/DangerResults"
+import { template as githubResultsTemplate } from "./templates/github-issue-template"
 import { createDangerfileRuntimeEnvironment, runDangerfileEnvironment } from "./DangerfileRunner"
-import type { DangerfileRuntimeEnv } from "./types"
+import { DangerfileRuntimeEnv } from "./types"
 
 // This is still badly named, maybe it really should just be runner?
 
-export default class Executor {
```
orta: is `export default` something that is done less in TS? I've only worked in vscode's source, and wrote my own projects from scratch - so I don't have much ambient cultural knowledge

source/ambient.d.ts
kwonoj: it is ambient declaration for module does not have any type definition. By default TS compiler without allow impicit any does not accept to import plain CJS module without corresponding type definition (either `@types` or package-provided typings). This ambient declaration makes all those CJS modules `any` type, can be imported where it doesn't have type definition still.

source/runner/Executor.ts
kwonoj: So, it's bit complicated issue. TS compiler does not interop `default` export, so if you export something like

`export default aaa` in a.ts

then you have to import like
```
import xxx from '...a'

xxx.default //have to explicitly specify 'default' namespace since TS compiler does not interop
```

while babel does. So it is typically recommended to use named export, unless it is expected to use explicit `x.default`everywhere.

orta:
@macklinu I'll leave it for you to give it a one over also, then if you're happy hit the merge 👍 :1st_place_medal: 

This is awesome work @kwonoj - totally awesome

source/ambient.d.ts
alloy: Ooh, does this mean one doesn’t have to change the syntax used like so? https://github.com/alloy/relational-theory/pull/16/files#diff-e4650f225ab079757ad601f52c84284fR8

kwonoj: @alloy yes, I usually prefer to avoid `require` to escape hatch for non-typed package as much as I can. Nearly most cases it won't be needed.

kwonoj: https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#wildcard-character-in-module-names

alloy: @kwonoj Not sure I follow 100%, do you still have to specify the `json!` hint?

kwonoj: see what I did in line 8, which covered by TS behavior above link. Point is you can specify ambient declaration with wildcard, *and add necessary hint if you want*. so

`declare module "*/package.json"` -> `import * as pkg 'package.json' //works`

such as https://github.com/danger/danger-js/pull/78/files#diff-4b47fd8c4ebfbe9a6c8992df39f6f64dR4 .

alloy: Got it. Sweet, thanks a bunch, @kwonoj 👍

kwonoj:
Note coverage report goes bit off, may need https://github.com/facebook/jest/pull/2290 to be included in next `jest` version bump.

PR #79, opened 2017-01-05
Remove flow from CircleCI build config #trivial

macklinu:
Should resolve the failing build in https://circleci.com/gh/danger/danger-js/113 after merging #78.

Questions:

- Does a change like this require a Changelog entry?
- Do we want to run the tests on CircleCI as well (before it was just the lint and flow npm tasks)

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -1,15 +1,16 @@
 // import { danger, warn } from "danger"
 import fs from "fs"
 
-// Request a CHANGELOG entry
+// Request a CHANGELOG entry if not declared #trivial
 const hasChangelog = danger.git.modified_files.includes("changelog.md")
-if (!hasChangelog) { warn("Please add a changelog entry for your changes.") }
+const isTrivial = (danger.github.pr.body + danger.github.pr.title).includes("#trivial")
+if (!hasChangelog && !isTrivial) {
+  warn("Please add a changelog entry for your changes.")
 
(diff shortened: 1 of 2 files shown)


orta:
IMO Changelog entries are for danger consumers, they don't care that we did this, but they do probably care if we move to TS because it might be the cause of a bug. In RB we allow adding the word "trivial" to the github pr title or body to skip the check

WRT Tests on circle, we should probably do that. On RB we did it to ensure that platform differences worked out fine (Artsy uses Circle on all mobile projects ) between travis and circle ( Circle was the most complex integration )

macklinu:
Cool, I'll push up

- adding that #trivial check to the Dangerfile
- an update to run the tests on CircleCI

dangerfile.js
```diff
@@ -1,15 +1,16 @@
...
-  warn("Please add your GitHub name to the changelog entry, so we can attribute you.")
+  // Politely ask for their name on the entry too
+  const changelogDiff = danger.git.diffForFile("changelog.md")
+  const contributorName = danger.github.pr.user.login
+  if (changelogDiff && changelogDiff.indexOf(contributorName) === -1) {
+    warn("Please add your GitHub name to the changelog entry, so we can attribute you.")
+  }
 }
-
-const jsFiles = danger.git.created_files.filter(path => path.endsWith("js"))
```
orta: 👍

PR #82, opened 2017-01-05
Remove all types of imports/requires

orta:
I forgot that `require` imports even exist, so when I was creating the Dangerfile for Jest I realized I needed to add this in 👍 

--

Note: Jest is currently infinitely looping on `yarn run jest -- --watch` - I assume it's the code coverage stuff that is triggering a new run, will try dig into it tomorrow.

### DIFF
--- a/source/runner/_tests/DangerRunner.test.ts
+++ b/source/runner/_tests/DangerRunner.test.ts
@@ -83,11 +83,40 @@ describe("cleaning Dangerfiles", () => {
     const path = resolve(os.tmpdir(), "fake_dangerfile_1")
     fs.writeFileSync(path, "import { danger, warn, fail, message } from 'danger'")
     updateDangerfile(path)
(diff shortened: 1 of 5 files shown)


source/runner/_tests/DangerRunner.test.ts
```diff
@@ -83,11 +83,40 @@ describe("cleaning Dangerfiles", () => {
...
+// Removed import
+// Removed import
+// Removed import
+`
+    expect(cleanDangerfile(before)).toEqual(after)
+  })
+
+  it("also handles require style imports", () => {
+        const before = `
+const { danger, warn, fail, message } from 'danger'
```
macklinu: I'm not sure if this is valid syntax/tests what your `it()` description states. The `require` imports would look like:

```js
const { danger, warn, fail, message } = require('danger')
```

orta: Nice one, good code reviewing!

PR #87, opened 2017-01-06
fix(getCISourceForExternal): supply env variable to constructor

kwonoj:
This PR deliver env variable to external module as well, to support cases external module want to consume env. variables.


orta:
Makes sense, can you add a changelog please?

orta:
Awesome, looks good to me :1st_place_medal:

PR #90, opened 2017-01-07
refactor(Executor): clarify result message status

kwonoj:
This PR adds small enhancement to console output, between failure message and non-critical failure messages.


orta:
Yes! This was on my TODO for the weekend! Nice one, I kept forgetting to write an issue for this.

PR #91, opened 2017-01-07
feat(debug): debug module for dump status

kwonoj:
This PR allows to debug dump if needed, by setting `DEBUG` environment. As this PR uses prefix `danger:`, running 

`DEBUG=danger:* danger` 

will output all debug dump specified.

### DIFF
--- a/source/commands/danger.ts
+++ b/source/commands/danger.ts
@@ -3,9 +3,13 @@
 // import app from "./app"
 import { version } from "../../package.json"
 import * as program from "commander"
+import * as debug from "debug"
+
+const d = debug("danger:runner")
+
+d(`argv: ${process.argv}`)
 
 // Provides the root node to the command-line architecture
-
 program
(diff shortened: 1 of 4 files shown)


source/commands/danger.ts
```diff
@@ -3,9 +3,12 @@
 // import app from "./app"
 import { version } from "../../package.json"
 import * as program from "commander"
+import * as debug from "debug"
+const d = debug("danger:runn")
```
orta: is "runn"  a typo?

orta:
This one is worth a CHANGELOG saying what it does and how to use it yourself - other than that - I' m a 👍

source/commands/danger.ts
kwonoj: oh yes. I just did for verification and haven't updated namespace correctly.

orta:
👍

Issue #88, opened 2017-01-07
Support async code in dangerfile


kwonoj:
How does async code works in dangerfile? I gave quick try like

```js
const fs = require("fs");
const os = require("os");
const summary = require("lcov-summary");

const file = fs.readFileSync("./coverage/lcov.info", "utf-8");

summary(file, (err, results) => {
  console.log('here');
  if (err) {
    warn("cannot retrieve code coverage");
  } else {
    console.log('there');
    const total = results
      .split(os.EOL)
      .filter((result) => result.match(/Total Coverage/))[0];
    message(total);
    warn(total);
    fail('fail');
  }
});
```

to write down simple message inside of callback. Callback's executed, but none of danger messages are written down. is this expected?

orta:
Not expected, this should work - I've not touched any async code yet in my own dangerfiles. However, I would have thought that we would have got a notification of when the vm script finished, which would include all potential async callbacks wrapping up. 

Maybe that assumption is wrong

kwonoj:
So, for this issue.. reason was quite simple, but I couldn't make a feasible solution to propose with current public interfaces.

What's happening inside of runner of danger imports dangerfile (https://github.com/danger/danger-js/blob/5ddce5937252bf7d915cb2cb40484e3ca7788444/source/runner/DangerfileRunner.ts#L73) and that's just it as whole execution and rely on all eval is done when module is being imported. It is naïvely working model, since when CJS module is being imported it'll be evaluated immediately (and statically), so all of synchronous danger execution will be evaluated as well. 

This becomes problematic with async code which starts evaluation but runner does not able to wait those async code to be completed. Runner may need to deliver some sort of way to get signal, and dangerfile need to explicitly signal once it's done.

orta:
Ouch, that's not optimal!

I took a look around to see if I could find any existing implementations that handle this problem, and there are two:

* https://github.com/dallonf/async-eval
* and https://github.com/brianmcd/contextify (which comes with it's own "you do not need this dep" warning)

async-eval looks like it actually does what you're talking about by wrapping the eval'd code. 

Another potential place where we could find answers is inside [vm2](https://www.npmjs.com/package/vm2), which is a lib I considered instead of jest-runtime - their example uses an async script

_sidenote:_ @cpojer - is this something you've had to handle in Jest?

cpojer:
[...]

PR #92, opened 2017-01-08
test(DangerRunner): add test cases for async dangerfile

kwonoj:
This PR adds simple test cases for https://github.com/danger/danger-js/issues/88

**NOTE : This PR does not have corresponding fixes for this test yet, should not be merged**


orta:
We've got working async via the schedule function, though it doesn't yet support Observables in a meaningful way :-/

I did head that https://github.com/tc39/proposal-observable is moving forwards :D

kwonoj:
as we have async support at least, I think this PR can be closed now - closing to avoid just PR hanging.

orta:
👍

PR #94, opened 2017-01-08
Adds support for a reading a PRs commits

### LINKED ISSUE #81: [DSL] Add support for pulling in all the commits for a PR
They're on a separate API call `curl https://api.github.com/repos/artsy/emission/pulls/327/commits` (see [this branch](https://github.com/danger/danger-js/tree/commits) ) 

I think we should just expose the full data like we do for `github.pr` - as this is github specific ( it's an API call, and would be different between github/gitlab ) then I think it probably needs to go on `github` so maybe `github.commits` seems fine?

I'd also be interested in discussion about having two versions: `git.commits` and `github.commits` the `git` version could _just_ have the raw cross-API data ( message, parent tree, committer, author ) whereas the github version can contain the full metadata (comments etc)

This means you can write a rule like

```js
const mergeCommits = danger.github.commits.select(commit => commit.commit.message.includes("merge")
if (mergeCommits.length > 0){ fail("I said no merges! sigh")
```

orta:
Fixes #81

First non-trivial PR since we moved to TypeScript, felt really nice. Will comment inline.


.vscode/spellchecker.json
```diff
@@ -0,0 +1,19 @@
...
+    ],
+    "documentTypes": [
+        "markdown",
+        "latex",
+        "plaintext"
+    ],
+    "ignoreRegExp": [],
+    "ignoreFileExtensions": [],
+    "checkInterval": 5000
+}
```
orta: I have a spell checker now

source/api/fetch.ts
```diff
@@ -50,7 +50,7 @@ export function api(url: string | any, init: any): Promise<any> {
     // Handle failing errors
     if (!response.ok) {
       process.exitCode = 1
-      console.error(`Request failed: ${response}`)
+      console.error(`Request failed: ${JSON.stringify(response, null, 2)}`)
       const msg = response.status === 0 ? "Network Error" : response.statusText
```
orta: wasn't showing anything useful, that said, this actually doesn't show the response -only the raw buffer data

source/dsl/DangerDSL.ts
```diff
@@ -23,9 +23,8 @@ export interface DangerDSLType {
...
   public readonly github: Readonly<GitHubDSL>
 
-  constructor(pr: GitHubPRDSL, public readonly git: GitDSL) {
-    this.github = {
-      pr
-    }
+  constructor(platformDSL: any, public readonly git: GitDSL) {
+    // As GitLab etc support is added this will need to be changed
+    this.github = platformDSL
   }
```
orta: it is now not the responsibility of this class to create the `github` object, that's moved to `getPlatformDSLRepresentation`. In the future maybe this should change, whoever is figuring out gitlab can figure that ;)

source/runner/DangerfileRunner.ts
```diff
@@ -82,7 +82,10 @@ export async function runDangerfileEnvironment(filename: Path, environment: Dang
  */
 export function updateDangerfile(filename: Path) {
   const contents = fs.readFileSync(filename).toString()
-  fs.writeFileSync(filename, cleanDangerfile(contents))
+  const cleanedDangerFile = cleanDangerfile(contents)
+  if (contents !== cleanedDangerFile) {
+    fs.writeFileSync(filename, cleanDangerfile(contents))
+  }
```
orta: this was the cause of my watcher issues with Jest

source/api/fetch.ts
macklinu: Would we want to log the JSON response here - `response.json()`? Just a suggestion, could go either way.

source/dsl/Commit.ts
```diff
@@ -0,0 +1,25 @@
+/** A platform agnostic refernce to a Git commit */
+export interface GitCommit {
+  /** The SHA for the commit */
+  sha?: string,
```
macklinu: Is the `sha` optional in a git commit?

source/dsl/GitHubDSL.ts
```diff
@@ -1,9 +1,27 @@
...
+  /** The SHA for the commit */
+  sha: string,
+  /** the url for the commit on GitHub */
+  url: string,
+  /** The GitHub user who wrote the code */
+  author: GitHubUser,
+  /** The GitHub user who shipped the code */
+  committer: GitHubUser,
+  /** An array of parent commit shas */
+  parents: Array<any>
```
macklinu: Two things:

- Would this be an `Array<string>`?
- I never know when to use `Array<string>` vs. `string[]` 🤔

source/platforms/GitHub.ts
```diff
@@ -12,7 +15,8 @@ import * as os from "os"
 
 export type APIToken = string
 
-// TODO: Cache PR JSON
+// TODO: Cache PR JSON?
+// TODO: Separate out a GitHub API client?
```
macklinu: I think that would make sense to do at some point. 👍

source/api/fetch.ts
orta: Yeah, this is not ideal as it shows the buffer, should probably do that

source/dsl/GitHubDSL.ts
orta: I was once told to favour `string[]` in Flow, and `Array<string>` in TS 🍡

kwonoj: Generally in TS any approach's ok as long as there's consistency. My personal preference is generic type for readability, you're getting type of actual object (`Array<>` to inner generic type `string`) left to right order by reading it. Also it gives consistency like 

`class MyArray extends Array<string> { ... }`

extending form can't use array of element type.

I also remember I've told to use generic type as preferred, but I don't recall where and when.

source/api/fetch.ts
kwonoj: maybe it's good place to use `debug` module via `d(response)`, expose only for debug purpose.

source/dsl/DangerDSL.ts
kwonoj: if this will be generic, what about

```counstructor(public readonly platformDSL: Readonly<DSL>, ....) {
}
```

and remove line 24? (basically, rename variable `github` + introduce interface `DSL` for generic instead of any)

source/dsl/GitHubDSL.ts
orta: Yeah this makes sense to me too

source/dsl/Commit.ts
orta: i doubt it, removing

source/dsl/DangerDSL.ts
orta: This object is the root of what we expose as "danger" inside a Dangerfile, so whatever it's shape is ends up being the API for consumers. So I can't really rename it ATM

PR #96, opened 2017-01-14
feat(danger): support custom dangerfile name

kwonoj:
This PR adds one additional commandline arg `-d`, allow to specify specific path / name of dangerfile other than default one. Additionally, it'll check if given / default dangerfile exists before excutes.

### DIFF
--- a/source/commands/danger-run.ts
+++ b/source/commands/danger-run.ts
@@ -1,13 +1,17 @@
 import * as program from "commander"
+import * as debug from "debug"
+import * as fs from "fs"
 import { getCISource } from "../ci_source/get_ci_source"
 import { getPlatformForEnv } from "../platforms/platform"
 import {Executor} from "../runner/Executor"
 
+const d = debug("danger:run")
 declare const global: any
 
 program
   .option("-v, --verbose", "Verbose output of files")
-  .option("-c, --external-ci-provider [modulePath]", "blah")
+  .option("-c, --external-ci-provider [modulePath]", "Specify custom CI provider")
+  .option("-d, --dangerfile [filePath]", "Specify custom dangefile other than default dangerfile.js")
   .parse(process.argv)
 
 process.on("unhandledRejection", function(reason: string, _p: any) {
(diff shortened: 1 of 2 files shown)


source/commands/danger-run.ts
```diff
@@ -1,13 +1,17 @@
...
 import { getCISource } from "../ci_source/get_ci_source"
 import { getPlatformForEnv } from "../platforms/platform"
 import {Executor} from "../runner/Executor"
 
+const d = debug("danger:run")
 declare const global: any
 
 program
   .option("-v, --verbose", "Verbose output of files")
-  .option("-c, --external-ci-provider [modulePath]", "blah")
```
kwonoj: this was my fault to forgot updating description. :(

macklinu:
Personally, I like having the simple convention of looking for a `dangerfile.js`, but I would like to know your use case for this feature, @kwonoj.

I was looking at Danger-rb to see if a similar feature existed, and I found https://github.com/danger/danger/pull/504. It's not a one-to-one correlation, but I would appreciate your thoughts on this and the direction of moving toward https://github.com/danger/danger/pull/504 in Danger-js, if applicable, @orta.

kwonoj:
My primary motivation is about having multiple dangerfile for custom danger execution per each build system.

For example, our team's build agents are consists of one for running default danger only to run simple validation and test coverage, then thinking about platform specific invocation for our platfrom build, which takes more than -30min per each run. 

With single dangerfile it'll work eventually, but means to get danger result per each PR we need to wait around 30min to get results. Instead, separating intoeach dangerfile and let time taking dependent danger runs separately. 

Other than build matrix specific, this could be branch-specific, or either who wants to have physical separations in their config.

kwonoj:
And I'd propse this as non-breaking change, always default to dangerfile but provides configuable manner if someone actively looking way.

source/commands/danger-run.ts
orta: haha, "blah"

source/commands/danger-run.ts
```diff
@@ -1,13 +1,17 @@
 import * as program from "commander"
+import * as debug from "debug"
+import * as fs from "fs"
 import { getCISource } from "../ci_source/get_ci_source"
 import { getPlatformForEnv } from "../platforms/platform"
 import {Executor} from "../runner/Executor"
 
+const d = debug("danger:run")
```
orta: this is good, we needed something like this

PR #97, opened 2017-01-14
feat(template): support timestamp via env. variable

kwonoj:
This PR adds additional check for environment variable `DANGER_TIMESTAMP`, to log date of time dangerfile posted latest update.


orta:
A pattern I found in Danger-rb was to just throw arbitrary data like this inside the table's HTML:

screen shot 2017-01-15 at 11 30 06 am

I'd have no problem with timestamps being on the table all the time e.g: `<table data-danger-timestamp="1980-12-23Z233..."/>`

This means less options, and that data is always there. How does that sound?

kwonoj:
hmm, my intention was actually emit timestamp, instead of having it as arbitrary data.

We sometimes had cases like code author frequently update commit and wasn't sure if danger validation is up to date / or sometimes danger run silently fails (like module installation fail before running danger) so danger itself never got updated but author hard to notice those incidents are happened. For simplicity sake, printing out timestampt each time its updated to check if it's latest or not. Course it does work if it's arbitrary but have to open comment each time.

orta:
Interesting, what about from this:

screen shot 2017-01-15 at 2 21 59 pm

To this:

screen shot 2017-01-15 at 2 22 31 pm

I think what you'd like to see is pretty universal

kwonoj:
Yes, that make sense.

deecewan:
a the risk of beating an old PR, I like that second option a lot.

orta:
I agree, let's ship that - happy to take a fresh PR on that one? could be the two PRs merged

kwonoj:
Yeah, this PR stall too long time as I couldn't give enough attention. Maybe this can be just closed and create new one instead for other way around.

orta:
Works for me 👍 - that's totally understandable

PR #99, opened 2017-01-14
[WIP] Support inline comments

### LINKED ISSUE #77: Support inline comments
Danger-rb's methods support the following parameters:

```rb
warn (message: String, sticky=false: Boolean, file=nil: String, line=nil: String)
```

These methods allow for adding a warning/error/messages inline on a specific file. What would be involved in porting that functionality to Danger-js?

The current type definition of `warn` in Danger-js is:

```ts
warn(message: MarkdownString): void
```

Could we just add more parameters? Would we accept an options object as a second parameter?

```ts
// perhaps this
warn(message: MarkdownString, file=undefined: String, line=undefined: String)

// or
warn(message: MarkdownString, options: { file: String, line: String })
```

Also, where would `sticky: Boolean` come into play - would that be a part of this work, or should that be a part of a separate issue?

macklinu:
Resolves #77 

This is a work-in-progress for #77. I've been sitting on this commit and wanted to get it pushed up and ask some questions about the API and if I'm heading in the right direction before digging in deeper.

TODO:

- [ ] add changelog entry with example of how to use
- [ ] update API documentation (in interfaces like `DangerContext`)
- [ ] create a comment on a specific file/line of a commit in GitHub


source/runner/Dangerfile.ts
```diff
@@ -1,35 +1,35 @@
...
-  message(message: MarkdownString): void
+  message(message: MarkdownString, options?: DangerOptions): void
 
   /**
    * Puts a message inside the Danger table
    *
    * @param {MarkdownString} message the String to output
    */
-  markdown(message: MarkdownString): void
+  markdown(message: MarkdownString, options?: DangerOptions): void
```
macklinu: `markdown()` should support `{ file, line }` options too, right?

source/runner/Dangerfile.ts
```diff
@@ -62,10 +67,10 @@ export function contextForDanger(dsl: DangerDSLType): DangerContext {
     markdowns: []
   }
 
-  const fail = (message: MarkdownString) =>  results.fails.push({ message })
-  const warn = (message: MarkdownString) => results.warnings.push({ message })
-  const message = (message: MarkdownString) => results.messages.push({ message })
-  const markdown = (message: MarkdownString) => results.markdowns.push(message)
```
macklinu: `results.markdowns: Array` differs in that it currently accepts `MarkdownString` as its type, whereas `fails`, `warns`, and `messages` accept `Violation` as its type. To keep parity with the Danger-rb API, `markdown()` should be able to accept `{ file, line }` options, so I believe moving `markdowns` from `Array<MarkdownString>` to `Array<Violation>` is the appropriate change.

source/runner/Executor.ts
```diff
@@ -84,6 +84,7 @@ export class Executor {
         console.log("Found some message, writing it down")
       }
       const comment = githubResultsTemplate(results)
+      // TODO filter out Danger items that contain file/line info
```
macklinu: I believe this is the area where we could filter out results that contained a complete `{ file, line }` object and make inline comments. if `{ file, line }` is incomplete/missing, add them to the main Danger comment (I think this refers to your abstractions comment in https://github.com/danger/danger-js/issues/77#issuecomment-270195643, @orta).

source/runner/templates/github-issue-template.ts
```diff
@@ -33,12 +33,13 @@ function table(name: string, emoji: string, violations: Array<Violation>): strin
...
  * @returns {string} HTML
  */
+// TODO update function to support new DangerResults.markdowns type
 export function template(results: DangerResults): string {
   return `
 ${table("Fails", "no_entry_sign", results.fails)}
 ${table("Warnings", "warning", results.warnings)}
 ${table("Messages", "book", results.messages)}
-${results.markdowns.join("\n\n")}
+${table("Markdowns", "information_source", results.markdowns)}
```
macklinu: `information_source` = ℹ️

source/runner/Dangerfile.ts
orta: I think so, I can see there are good reasons for wanting to put arbitrary markdown on a file 👍

source/runner/Dangerfile.ts
orta: Yeah, this makes sense 👍

source/runner/Dangerfile.ts
```diff
@@ -48,6 +48,11 @@ export interface DangerContext {
   results: DangerResults
 }
 
+export interface DangerOptions {
```
orta: maybe `MessagingOptions` ?

source/runner/Executor.ts
orta: Yes, this is where you want to find:

* Things with no file/line (main comment)
* Things with file/line inside the diff (inline)
* Things with file/line outside the diff (main comment)

source/runner/templates/github-issue-template.ts
orta: 👍

source/runner/Dangerfile.ts
macklinu: That's a better name, will update. 👍

macklinu:
I haven't forgotten about this PR. Been focused on client work recently but should have more time to devote to this feature during the rest of this week/weekend. 👨‍💻

orta:
Sidenote: here is an amazing example of inlining on to the main comment ( https://twitter.com/tivac/status/821994608984924160 )with syntax highlighting etc

source/runner/templates/github-issue-template.ts
macklinu: Actually, I'm going to change this back to just displaying markdown as it's provided, based on this comment - https://github.com/danger/danger-js/issues/109#issuecomment-274298803.

source/runner/Executor.ts
macklinu: This function only accepts the `results: DangerResults`. I'm thinking we will need the `context: DangerContext` (or at least `danger: DangerDSLType`), so that we can call `danger.git.diffForFile(violation.options.file)` and determine if the `{ file, line }` supplied are valid for creating an inline comment. Does that make sense?

macklinu:
This PR has been sitting here and I haven't gotten a chance to devote serious time to it since early January. I'm going to close this for now and leave #77 open for someone to pick up if they'd like to. I'll comment on the issue if I dive back into it, but this is a desirable feature that I don't want to hold up if someone has more time to work on it.

orta:
👍

Issue #20, opened 2016-10-27
Run Danger against existing PRs


orta:
`danger pr https://github.com/github/fetch/pull/413` should run the local Dangerfile.js against that PR. There is an implementation of this in Danger-rb.

macklinu:
I think #100 took care of this - good to close?

orta:
Yep :D

PR #100, opened 2017-01-15
Add support for `danger pr`

orta:
Relates to #20 

Allows running your Dangerfile against an arbitrary PR

screen shot 2017-01-15 at 3 53 57 pm

Will need polish etc, but this is how you can build out a dangerfile without all the crazy env vars we have ATM.

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -1,4 +1,4 @@
-// import { danger, warn } from "danger"
+// Removed import
 import fs from "fs"
 
 // Request a CHANGELOG entry if not declared #trivial
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -1,4 +1,4 @@
-import { danger, warn } from "danger"
+// Removed import
 
 import fs from "fs"
 import includes from "lodash.includes"
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -1,4 +1,5 @@
-// Removed import
+import { danger, warn } from "danger"
+
 import fs from "fs"
 import includes from "lodash.includes"
 
--- a/package.json
+++ b/package.json
@@ -96,9 +96,11 @@
   },
   "dependencies": {
     "babel-polyfill": "^6.20.0",
+    "chalk": "^1.1.1",
     "commander": "^2.9.0",
     "debug": "^2.6.0",
     "jest-runtime": "^18.0.0",
+    "jsome": "^2.3.25",
     "lodash.find": "^4.6.0",
     "lodash.includes": "^4.3.0",
     "node-fetch": "^1.6.3",
--- a/source/commands/danger-pr.ts
+++ b/source/commands/danger-pr.ts
@@ -1,5 +1,50 @@
 import * as program from "commander"
+import * as debug from "debug"
+import * as fs from "fs"
+import { FakeCI } from "../ci_source/providers/Fake"
+import { GitHub } from "../platforms/GitHub"
+import { Executor } from "../runner/Executor"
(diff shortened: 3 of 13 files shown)


source/commands/danger-pr.ts
```diff
@@ -1,5 +1,50 @@
...
-program.parse(process.argv)
+const d = debug("danger:pr")
 
-console.log("Not Yet implmented")
+program
+  .option("-d, --dangerfile [filePath]", "Specify custom dangefile other than default dangerfile.js")
+  .parse(process.argv)
+
+const dangerFile = (program as any).dangerfile || "dangerfile.js"
+const stat = fs.statSync(dangerFile)
```
kwonoj: `fs.statSync` will throw exception if file does not exists, so should catch for non-existing file cases.

orta:
Thanks!


 --
[·/ ]       Orta Therox

w/ Artsy <http://artsy.net/>
CocoaPods <http://cocoapods.org/> / CocoaDocs <http://cocoadocs.org/>
@orta / orta.github.com

source/commands/danger-pr.ts
```diff
@@ -1,5 +1,50 @@
...
+    console.error("Could not get a repo and a PR number from your URL, bad copy & paste?")
+    process.exitCode = 1
+  } else {
+
+    const source = new FakeCI({
+      DANGER_TEST_REPO: pr.repo,
+      DANGER_TEST_PR: pr.pullRequestNumber
+    })
+
+    const platform = new GitHub(null, source)
```
macklinu: It would be cool to read in a GitHub API token somehow for testing against private repo PR URLs. 

Perhaps the user should pass `DANGER_GITHUB_API_TOKEN` as an environment variable when executing the `danger pr` command, or perhaps `danger pr` could support a `--token` option. What do you think?

dangerfile.js
```diff
@@ -1,4 +1,4 @@
-// import { danger, warn } from "danger"
+// Removed import
```
orta: bah :D

source/commands/danger-pr.ts
orta: Yep - we should support the same ENV vars as the Ruby version - http://danger.systems/guides/getting_started.html#enterprise-github

orta: Can look for 404s on the repo, and highlight the options ^

orta:
Added a new dependency, [jsome](https://github.com/Javascipt/Jsome) for JSON formatting, all of it's transitory deps are inside our tree already 👍 

screen shot 2017-01-20 at 10 38 47 am

package.json
```diff
@@ -96,9 +96,11 @@
   },
   "dependencies": {
     "babel-polyfill": "^6.20.0",
+    "chalk": "^1.1.1",
```
orta: Jest uses this version, as does jsome - figured I might start colorizing text for local setup or for [danger init](https://github.com/danger/danger/blob/master/lib/danger/commands/init.rb#L28)

source/commands/danger-pr.ts
```diff
@@ -1,5 +1,58 @@
...
+    let stat: fs.Stats | null = null
+    try {
+      stat = fs.statSync(dangerFile)
+    } catch (error) {
+      console.error("Could not get a repo and a PR number from your URL, bad copy & paste?")
```
macklinu: I think this error should tell the user that it couldn't find the supplied `--dangerfile` (or "dangerfile.js" by default) - parsing the URL happens earlier in the file.

orta: hah - yeah totally

PR #101, opened 2017-01-17
feat(template): provides message summary on top of comment

### LINKED ISSUE #98: Add a Danger comment preview summary
Currently dangerfile support markdown via `markdown()` explicitly, while any other messages are using HTML template tables.

Case of HTML templates, notifications other than github page sometimes does not able to display full contents, by HTML tags takes most of texts and actual contents are being trimmed. For example, below is our current slack github hook notification - 

image

can't see actual messages and have to open github page.

What about github template renders all contents via markdown by default, then introduce some configuration to select HTML based templates instead? In those cases, danger messages will behave 

1. if template's markdown
`message() === markdown()`, all other messages will be rendered to markdown as well

2. if template's html
`message() !== markdown()`, user selectively choose between message & markdown to render contents. All other messages will be rendered into HTML table.

kwonoj:
This PR adds top-level comment in danger messages to provide short summary, formed as

```
<!--
  x failure:  Failing message F...
  x warning:  Warning message W...
  x messages
  x markdown notices
-->
```

in case of failure / warning message, each individual message's truncated by 20 chars.

- closes #98

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,6 +102,7 @@
     "lodash.find": "^4.6.0",
     "lodash.includes": "^4.3.0",
     "node-fetch": "^1.6.3",
-    "parse-diff": "^0.4.0"
+    "parse-diff": "^0.4.0",
+    "voca": "^1.2.0"
   }
 }
--- a/source/runner/templates/github-issue-template.ts
+++ b/source/runner/templates/github-issue-template.ts
@@ -1,5 +1,6 @@
 import { DangerResults } from "../../dsl/DangerResults"
 import { Violation } from "../../platforms/messaging/violation"
+import * as v from "voca"
 
 /**
  * Converts a set of violations into a HTML table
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -102,6 +102,7 @@
     "lodash.find": "^4.6.0",
     "lodash.includes": "^4.3.0",
     "node-fetch": "^1.6.3",
-    "parse-diff": "^0.4.0"
+    "parse-diff": "^0.4.0",
+    "voca": "^1.2.0"
```
orta: I don't think it's worth a whole new dependency for a string truncation method

source/runner/templates/github-issue-template.ts
```diff
@@ -28,13 +29,31 @@ function table(name: string, emoji: string, violations: Array<Violation>): strin
...
+
+function buildSummaryMessage(results: DangerResults): string {
+  const {fails, warnings, messages, markdowns } = results
+  const summary =
+`  ${getSummary("failure", fails)}
+  ${getSummary("warning", warnings)}
+  ${messages.length > 0 ? `${messages.length} messages` : ""}
+  ${markdowns.length > 0 ? `${markdowns.length} markdown notices` : ""}`
+  return summary
+}
```
orta: 👍

package.json
kwonoj: I usually prefer to rely on well-handled module instead of writing own (i.e, lodash, etcs) when there isn't huge constraint about sizes / dependencies. Especially voca is battle tested, probably won't need to worry about edge cases of message strings in general. Is there specific reason to avoid? I can try to switch own impl.

orta: I come from the native world, where for libraries we try to use as few dependencies as possible unless its for something obviously outside of the context of danger (diff parsing for example)

But I guess I can adapt :D

kwonoj: I did similar before :) there's obvious some cases need to be considered such as size of dep (especially on browser target), nested dependency (if there is huge dependency sit behind, even dep itself is small it can bloat up dependency tree) but this specific lib's fine for both cases - by danger targets node only, as well as voca does not have other dependency. 

I'll leave this up to your call.

orta: Yeah, that's 👍 for me

panzerdp: Thanks guys for using Voca library! :+1:  (I am the author)

Voca is trying to fill the gap that JavaScript doesn't have an extensive native string library (other than the few methods on the string type itself). The library does not have any dependencies and will never have, because keeping the dependencies at minimum is a good strategy.

A nice bonus is that the library is decomposable into separated functions. If you want to use only the `truncate()` function, you can easily write:
```javascript
import truncate from 'voca/truncate';

truncate('Once upon a time', 7);
```
And no other dependecies are loaded other than related to `truncate()` function.

PR #102, opened 2017-01-17
Warn if changes to package.json and not yarn.lock

macklinu:
#trivial

Based on https://github.com/facebook/jest/pull/2508

This would have applied to https://github.com/danger/danger-js/pull/101 (discovered when I merged master into and ran `yarn install` on #99).

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -1,5 +1,6 @@
 // import { danger, warn } from "danger"
 import fs from "fs"
+import includes from "lodash.includes"
 
 // Request a CHANGELOG entry if not declared #trivial
 const hasChangelog = danger.git.modified_files.includes("changelog.md")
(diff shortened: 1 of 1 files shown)


macklinu:
I force pushed up an amended commit that would fix the build, but GitHub doesn't seem to be aware of that amended commit yet. Will check back later.

dangerfile.js
```diff
@@ -1,5 +1,6 @@
 // import { danger, warn } from "danger"
 import fs from "fs"
+import includes from "lodash.includes"
```
macklinu: Do we need to worry about Node 4 support, or should I just use the built-in `Array.prototype.includes()` function here?

macklinu:
It looks like there are Appveyor builds queueing up too. I logged in with GitHub but don't see a button to cancel old builds.

image

orta:
Yeah, I've been seeing it on other repos - I'm cool with merging without it

and yeah, I don't think TS backports the includes function, so best use the loads version IMO

orta:
👍

PR #106, opened 2017-01-20
[Danger Runner] Ensure that Danger does not leave changes to the Dangerfile around

orta:
An implementation detail of Danger requires that we "sanitize" the dangerfile to not include a reference to danger as an import. Realistically though, if you want awesome tooling as a user then you need to have that include for the global imports.

So now running `danger pr` or doing the [manual route](https://github.com/danger/danger-js#runningtesting-manually-against-a-repo) won't change your `Dangerfile.js`.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@ then edit the local `Dangerfile.js` and run `yarn run danger pr https://github.c
 
 This will post the results to your console, instead of on the PR itself.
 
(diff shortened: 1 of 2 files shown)


orta:
Gonna make a release and do a quick docs run, so I'm self merging in about ~10-15 once I get the readme/docs up to date.

tivac:
This was super-confusing when getting started, so very glad to see this land!

PR #107, opened 2017-01-20
Update docs and version bump

orta:
* Highlights `danger pr`
* Explains how to do the token dance
* Fixes links for the DSL shapes
* Changes the DSL shape ordering so it's easier to read

### DIFF
--- a/source/dsl/GitDSL.ts
+++ b/source/dsl/GitDSL.ts
@@ -1,5 +1,8 @@
 import { GitCommit } from "./Commit"
 
+// This is `danger.github`
+
+/** The git specific metadata for a PR */
 export interface GitDSL {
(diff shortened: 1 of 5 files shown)


orta:
Alright :1st_place_medal:

source/dsl/GitDSL.ts
```diff
@@ -1,5 +1,8 @@
 import { GitCommit } from "./Commit"
 
+// This is `danger.github`
```
tivac: Should that be `danger.git`?

orta: damnnnnnnnn

orta: :)

Issue #109, opened 2017-01-21
Natural language responses


mxstbr:
I really like the style of the eslint bot, which responds like this:

<img width="722" alt="screen shot 2017-01-21 at 08 26 11" src="https://cloud.githubusercontent.com/assets/7525670/22172908/4acbb978-dfb3-11e6-8011-b8108c63f395.png">

I've noticed that with the "tabular" robotic style people sometimes ignore the bot. Making it respond in natural language would help with that, as we could make it seem like a normal user on the first glance.

Not sure what the API for this would look like, thought I'd kick off a discussion!

KrauseFx:
Interesting idea, definitely something that could easily be achieved, you'd need an easy way to configure a `before` and `after` block. For the user it's also important that it's clear that it's a bot (just having bot in the name isn't enough in my experience)

For now, you can easily start using that kind of responses using the `markdown` method in your `Dangerfile`

mxstbr:
So `markdown` is brilliant, there's only one tiny problem. If one solely uses `markdown`, one cannot make the build fail.

Calling `fail()` without a message renders the table above the markdown, and shows "Undefined" as the message, which is not optimal 😊

If `fail()` without a message failed the build but didn't output any table that'd be epic and would totally suffice for my use case!

macklinu:
@mxstbr your changes from #116 have been published in `danger@0.11.1`. If you are able to update and let us know if using `markdown()` achieves your desired behavior, that'd be wonderful. 😃  

Are we interested in `before` and `after` blocks as suggested by @KrauseFx in https://github.com/danger/danger-js/issues/109#issuecomment-274298803?

macklinu:
Closing this issue, but please let us know if you have suggestions on an API change that would help creating natural language responses. 😊

PR #116, opened 2017-01-29
Don't output table if no violations are specified

mxstbr:
Right now when using `fail` just to fail the build, but not to output
any information (i.e. just using `fail()`), Danger outputs an empty
table with a message of "undefined", which is annoying. (see #109 for
context)

I _think_ this fixes it, but I'm not sure if it's the right fix and/or
if there's a better way to do it. Please let me know!

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,7 @@
 
 //  Add your own contribution below
 
-* Add support for failing the build without outputting any text - mxstbr
+* Builds which only use markdown now only show the markdown, and no violations table is shown - mxstbr
 
 ### 0.10.0
 
--- a/changelog.md
+++ b/changelog.md
@@ -2,11 +2,13 @@
 
 //  Add your own contribution below
 
+* Add support for failing the build without outputting any text - mxstbr
+
 ### 0.10.0
 
 * Adds support for running Danger against a PR locally - orta
 
-The workflow is that you find a PR that exhibits the behavior you'd like Danger to run against, 
+The workflow is that you find a PR that exhibits the behavior you'd like Danger to run against,
 then edit the local `Dangerfile.js` and run `yarn run danger pr https://github.com/facebook/jest/pull/2629`.
 
 This will post the results to your console, instead of on the PR itself.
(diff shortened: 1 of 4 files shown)


source/runner/templates/github-issue-template.ts
```diff
@@ -11,7 +11,7 @@ import * as v from "voca"
  * @returns {string} HTML
  */
 function table(name: string, emoji: string, violations: Array<Violation>): string {
-  if (violations.length === 0) { return "" }
+  if (violations.length === 0 || (violations.length === 1 && !violations[0].message)) { return "" }
```
orta: Lets get a test on this logic, as this is not something which would get manually tested very often

changelog.md
```diff
@@ -2,11 +2,13 @@
 
 //  Add your own contribution below
 
+* Add support for failing the build without outputting any text - mxstbr
```
orta: This should probably be something more like:

> Builds which only use `markdown` now only show the markdown, and no violations table is shown - mxstbr

Which makes a bit more sense without the context of this PR 👍

source/runner/templates/github-issue-template.ts
orta: This area is already pretty well covered, so it should be pretty trivial to C&P an existing one

mxstbr:
I just realised that this would still show a table in this case:

```JS
fail()
fail()
```

Which doesn't make sense. Will update, one sec.

macklinu:
@mxstbr haha I was just about to comment on that potential edge case - nice catch! 😄

orta:
@macklinu I'd love to give you permission to push

source/runner/templates/github-issue-template.ts
```diff
@@ -11,7 +11,7 @@ import * as v from "voca"
  * @returns {string} HTML
  */
 function table(name: string, emoji: string, violations: Array<Violation>): string {
-  if (violations.length === 0) { return "" }
+  if (violations.length === 0 || violations.every(violation => !violation.message)) { return "" }
```
orta: cool, the `every` function is new to me

orta:
My rule of thumb on releasing is roughly every 2-3 PRs, I have zero problems with every PR, assuming every PR is big enough to warrant the ~2 minutes to deploy

orta:
screen shot 2017-01-29 at 2 08 59 pm

source/runner/templates/github-issue-template.ts
mxstbr: Yeah, I've only known it for a few weeks! (there's also `some` FYI)

mxstbr:
> I have zero problems with every PR, assuming every PR is big enough to warrant the ~2 minutes to deploy

That sounds like the perfect use case for `semantic-release` 😉

Issue #117, opened 2017-01-30
Document release process


macklinu:
@orta thanks for giving me NPM publish access (https://github.com/danger/danger-js/pull/116#issuecomment-275937542). I wanted to open an issue to add documentation to the README about publishing this package.

My assumptions on how this would go:

* Bump the version number. In the case of #116, I would run `npm version patch` to create an 0.10.1 bugfix version (which would also create a git tag). We could coordinate on minor and especially major version releases, but I would feel comfortable shipping 0.10.1 as a bugfix release.
* Push that commit to GitHub
* Run `npm publish`, which would build and publish danger to NPM

Could you please clarify? I'll turn this issue into a PR with documentation. Thanks! 😄 🚢 

EDIT: should have checked before opening the issue, but 0.10.1 was shipped in 8775f48fdb4c0ff8a9e034aec92eaa9bb9896abf, which looks like we just:

* Bump the version number manually
* Modify the changelog to include the new version number
* Commit that and push to GitHub
* Run `npm publish`?

orta:
Yep, it's all of that on the bottom 👍

orta:
In Ruby I tend to do these: 
https://github.com/danger/danger/pull/705

tivac:
Any reason not to have Travis do the publishing of tagged commits to NPM? It's [super-easy](https://docs.travis-ci.com/user/deployment/npm/)!

I've started setting up my repos so that running `npm version` will kick off tests, update the changelog, push the commit, and then travis publishes. I love it.

orta:
I'm into this, yeah :D

For me `npm version` just shows some metadata?

tivac:
`npm version <new-version>` where `new-version` is one of `major`/`minor`/`patch` to auto-increment that field, or some arbitrary version like `1.2.3-rc6` to set it to that specific value.

[docs](https://docs.npmjs.com/cli/version)

macklinu:
This sounds like a great idea, @tivac. 👍 

So we should be able to use something like this in `.travis.yml`:

```yml
deploy:
  ...
  on:
    tags: true
```

And when `npm version` creates a new tag and we push to GitHub, it will deploy that new version to NPM, right?

[...]

PR #129, opened 2017-01-30
Support deploying via travis

orta:
This adds all of the authentication part of deploying via travis, via a unique dangersystems user account: https://www.npmjs.com/~dangersystems


orta:
Re: #117

PR #131, opened 2017-01-30
add dockercloud to list of providers

camacho:
forgot to add it to the list of providers :man_facepalming:


orta:
Hah - ouch, I wonder if we should include a test in each CI provider that it returns itself for a set of ENV vars?

orta:
I can do this, I got a bit of time this second 👍

orta:
also, I've invited you to the org, as a part of [Moya Community Continuity](https://github.com/Moya/contributors/issues/6) - you're welcome to participate in PRs and help out at a level that works for you. No pressure.

Issue #80, opened 2017-01-05
[DSL] Migrate in some generic helper functions


orta:
From https://github.com/facebook/jest/pull/2508

`danger.github.linkFiles` ->

```js
// Takes a list of file paths, and converts it into clickable links
const linkableFiles = (paths: Array<string>): string => {
  const repoURL = danger.github.pr.head.repo.html_url;
  const ref = danger.github.pr.head.ref;
  const links = paths.map(path => {
    return createLink(`${repoURL}/blob/${ref}/${path}`, path);
  });
  return toSentence(links);
};
```

I think we probably also need util functions that are specific to universal Danger use-cases, mainly as the JS core lib is relatively weak:

`danger.utils.sentence`:

```js
// ["1", "2", "3"] to "1, 2 and 3"
const toSentence = (array: Array<string>) : string => {
  if (array.length === 1) { return array[0]; }
  return array.slice(0, array.length - 1).join(', ') + ' and ' + array.pop();
};
```

`danger.utils.href`:

```js

// ("/href/thing", "name") to "<a href="/href/thing">name</a>"
const createLink = (href: string, text: string): string => 
  `<a href='${href}'>${text}</a>`;
```

orta:
These are in 👍

PR #134, opened 2017-01-31
Add danger.utils DSL

macklinu:
Relates to #80

`danger.utils` includes two functions:

* `href(href: string, text: string): string`
* `sentence(array: Array<string>): string`

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,18 @@
 
 //  Add your own contribution below
 
+* Add `danger.utils` DSL, which includes `danger.utils.href()` and `danger.utils.sentence()` - macklinu
+
+  We were finding that a lot of Dangerfiles needed similar functions, so we've added a `utils` object
+  to offer functions that are going to be used across the board. If you can think of more 
+  functions you use, we'd love to add them. Ideally you shouldn't need to use anything but Danger + utils
+  to write your Dangerfiles.
+
+```js
+danger.utils.href("http://danger.systems", "Danger") // <a href="http://danger.systems">Danger</a>
+danger.utils.sentence(["A", "B", "C"]) // "A, B and C"
+```
+
 ### 0.12.1
 
 * Add support for [Drone](http://readme.drone.io) - gabro
--- a/source/dsl/DangerDSL.ts
+++ b/source/dsl/DangerDSL.ts
@@ -1,5 +1,6 @@
 import { GitDSL } from "../dsl/GitDSL"
 import { GitHubDSL } from "../dsl/GitHubDSL"
+import { DangerUtilsDSL } from "./DangerUtilsDSL"
 
 /**
  *  The Danger DSL provides the metadata for introspection
@@ -16,6 +17,11 @@ export interface DangerDSLType {
    *  The GitHub metadata.
    */
   readonly github: Readonly<GitHubDSL>
+
+  /**
+   * Danger utils
+   */
+  readonly utils: Readonly<DangerUtilsDSL>
 }
 
 /* END FLOWTYPE EXPORT */
@@ -23,7 +29,7 @@ export interface DangerDSLType {
 export class DangerDSL {
   public readonly github: Readonly<GitHubDSL>
 
-  constructor(platformDSL: any, public readonly git: GitDSL) {
+  constructor(platformDSL: any, public readonly git: GitDSL, public readonly utils: DangerUtilsDSL) {
     // As GitLab etc support is added this will need to be changed
     this.github = platformDSL
   }
--- a/source/dsl/DangerUtilsDSL.ts
+++ b/source/dsl/DangerUtilsDSL.ts
@@ -1,0 +1,23 @@
+/**
+ * The Danger Utils DSL contains utility functions
+ * that are specific to universal Danger use-cases.
+ */
+export interface DangerUtilsDSL {
+
+  /**
+   * Creates an HTML link.
+   *
+   * @param {string} href The HTML link's destination
+   * @param {string} text The HTML link's text
+   * @returns {string} The HTML <a> tag
+   */
+  href(href: string, text: string): string
+
+  /**
+   * Converts an array of strings into a sentence.
+   *
+   * @param {Array<string>} array The array of strings
+   * @returns {string} The sentence
+   */
+  sentence(array: Array<string>): string
+}
--- a/source/dsl/DangerUtilsDSL.ts
+++ b/source/dsl/DangerUtilsDSL.ts
@@ -7,17 +7,22 @@ export interface DangerUtilsDSL {
   /**
    * Creates an HTML link.
    *
-   * @param {string} href The HTML link's destination
-   * @param {string} text The HTML link's text
-   * @returns {string} The HTML <a> tag
+   * If `href` and `text` are falsy, null is returned.
+   * If `href` is falsy and `text` is truthy, `text` is returned.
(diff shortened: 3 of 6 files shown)


source/dsl/DangerDSL.ts
```diff
@@ -16,14 +17,19 @@ export interface DangerDSLType {
    *  The GitHub metadata.
    */
   readonly github: Readonly<GitHubDSL>
+
+  /**
+   * Danger utils
```
macklinu: Better JSDoc for this?

source/dsl/DangerDSL.ts
```diff
@@ -16,14 +17,19 @@ export interface DangerDSLType {
...
+  readonly utils: Readonly<DangerUtilsDSL>
 }
 
 /* END FLOWTYPE EXPORT */
 
 export class DangerDSL {
   public readonly github: Readonly<GitHubDSL>
 
-  constructor(platformDSL: any, public readonly git: GitDSL) {
+  constructor(platformDSL: any, public readonly git: GitDSL, public readonly utils: DangerUtilsDSL) {
```
macklinu: Should `utils` be passed in or created in the constructor, like:

```ts
constructor(platformDSL: any, public readonly git: GitDSL) {
  this.utils = { href, sentence }
}
```

source/runner/DangerUtils.ts
```diff
@@ -0,0 +1,13 @@
+export function sentence(array: Array<string>): string {
```
macklinu: The JSDoc for these functions live in `source/dsl/DangerUtilsDSL.ts`. Is it appropriate to break up the interface and implementation like this?

source/runner/DangerUtils.ts
```diff
@@ -0,0 +1,13 @@
...
+    return ""
+  }
+  if (array.length === 1) {
+    return array[0]
+  }
+  return array.slice(0, array.length - 1).join(", ") + " and " + array.pop()
+}
+
+export function href(href: string, text: string): string {
+  return `<a href="${href || "#"}">${text || ""}</a>`
```
macklinu: Handling falsy input if that's something to worry about - let me know what you think.

changelog.md
```diff
@@ -2,6 +2,13 @@
 
 //  Add your own contribution below
 
+* Add `danger.utils` DSL, which includes `danger.utils.href()` and `danger.utils.sentence()` - macklinu
+
+```js
+danger.utils.href("http://danger.systems", "Danger") // <a href="http://danger.systems">Danger</a>
+danger.utils.sentence(["A", "B", "C"]) // "A, B and C"
```
macklinu: If you have suggestions for a better changelog entry (both the entry and the examples), I'm all ears. 👂 👍

source/dsl/DangerDSL.ts
orta: I don't think it should be needed, I expect the `DangerUtilsDSL` to show up here 👍

source/dsl/DangerDSL.ts
orta: I think as-is feels more consistent (which is probably why you went this way in the first place ) - otherwise there's a lot of DI

source/dsl/DangerUtilsDSL.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * The Danger Utils DSL contains utility functions
+ * that are specific to universal Danger use-cases.
+ */
+export interface DangerUtilsDSL {
+
+  /**
+   * Creates an HTML link.
+   *
+   * @param {string} href the HTML link's desintation
```
orta: > desintation

source/dsl/DangerUtilsDSL.ts
```diff
@@ -0,0 +1,23 @@
...
+   */
+  href(href: string, text: string): string
+
+  /**
+   * Converts an array of strings into a sentence.
+   *
+   * @param {Array<string>} array the array of strings
+   * @returns {string} the sentence
+   */
+  sentence(array: Array<string>): string
```
orta: haha, the objective-c person in me is screaming "call this toSentence" 

```js
import { utils } from 'danger'
const results = utils.sentence(problems)
```

Yeah, I guess this feels about right

source/runner/DangerUtils.ts
orta: Yes please, specifically so we can easily parse out the interfaces to make d.ts and flow defs

source/runner/DangerUtils.ts
orta: could make this return `string | null` and have the above `sentence` strip null objects?

changelog.md
orta: Only a minor adjustment on _why_ it exists (I also added a "open to more ideas" too )

```
* Add `danger.utils` DSL, which includes `danger.utils.href()` and `danger.utils.sentence()` - macklinu

  We were finding that a lot of Dangerfiles needed similar functions, and so have added a `utils` object
  where we could offer functions that are going to be used across the board. If you can think of more 
  functions you use, we'd love to add them. Ideally you shouldn't need to use anything but Danger + utils
  to write your Dangerfiles.

  ```js
  danger.utils.href("http://danger.systems", "Danger") // <a href="http://danger.systems">Danger</a>
  danger.utils.sentence(["A", "B", "C"]) // "A, B and C"
  ``X

```

source/runner/DangerUtils.ts
macklinu: I guess my concern here was that these functions don't have an JSDoc, but they correspond directly to the `DangerUtilsDSL` interface definition. I don't want to write the JSDoc twice, so is it cool to leave these functions without JSDoc? If the signature of these functions ever changed and the `DangerUtilsDSL` did not, TypeScript would not compile, so I'm thinking we are 👍  here.

source/runner/DangerUtils.ts
macklinu: I'm not sure I follow with the `string | null` type. If `href` or `text` are null, would we return `null` to conform to the `string | null` return type? I figured `"#"` and `""` were sane defaults, respectively, but I was just throwing those out in case someone decides to supply `null` or `undefined` for either parameter. 🤷‍♂️

macklinu:
Rebased against master, pushed up, and addressed a couple comments. Left some questions on others. Let me know what other changes I can make here! 👍

source/runner/DangerUtils.ts
orta: I don't mind the if JSDoc for them is extremely limited ( if for example we had to add it due to tslint ) mainly because the type system would keep us in line as you say

source/runner/DangerUtils.ts
orta: Hrm, what about:

```
// No href basically means don't bother with the URL?
!href && !text = null
!href && text = text

// Use the last component cause it's nearly always gonna be a file
href && !text = href, text = last component of href

href && text = href, text
```

It's a bit more optimized towards showing file paths

orta: ^ make sense?

macklinu:
@orta pushed up 17e70fa based on https://github.com/danger/danger-js/pull/134#discussion_r98809406 - let me know what you think about the direction of `danger.utils.href()`. 👍

orta:
Gonna see if I can get a GitHub utils in, then I'll ship this and #173

PR #136, opened 2017-01-31
Separates out the GitHub and GitHub api

orta:
This isn't exactly perfect, but I dropped a feature that I think may still be needed for Peril to migrate to the latest Danger - so I'm going to re-add it, and cover it with tests :D

- [x] Add some more tests to the GitHubAPI
- [x] Add support for additional headers to the API object


orta:
OK, this should be good for review - glad we got so much infra to tell me I'm doing things wrong :D

orta:
Alright, doing this 👍

macklinu:
Thanks for cleaning this up @orta 👍

PR #141, opened 2017-02-08
clarify scripts setup in readme

zeke:
This PR clarifies how the run script should be added to package.json, to prevent new folks from mistakenly adding `danger` as a top-level key in the file.


orta:
👍

orta:
nice thinking

PR #142, opened 2017-02-17
add a codeship provider

deecewan:
This is an initial implementation

### DIFF
--- a/source/ci_source/providers/Codeship.ts
+++ b/source/ci_source/providers/Codeship.ts
@@ -1,5 +1,5 @@
 import { Env, CISource } from "../ci_source"
-import { ensureEnvKeysExist, ensureEnvKeysAreInt } from "../ci_source_helpers"
+import { ensureEnvKeysExist } from "../ci_source_helpers"
 
 export class Codeship implements CISource {
   constructor(private readonly env: Env) {
@@ -19,25 +19,9 @@ export class Codeship implements CISource {
     return false;
   }
 
-  private _prParseURL(): {owner?: string, reponame?: string, id?: string} {
-    const prUrl = this.env.CI_PULL_REQUEST || ""
-    const splitSlug = prUrl.split("/")
-    if (splitSlug.length === 7) {
-      const owner = splitSlug[3]
-      const reponame = splitSlug[4]
-      const id = splitSlug[6]
-      return {owner, reponame, id}
-    };
-    return {}
-  }
-
   get pullRequestID(): string {
-    if (this.env.CIRCLE_PR_NUMBER) {
-      return this.env.CIRCLE_PR_NUMBER
-    } else {
-      const {id} = this._prParseURL()
-      return id || ""
-    }
+    // this will need to retrieve from the github server, if it retrieves at all
+    return '';
   }
 
   get repoSlug(): string {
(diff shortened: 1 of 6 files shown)


deecewan:
I mostly just have a question right now.

Currently, Codeship does not support the Pull Request functionality.
What are your thoughts on querying the Github API for currently open PRs and checking if the currently pushed branch belongs one of them?  It's nowhere near as clean as using a simple environment variable, sure.  But it should give a fairly good approximation.

orta:
Yeah, I think the GitHub bfanch checker is a good call - try make that as an abstract class because I'm pretty sure that other CI providers maty be able to use it in the future 👍

deecewan:
@orta thanks for getting back to me.

Is the github API key exposed on `this.env`?  Also, any thoughts on what to do in the case that one branch is in 2 PR's at the same time?  Just take the first?

deecewan:
So, this change has ended up being a lot bigger than I expected it to be.  Open to any feedback or changes you think are necessary, but I couldn't see a straightforward, agnostic was of performing async setup without doing this.

source/ci_source/providers/Codeship.ts
```diff
@@ -0,0 +1,40 @@
+import { Env, CISource } from "../ci_source"
+import { ensureEnvKeysExist, getPullRequestIDForBranch } from "../ci_source_helpers"
+
+export class Codeship implements CISource {
+  private default = { prid: "0" }
```
orta: I learn something new on almost every code review 👍

source/ci_source/providers/Codeship.ts
```diff
@@ -0,0 +1,40 @@
...
+  constructor(private readonly env: Env) {
+  }
+
+  async setup(): Promise<any> {
+    const prid = await getPullRequestIDForBranch(this, this.env, this.branchName)
```
orta: any chance this can be `prID` (makes it more consistent with the rest of the file)

source/commands/danger-run.ts
```diff
@@ -26,45 +26,54 @@ if (process.env["DANGER_VERBOSE"] || (program as any).verbose) {
...
       }
-    } catch (error) {
-      process.exitCode = 1
-      console.error(error.message)
-      console.error(error)
     }
   }
 }
+
+run()
```
orta: Hah, I hadn't thought of that workaround, previously I had just avoided doing anything async in here

source/ci_source/providers/Codeship.ts
deecewan: definitely.  i think it was that at some point during dev, too.

source/ci_source/providers/Codeship.ts
deecewan: is this not standard?  I've never used typescript before.

orta:
Hrm, interesting - Danger is definitely failing, but is not presenting the github message

screen shot 2017-02-19 at 19 45 17

deecewan:
that's interesting.  I ran this locally by emulating the CI environment, and all went perfectly.

to clarify, Danger is receiving a `fail` call, but isn't displaying it or posting it to GH?

orta:
Oh, it's commenting, did I miss something? Everything looks 100%

screen shot 2017-02-20 at 08 33 20

This is all fine. Maybe I didn't notice that comment?

orta:
I'll add a changelog and update the README

PR #144, opened 2017-02-20
[WIP] initial implementation to handle async

deecewan:
This still does not work.

It currently allows functions to be pushed to the scheduled array, and it tries to call them.  I have not had it call successfully yet.

It was pretty late at night when I wrote this, so I apologise if I've missed anything.

Tested with the following dangerfile:

```js
import { warn, scheduled } from 'danger';

const p = new Promise(res => {
  setTimeout(() => {
    console.log('just finished doing something async');
    warn('Async function call');
    res();
  }, 1000);
});

schedule(async () => {
  await p;
});
```

I end up with, at the end, a context that looks like:

```js
{
 // ...
 warnings: [],
 scheduled: [null],
}
```

### DIFF
--- a/source/dsl/DangerResults.ts
+++ b/source/dsl/DangerResults.ts
@@ -8,6 +8,11 @@ import { MarkdownString } from "../dsl/Aliases"
  */
 export interface DangerResults {
   /**
+   * Asynchronous functions to be run after parsing
+   * @type {Function[]}
+   */
+  scheduled: Array<Function>
+  /**
    * Failed messages
    * @type {Violation[]}
    */
--- a/source/dsl/DangerResults.ts
+++ b/source/dsl/DangerResults.ts
@@ -7,11 +7,7 @@ import { MarkdownString } from "../dsl/Aliases"
  * do not add functions, only data to this interface.
  */
 export interface DangerResults {
-  /**
-   * Asynchronous functions to be run after parsing
-   * @type {Function[]}
-   */
-  scheduled: Array<Function>
+
   /**
    * Failed messages
    * @type {Violation[]}
@@ -35,4 +31,12 @@ export interface DangerResults {
    * @type {MarkdownString[]}
    */
   markdowns: Array<MarkdownString>
+}
+
+export interface DangerRuntimeContainer extends DangerResults {
+  /**
+   * Asynchronous functions to be run after parsing
+   * @type {Function[]}
+   */
+  scheduled: Array<Function>
 }
--- a/source/runner/Dangerfile.ts
+++ b/source/runner/Dangerfile.ts
@@ -1,4 +1,4 @@
-import { DangerResults } from "../dsl/DangerResults"
+import { DangerRuntimeContainer } from "../dsl/DangerResults"
 import { DangerDSLType } from "../dsl/DangerDSL"
 import { MarkdownString } from "../dsl/Aliases"
 
@@ -52,7 +52,7 @@ export interface DangerContext {
    *
    * @type {DangerDSLType}
    */
-  results: DangerResults
+  results: DangerRuntimeContainer
 }
 
 /** Creates a Danger context, this provides all of the global functions
@@ -62,7 +62,7 @@ export interface DangerContext {
  * @returns {DangerContext} a DangerContext-like API
  */
 export function contextForDanger(dsl: DangerDSLType): DangerContext {
-  const results: DangerResults = {
+  const results: DangerRuntimeContainer = {
     fails: [],
     warnings: [],
     messages: [],
--- a/source/runner/Dangerfile.ts
+++ b/source/runner/Dangerfile.ts
@@ -4,6 +4,13 @@ import { MarkdownString } from "../dsl/Aliases"
 
 export interface DangerContext {
   /**
+   * Contains asynchronous code to be run after the application has booted.
+   *
+   * @param {Function} asyncFunction the function to run asynchronously
+   */
+  schedule(asyncFunction: (p: Promise<any>) => void): void
+
+  /**
    * Fails a build, outputting a specific reason for failing
    *
    * @param {MarkdownString} message the String to output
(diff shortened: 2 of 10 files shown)


orta:
I've taken a look also, resulting in https://github.com/danger/danger-js/compare/async-orta-dangerfile?expand=1

I messed around a bunch too (trying to figure a promises thing with Peril )

I think if you try move the `Promise.all` Into `runDangerfileEnvironment` you get better results ( and it would work across the board, as opposed to just inside `danger run`

This seems to act as we'd expect. If I had to guess why, I would guess all the runtime objects were going out of memory scope before we waited on the results.

orta:
Ended up with:
screen shot 2017-02-21 at 18 34 23

Rather than:
screen shot 2017-02-21 at 18 40 41

deecewan:
looks pretty good! seems i did put the runner in the wrong place 😂 . New to the project and still trying to figure my way around the flow properly.

did you happen to try it out pushing a warning/message from within the asynchronous code?

orta:
I didn't then, just tested on that branch and it works 👍

orta:
I'd screenshot, but my MBP trackpad and keyboard sdoesnt work, so in the evenings I code in bed with an external keyboard :D 🍡

deecewan:
brilliant.  i'll close this PR and maybe you can open yours?

orta:
I'd recommend taking my change and cleaning it ups, I wouldn't bother taking the commit either - this will still need a test or two to ensure it all works in the future ( and I want to concentrate on Peril support for the next 2-3 days )

deecewan:
sure thing. I'll use your commit for inspiration.

also, yes. tests. good point.

deecewan:
okay, I've added some tests as well.

I should point out that, at this stage, this won't postpone the execution of functions and only allows promises to be scheduled.  That is to say,

```js
schedule(new Promise(res => { warn('warning'); res() })); // successfully adds a warning

schedule(() => warn('warning')); // does nothing

schedule(() => new Promise(res => { warn('warning'); res() })); // does nothing

schedule(async () => {
  const result = await someAsyncFunction();
  if (result) warn(result);
}); // does nothing
```

I think, but I'm open to opinions, that we should be expecting functions inside `scheduled` which are then executed and awaited if they return promises, or just executed if they are regular functions (although I'm not sure why someone would want to delay execution like that...)

kwonoj:
How does `schedule` works with other type of async than promises? 

For example, most of node.js apis are callback-based async (`fs.readfile(opt, (err, result) => {});`), or RxJS (`Observable.interval(500).map(_ => doSomething()).subscribe(...)`) has different kind of async patterns. Will user writes dangerfile should conform specific type of async (promise) and should have their own way to interop between other type of async? This was my primary question when I first tried to tackle down this issue and couldn't come up with feasible solutions.

deecewan:
I think it would be a bad idea to try and support *every* type of async. The standard ones, for sure. 

I'll add support for a callback tonight, too. And update the readme (maybe somewhere else?) with how it works.

deecewan:
@orta any reason that only appveyor is running here?

orta:
Hrm, good Q, circle was having issues last night

orta:
I don't think we can do anything about the callback based APIs, but we should be able to handle Promises and Observables without much effort on Danger's part I imagine? 

We could have two functions `schedule` for promises, and an `observer`

I've only use RxSwift, so excuse my knowledge on RxJS but would the observables also need to be manually closed in the Dangerfile? Or would it be reasonable for us to assume that when an object is passed down the chain we can then stop? e.g.

```js
const source = getAsyncStockData();

const subscription = source
  .filter(quote => quote.price > 30)
  .map(quote => quote.price)
  .forEach(price => {
    fail(`Price is too much ${price}`)
   });

danger.observe(subscription)
``

kwonoj:
> I think it would be a bad idea to try and support every type of async.

: I have a bit different opinion to this matter. `dangerfile` is plain javascript and it has implicit assumption to work as plain javascript - cases I'm concerning is about dependencies user will try to use. If certain module have non-supported underlying async mechanism, user will face unexpected behavior in dangerfile and have to dig each module's internals to figure out what's happening. I'd expect that'll be quite troublesome for users, especially if dependency module's quite complicated implementation.

If we're going to have certain constraint to dangerfile to support one and not support one, it'll be probably better to introduce strong interface / contract to dangerfile itself to give explicit indication dangerfile is *not* plain javascript that user can execute anything as same as other js code, like config files of `webpack` or similar does.

orta:
I don't have too strong opinions on it TBH, I've always considered the Dangerfile to be a normal JS file with a few injected globals. You should write it with the same tools that you use to write the rest of your apps. Whether that is Promises, Observables or Callbacks is fine by me.

I think a lot of the issue lies around priming and docs, so long as we document the right way to use any of them in a way that works as expected, I'm 👍 with adding individual aspects of each the DSL.

deecewan:
@orta I have already included support for callback-based APIs.  If the function accepts a parameter, I assume it's a callback and wrap it in a promise.  See https://github.com/danger/danger-js/pull/144/commits/9e7515abaa20554f36606cfb47103a937e0bfb6d for details.

I think I agree with you if what you're saying is 'support what we can, document how it works'.  I think if we can add more support in the future, that's not a bad thing. But I think that a lack of support for Observables is not a good reason for lacking functionality for almost all other types of async.

Excuse my ignorance about observables, I've never actually used them.  But, surely they continously stream?  How do you know when they 'complete'?  If they have a way to know if they're complete, they can use the callback method to fire the 'done' function.  If they don't, then how would Danger ever know when to continue processing?

orta:
Yeah, I'm with this - I'd love to see this handle observables in a nice way also! For now though, that can be another PR

I'm gonna merge this and ship up some docs fixes - thanks @deecewan  
also, I've invited you to the org, as a part of [Moya Community Continuity](https://github.com/Moya/contributors/issues/6) - you're welcome to participate in PRs and help out at a level that works for you.

PR #149, opened 2017-02-21
Don't export our own promise interfaces to the d.ts

orta:
Currently we use the Promise implementation provided by TypeScript for ES2015 support, so don't include references to a different version.


orta:
interesting, looks like circle is having issues

orta:
OK, yeh, this is fine

PR #158, opened 2017-03-03
BuddyBuild support

benkraus:
This PR adds support for BuddyBuild. They were awesome and got me the env vars needed.


benkraus:
Hm, for some reason it's still failing to detect the CI when running on BuddyBuild. I'll look into that...

benkraus:
Hm, @orta am I being obviously stupid somehow? BuddyBuild's logs say it can't find a CI source... The feedback loop is huge to figure this out with having to update this then my other app that's pointing at this fork, and see if it works or not.

orta:
Sorry, on my honeymoon - conceptually this is all I think you should need to do: create the class, add to the providers index. Which it does look like what you've done

orta:
It's worth reaching out to the buddy build devs, I think for Danger RB they have code on their side to run Danger correctly

benkraus:
Yeah, I talked to them and they don't support DangerJS currently, but eventually will. This was meant to sidestep that, get it working in the short term. Still need to figure out why it's not detecting this...

Palleas:
Hey! We're currently looking into this, what do you need from us? :)

orta:
@Palleas it's mainly about exposing ENV vars - should be the exact same ones as Danger on Ruby

clintam:
Howdy. 

Thanks @benkraus for producing this change. We are in the process of adding support for dangerjs to buddybuild, and this change was enough to get it working for me.

I've taken it, rebased on master, and applied some changelog/readme as danger suggested here: https://github.com/danger/danger-js/pull/286

orta:
Deprecated by those other PRs - great job @benkraus and @clintam

orta:
This is shipped in `0.21.1`

PR #161, opened 2017-03-12
Update TypeScript, Jest and the CHANGELOG for async

orta:
Doesn't handle all of the greenskeeper PRs, but it does a bunch and I could manually go through them 👍


orta:
Alright, that should bring the PRs back to some stability and update the CHANGELOG 👯‍♂️

Issue #84, opened 2017-01-06
create a dummy test repo and run it against every new release


dblock:
For issues like https://github.com/danger/danger-js/issues/83 it would be good to have a test repo that maybe makes a version update on a trigger for every new danger.js release.

kwonoj:
Or maybe this repo's PR run danger twice, one for latest stable release from npm, one for linked from given PR to verify regression & sanity.

orta:
We do have CircleCI running a separate build, I think it would be 👍 to move that to use the latest stable build instead of master which Travis runs.

orta:
Fixed by https://github.com/danger/danger-js/pull/172

Issue #169, opened 2017-03-14
Support Danger.ts files 


orta:
As a lot of our Artsy JS projects are moving/moved to TypeScript, I think it makes sense to consider a `Dangerfile.ts` legit also. 

When you use TypeScript with jest, you declare a different transformer (ts-jest) so I wonder if this will work the same in Danger.

deecewan:
interesting.  I haven't really used typescript enough, but with babel, you can `require('babel-core/register')`.  We could do a check on the file, and if it ends in `ts`, require the typescript runtime transformer (assuming it exists).

orta:
Added in https://github.com/danger/danger-js/pull/172

Issue #166, opened 2017-03-14
Support editing existing comments instead of overwriting


deecewan:
I think it'd be nice (maybe it happens on `danger.rb`?) for problems that are resolved to be struck out? It makes it hard to track progress and reasons for certain commits.

In this, I'd also like to see some logic to check if the comment if actually a danger comment before overwriting.  We use the same bot on a bunch of repos, and it comments about various things in various places.  It took me a while to realise that it was still working, but danger was overwriting the comment instead of creating a new one.

orta:
We do support strikeouts in Danger.rb - I'm hesitant to add it in, as it adds a lot of complexity but ended up rarely used when I asked people if they used it.

In rb there is a `--danger_id` flag that you can pass, and then it would only update/edit/delete the comment with the same ID. This allowed for multiple Danger's to comment on the same thread. Would that fix it?

deecewan:
yeah, as i was writing the strikeout thing, I don't really think it adds much to be honest.

Editing comments would be 100% great, although I'm not sure how I'd know which comment to flag ahead of time?

orta:
Danger JS is already editing comments, and deleting them as they become empty. 

Markdown allows for arbitrary text to be inserted inside HTML elements ;)

<img width="683" alt="screen shot 2017-03-15 at 08 10 25" src="https://cloud.githubusercontent.com/assets/49038/23926321/dfb73900-0956-11e7-9a4d-f4e3a16b610e.png">

So I usually put data like that inside the table element

deecewan:
i guess the thing is how does it know which comment to edit?

I had a case where the bot would comment with the IP of a server, and then danger would run and overwrite the server IP.  After that it just edited it's own comment.

orta:
So the danger-id thing is used for when you have `danger` being ran multiple times on the same commit ( maybe from different build machines etc )

in RB, the flag `--dangerID=123` would mean that inside the markdown there would be a `data-danger-id="123"` then later as Danger is looking for the right comment to edit, it checks that there is a corresponding `dangerID` and the one that was ran during the argument. If it doesn't find one, then it creates a new comment instead

deecewan:
right.  I think what i'm proposing is more 'only edit a comment that Danger has created, otherwise add a new one'.  Maybe we can use a flag for that to retain existing behavior if people want it.

orta:
This feels like the current behavior to me though? can you write out a flow
of what would happen after these changes

orta:
Related #196

[...]

Issue #162, opened 2017-03-12
Support text only output for Danger


orta:
This could be useful for this project, as it would allow us to run `danger` on any CI service that Danger supports, but we would not see duplicate messages.

Probably should be a command line flag on `danger run`.

PR #172, opened 2017-03-16
TypeScript support, + danger.d.ts, + danger DSL change

orta:
Adds support for `dangerfile.ts` and uses the Jest config system to handle transformation.

---

**Previous notes:**

This PR started small, but had a while to do some work on the trainf, so I programmed away.

This provides the initial basic setup for TypeScript. I added a test but it is way slow, 41 seconds. 

So this will fail. The fail does lead to an interesting culprit for why It is slow though: 

screen shot 2017-03-16 at 18 02 01

I think when we load up Babel or TypeScript then every JS file is sent through babel including all `node_modules` - which could definitely slow it down. This is likely something inside the Jest env, that should be pretty easy to track down.


deecewan:
I'm assuming that this references #169?

orta:
Yeah #169 mainly, I've used up my computer time for today on #174 - but will probably start going over this again in the next few days

orta:
Note to self, [process.execSync](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options) can be used to run a child process synchronously. For our Dangerfile.

orta:
Alright, I think this should be good to go from the typescript support side. I'd like reviews around that.

I'd like to continue working on this PR specifically on two things, making Circle CI use a `dangerfile.ts` and adding a Dangerfile rule about ensuring the d.ts in the repo is completely up to date.

orta:
OK, the circle CI dangerfile is 👍 

Fixes #84 

screen shot 2017-03-19 at 09 09 04

orta:
Our current setup makes it possible for the circle ci danger to pass, and to delete the current message on a PR.

This can be fixed by either https://github.com/danger/danger-js/issues/162 or the Danger ID thing discussed in https://github.com/danger/danger-js/issues/166

orta:
green!

Issue #170, opened 2017-03-15
Github enterprise, 404 when trying to get PR reviews


alex3165:
Is there a way to avoid trying to get the reviews from the API ? I am not using it and my enterprise Github API doesn't have this endpoint. Error message :
```
❯ npm run danger

> lerna-desktop@0.0.0 danger /Users/alexandrer/development/tango/trainline-desktop
> scripts/danger.sh

OK, looks good Fake Testing CI on GitHub
Request failed [404]: https://github.thetrainline.com/api/v3/repos/Tango/trainline-desktop/pulls/752/reviews
Response: {
  "message": "Not Found",
  "documentation_url": "https://developer.github.com/enterprise/2.8/v3"
}
Error:  Error: 404
    at GitHubAPI.<anonymous> (/Users/alexandrer/development/tango/trainline-desktop/node_modules/danger/distribution/api/fetch.js:92:27)
    at step (/Users/alexandrer/development/tango/trainline-desktop/node_modules/danger/distribution/api/fetch.js:32:23)
    at Object.next (/Users/alexandrer/development/tango/trainline-desktop/node_modules/danger/distribution/api/fetch.js:13:53)
    at fulfilled (/Users/alexandrer/development/tango/trainline-desktop/node_modules/danger/distribution/api/fetch.js:4:58)
    at process._tickCallback (internal/process/next_tick.js:103:7)
```

alex3165:
I think https://github.com/danger/danger-js/commit/123cb9d300d5d1210972601ff1de9cb10721b877 is the problem

orta:
This relates to #163 - right now if _any_ API requests fail, then the build fails. However, if GH enterprise doesn't have the new reviews systems, then it'll need to fail and return an empty object/array without failing the build.

Think you can make the error handling move inline?

alex3165:
Hey @orta, I have opened a PR that should improve consistency as discussed in #163 as well.

See: https://github.com/danger/danger-js/pull/173

Issue #163, opened 2017-03-12
Consolidate and apply tests to the GitHub API object


orta:
It's a bit of a mess in there:

some return JSON

```typescript
  async getUserInfo(): Promise<any> {
    const response: any = await this.get("user")
    return await response.json()
  }

async getReviewerRequests(): Promise<any> {
    const repo = this.ciSource.repoSlug
    const prID = this.ciSource.pullRequestID
    const res = await this.get(`repos/${repo}/pulls/${prID}/requested_reviewers`, {
      accept: "application/vnd.github.black-cat-preview+json"
    })
    if (res.ok) {
      return res.json()
    }
    return []
  }
```

some return the fetch call

```typescript

  getPullRequestDiff(): Promise<any> {
    const repo = this.ciSource.repoSlug
    const prID = this.ciSource.pullRequestID
    return this.get(`repos/${repo}/pulls/${prID}`, {
      accept: "application/vnd.github.v3.diff"
    })
  }
```

some return a string

```typescript
  async fileContents(path: string, ref?: string): Promise<string> {
    // Use head of PR (current state of PR) if no ref passed
    if (!ref) {
      const pr = await this.getPullRequestInfo()
      const prJSON = await pr.json()

      ref = prJSON.head.ref
    }
    const fileMetadata = await this.getFileContents(path, ref)
    const data = await fileMetadata.json()
    const buffer = new Buffer(data.content, "base64")
    return buffer.toString()
  }
```

and the names of the functions don't really mean too much. Perhaps the answer is that the the verb is put before, then the return type at the end: turning `getReviewerRequests` into `getReviewerRequestsJSON`. It may make sense to consider that two functions:

```typescript
async getReviewerRequests(): Promise<any> {
    const repo = this.ciSource.repoSlug
    const prID = this.ciSource.pullRequestID
    return await this.get(`repos/${repo}/pulls/${prID}/requested_reviewers`, {
      accept: "application/vnd.github.black-cat-preview+json"
    })
}

async getReviewerRequestsJSON(): Promise<any> {
    const res =  await this.getReviewerRequests()
    if (res.ok) {
      return res.json()
    }
    return []
}
```

deecewan:
my thoughts would be to:
- only return JSON.  The Github API returns JSON for everything.
- error handling should be done inside the API call (I kinda made a move at this by testing `res.ok` in the `reviewerRequests` call

alex3165:
PR up https://github.com/danger/danger-js/pull/173

PR #173, opened 2017-03-16
Don't throw on request error / improve api consistency

alex3165:
This PR improve consistency in `GithubAPI.ts` and avoid throwing error on fetch but only log a warning if something failed.

Fix: https://github.com/danger/danger-js/issues/170
Fix: https://github.com/danger/danger-js/issues/163

### DIFF
--- a/source/ci_source/providers/Fake.ts
+++ b/source/ci_source/providers/Fake.ts
@@ -6,8 +6,8 @@ export class FakeCI implements CISource {
 
   constructor(env: Env) {
     const defaults = {
-      repo: env.DANGER_TEST_REPO || "artsy/emission",
-      pr: env.DANGER_TEST_PR || "327"
+      repo: env.DANGER_TEST_REPO || "artsy/emission", // TODO: default to empty string ?
+      pr: env.DANGER_TEST_PR || "327" // TODO: default to empty string ?
     }
 
     this.env = {...env, ...defaults}
--- a/source/platforms/_tests/GitHub.test.ts
+++ b/source/platforms/_tests/GitHub.test.ts
@@ -14,7 +14,7 @@ const EOL = os.EOL
 const mockGitHubWithGetForPath = (expectedPath): GitHub => {
   const mockSource = new FakeCI({})
 
-  const api = new GitHubAPI("token", mockSource)
+  const api = new GitHubAPI(mockSource)
   const github = new GitHub(api)
 
   api.get = (path: string, headers: any = {}, body: any = {}, method: string = "GET"): Promise<any> => {
@@ -50,7 +50,7 @@ export const requestWithFixturedContent = async (path: string): Promise<any> => 
 describe("with fixtured data", () => {
   it("returns the correct github data", async () => {
     const mockSource = new FakeCI({})
-    const api = new GitHubAPI("token", mockSource)
+    const api = new GitHubAPI(mockSource)
     const github = new GitHub(api)
     api.getPullRequestInfo = await requestWithFixturedJSON("github_pr.json")
     api.getPullRequestCommits = await requestWithFixturedJSON("github_commits.json")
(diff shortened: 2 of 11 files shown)


source/platforms/github/GitHubAPI.ts
```diff
@@ -15,7 +15,7 @@ export class GitHubAPI {
   fetch: typeof fetch
   additionalHeaders: any
 
-  constructor(public readonly token: APIToken | undefined, public readonly ciSource: CISource) {
+  constructor(public readonly ciSource: CISource, public readonly token?: APIToken) {
```
orta: note to self: this will need changing in Peril

source/platforms/github/_tests/_GitHubAPI.test.ts
```diff
@@ -52,23 +51,19 @@ describe("API testing", () => {
...
-        "Content-Type": "application/json",
-      },
-      method: "PATCH",
-    })
+    api.fetch = fetch
+    api.patch = jest.fn(() => ({ json: jest.fn() }))
+
+    await api.updateCommentWithID(123, "Hello!")
+
+    expect(api.patch).toHaveBeenCalledWith("repos/artsy/emission/issues/comments/123", {}, {"body": "Hello!"})
```
orta: this is 🥇

source/platforms/_tests/GitHub.test.ts
```diff
@@ -28,32 +28,21 @@ const mockGitHubWithGetForPath = (expectedPath): GitHub => {
...
-  const content = readFileSync(`${fixtures}/${path}`, {}).toString()
-  return () => {
-    return {
-      text: () => Promise.resolve(content)
-    }
-  }
-}
+export const requestWithFixturedContent = async (path: string): Promise<() => Promise<string>> => () => (
+  Promise.resolve(readFileSync(`${fixtures}/${path}`, {}).toString())
+)
```
orta: nice code-golfing 💃

alex3165:
@orta Is it what you meant: https://github.com/danger/danger-js/pull/173/commits/019fbce18ee02bb0287722669d92f578a4947112 ? Actually I would completely remove the exitCode because the first PR could be the `reviews` PR potentially right ? Why debugging would be difficult without exitCode ? (we would still have the console.warn to help right ?)

alex3165:
Could we maybe set `exitCode = 1` and throw an error if all the request fails ?

orta:
The order of the requests is pretty much deterministic, it's handled by `async getPlatformDSLRepresentation(): Promise<GitHubDSL>` so I think the simplest thing to do is to make that look something like this:

```js
async getPlatformDSLRepresentation(): Promise<GitHubDSL> {
    const pr = await this.getReviewInfo()
    if (pr === {}) {
      process.exitCode = 1
      throw "Could not find pull request information, if you are using a private repo then perhaps Danger does not have permission to access that repo."
    }
    const issue = await this.getIssue()
    const commits = await this.api.getPullRequestCommits()
    ...
}
```

Does that make sense?

alex3165:
@orta Make total sense 👍 I did the update

alex3165:
What is happening with the CI ? I don't see why it is failing ?

macklinu:
For some reason, Travis tried to run our [Node 7 build with Node 0.10.36](https://travis-ci.org/danger/danger-js/jobs/212043294#L181) - not sure why. The Node 6 Travis build is green, though. 👍

orta:
I just hit rebuild, lets see if it was a temporary thing

orta:
Alright - yep

source/ci_source/providers/Fake.ts
```diff
@@ -6,8 +6,8 @@ export class FakeCI implements CISource {
 
   constructor(env: Env) {
     const defaults = {
-      repo: env.DANGER_TEST_REPO || "artsy/emission",
-      pr: env.DANGER_TEST_PR || "327"
+      repo: env.DANGER_TEST_REPO || "artsy/emission", // TODO: default to empty string ?
+      pr: env.DANGER_TEST_PR || "327" // TODO: default to empty string ?
```
orta: Hrm, I generally like to have test data that is real for my desalts during dev - I'm still OK with keeping these as this repo mainly because it keeps it consistent with all the fixture data that is inside the tests

orta:
👍  thanks @alex3165

PR #174, opened 2017-03-18
Create a utils function for GitHub links

orta:
This builds on #134 to support easily making a list of GitHub clickable URLs - with this we can remove a great big chunk of [Jest's](https://github.com/facebook/jest/blob/master/dangerfile.js#L20-L39) and [Emissions'](https://github.com/artsy/emission/pull/457) Dangerfiles.

🌮

### DIFF
--- a/source/dsl/DangerUtilsDSL.ts
+++ b/source/dsl/DangerUtilsDSL.ts
@@ -5,7 +5,7 @@
 export interface DangerUtilsDSL {
 
   /**
-   * Creates an HTML link.
+   * Creates a link using HTML.
    *
    * If `href` and `text` are falsy, null is returned.
    * If `href` is falsy and `text` is truthy, `text` is returned.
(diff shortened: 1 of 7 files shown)


source/dsl/DangerUtilsDSL.ts
```diff
@@ -5,7 +5,7 @@
 export interface DangerUtilsDSL {
 
   /**
-   * Creates an HTML link.
+   * Creates a HTML link.
```
deecewan: ooooh. controversial.  I guess this depends on how you pronounce 'H'.  I say 'aitch', so 'an' makes sense.  I guess it makes sense if you say 'haitch', or pronounce 'HTML' in some other way.

orta: hah - great point, I just asked my wife, she also agrees with you - instead, I'll re-write the sentence :D

source/dsl/DangerUtilsDSL.ts
```diff
@@ -5,7 +5,7 @@
 export interface DangerUtilsDSL {
 
   /**
-   * Creates an HTML link.
+   * Creates a link using HTML.
```
orta: The previous one is not wrong, but if I already tried fixing it, I bet we'd get people in the future doing the same 👯‍♂️

PR #183, opened 2017-03-21
Support Text only output + improve wording throughout Danger

orta:
In #172 I let danger run separately on circle ci. This could delete an existing comment by travis, which is actually doing real work. So this fixes that, and improves the verbose option.

I also tried to run through all of the times when Danger outputs text to the screen to improve the English and provide more info when I can.


orta:
I want some of this stuff for #184

Issue #175, opened 2017-03-18
API idea: JSON Diff


orta:
We care a lot about JSON files in the JS community, so pretty often I think to myself "hrm, I'd like to do a diff to know whether the package version had changes, or dependencies" - it's pretty tough to do 

It's definitely a bunch of code in a Dangerfile, that so far I've been hesitant to write. Perhaps it could be a new function inside the `danger.utils` class?

Here's some examples of what I'd imagine this could look like:

Promise vs callback API is worth a discussion, for my vote TBH I'm a promises, mainly because I use `async/await` but for now I'll just write a callback

```js
danger.utils.jsonDiff("package.json", (diff) => {
  // diff could look something like
  // {
  //   "version" : {
  //     "before": 0.0.5,
  //     "after": 0.0.6
  //   }
  // }
  if (diff && diff.version && diff.version.before < diff.version.after) {
    fail("No retreat, no surrender.")
  }
}) 
```

The nullability aspect is interesting, initially I thought of passing `null` instead of an empty object to the params, but maybe it makes more sense to just not run the closure at all if there are no changes?

orta:
Could have a much more fine grained check so that things like this don't happen

<img width="808" alt="screen shot 2017-03-18 at 17 08 42" src="https://cloud.githubusercontent.com/assets/49038/24070292/99f13962-0bfd-11e7-8db3-7de6b1513df4.png">

orta:
That could look more like

```js
danger.utils.jsonDiff("package.json", (diff) => {
  const depChanges = diff.dependencies || diff.devDependencies
  const lockfileChanges = danger.git.modified_files.include("yarn.lock")
  if (depChanges && !lockfileChanges) {
    fail("No lock file changes - yada yadda")
  }
}) 
```

Or you can start picking out the specific deps that are changed and do something like http://pkgsize.com on those locally

orta:
I bet @wtgtybhertgeghgtwtg would love that have that on each dep changes 💯

wtgtybhertgeghgtwtg:
I hadn't realized that I had been making a name for myself.
The implementation might not be that complex.
1. Generate a JSON Patch from the original `package.json` to the proposed `package.json` (only `add`, `replace`, and `remove`, one entry per key).
2. Iterate over the patch, expand the key and create an object per `path`.
  a. If `op` is `add`, `before` is undefined, `after` is `value`.
  b. If `op` is `replace`, `before` is the value from the original `package.json`, `after` is `value`.
[...]

PR #184, opened 2017-03-21
Adds Json patch APIs

orta:
This turned out to be a much bigger PR than I expected. This addresses #175 

It makes a few more of our transitive dependencies explicit, and adds a new one around the JSON patch rfc6902. I wrote an extensive CHANGELOG, so here is that

---

* Added a two new `git` DSL functions: `git.JSONDiffForFile(filename)` and `git.JSONPatchForFile(filename)`.

* `git.JSONPatchForFile`
  
  This will generate a rfc6902 JSON patch between two files inside your repo. These patch files are useful as a standard, but are pretty tricky to work with in something like a Dangerfile, where rule terseness takes priority.

* `git.JSONDiffForFile`
  
  This uses `JSONPatchForFile` to generate an object that represents all changes inside a Dangerfile as a single object, with keys for the changed paths. For example with a change like this:

  ```diff
  {
    "dependencies": {
      "babel-polyfill": "^6.20.0",
    +  "chalk": "^1.1.1",
      "commander": "^2.9.0",
      "debug": "^2.6.0"
    },
  }
  ```

  You could become aware of what has changed with a Dangerfile like:

  ```js
  const packageDiff = await git.JSONDiffForFile("package.json")
  if (packageDiff.dependencies) {
    const deps = packageDiff.dependencies

    deps.added   // ["chalk"],
    deps.removed // []
    deps.after   // { "babel-polyfill": "^6.20.0", "chalk": "^1.1.1", "commander": "^2.9.0", "debug": "^2.6.0" }
    deps.before  // { "babel-polyfill": "^6.20.0", "commander": "^2.9.0", "debug": "^2.6.0" }
  }
  ```
  The keys: `added` and `removed` only exist on the object if:
  
  * `before` and `after` are both objects - in which case `added` and `removed` are the added or removed keys
  * `before` and `after` are both arrays - in which case `added` and `removed` are the added or removed values


orta:
Looks like travis is posting responses, then circle is deleting them, will be fixed on the next production Danger release.

macklinu:
This is awesome! 💟

PR #188, opened 2017-03-21
Switch circle to use the new text-only output

orta:
Should stop the disappearing messages.


orta:
I'm happy to not have a CHANGELOG too.

PR #190, opened 2017-03-21
Update TS definitions to work correctly on 3rd party repos

orta:
I was seeing errors around `Readonly`, and it looks like the `declare module` part isn't needed for TypeScript.


orta:
screen shot 2017-03-21 at 10 05 14

PR #198, opened 2017-03-23
Updated Jest devDependencies

delianides:
The mismatch of Jest dependencies causes jest to fail on new projects using danger. Danger runs, but Jest fails with:

```
yarn test v0.19.1
$ jest
 FAIL  src/index.spec.js
  ● Test suite failed to run

    TypeError: Cannot read property 'bind' of undefined

      at Runtime._createRuntimeFor (node_modules/jest-cli/node_modules/jest-runtime/build/index.js:640:37)
      at process._tickCallback (internal/process/next_tick.js:109:7)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        0.03s, estimated 1s
Ran all test suites.
error Command failed with exit code 1.
```

Probably related to https://github.com/facebook/jest/issues/3114


orta:
May need to also update ts-jest

orta:
( I'm now heading out for the night, so can't be of too much immediate advice alas)

orta:
Alright, that was a bit more nuanced than I expected, jest-config used to return an object like `{ key: value }` but now it returns `{ config: {key : value } }` - this now inside #194 

So thanks for the PR and the heads up @delianides - appreciated

Issue #200, opened 2017-03-24
Debate: Should we include matchers?


orta:
I had a fail based on two long strings which I fixed this way:

![screen shot 2017-03-24 at 10 13 05](https://cloud.githubusercontent.com/assets/49038/24289751/82805d18-107a-11e7-8222-9fafcc19d197.png)

After putting a `console.log` in the dangerfile. This ain't good.

---

Code looked like this:

```ts
import dtsGenerator from "./scripts/danger-dts"
const currentDTS = dtsGenerator()
const savedDTS = fs.readFileSync("source/danger.d.ts").toString()
if (currentDTS !== savedDTS) {
  const message = "There are changes to the Danger DSL which are not reflected in the current danger.d.ts."
  const idea = "Please run <code>yarn declarations</code> and update this PR."
  fail(`${message}<br/><i>${idea}</i>`)
}
```

We already have `jest-diff` inside the app, due to `jest-config`, so what about instead we provide something like:

```ts
import dtsGenerator from "./scripts/danger-dts"
const currentDTS = dtsGenerator()
const savedDTS = fs.readFileSync("source/danger.d.ts").toString()

const message = "There are changes to the Danger DSL which are not reflected in the current danger.d.ts."
const idea = "Please run <code>yarn declarations</code> and update this PR."
const failureMessage =`${message}<br/><i>${idea}</i>`

danger.utils.checkEqual(fail, failureMessage, currentDTS, savedDTS)
```

Where we can present jest-matcher like messages inline. For example:

![screen shot 2017-03-24 at 10 16 57](https://cloud.githubusercontent.com/assets/49038/24289944/1a26f1b8-107b-11e7-84e6-27fc73fa3455.png)

or 

![screen shot 2017-03-24 at 10 18 26](https://cloud.githubusercontent.com/assets/49038/24289987/42d446ec-107b-11e7-82bf-a7459e923f08.png)

orta:
Messages could look like:

<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Warnings</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>There are changes to the Danger DSL which are not reflected in the current danger.d.ts.<br><I>Please run <code>yarn declarations</code> and update this PR.</I></td>
    </tr>
  </tbody>
</table>

```diff
Expected:
[...]

PR #195, opened 2017-03-22
Fixes a typo

steipete:
Just a small typo. Love Danger! Blog post coming up soon (Ruby version, but still)

PR #204, opened 2017-03-29
Run yarn and yarn declarations #trivial

macklinu:
I ran `yarn declarations` to update the `danger.d.ts` file with the typo fix from https://github.com/danger/danger-js/pull/195. Greenkeeper PRs like https://github.com/danger/danger-js/pull/202 are "failing" because of the `yarn declarations` check in our dangerfile. (I also think https://github.com/danger/danger-js/issues/200 relates to this.)

Also, when I just ran `yarn` on master, it updated the `yarn.lock` file, which I think more appropriately reflects our `package.json` (for example, `yarn.lock` now shows ts-node 3.x, which is also present in our `package.json`).

Once this is merged, I can merge master back into #201, #202, and #203 to get them passing. 👍

### DIFF
--- a/source/danger.d.ts
+++ b/source/danger.d.ts
@@ -1,6 +1,6 @@
 export type MarkdownString = string
 
-/** A platform agnostic refernce to a Git commit */
+/** A platform agnostic reference to a Git commit */
 export interface GitCommit {
   /** The SHA for the commit */
   sha: string,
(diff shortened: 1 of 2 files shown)


orta:
Ah! Yeah. My bad, I expected merging https://github.com/danger/danger-js/pull/194 to take less time than it did. Basically this https://twitter.com/orta/status/846083534909530115

macklinu:
No worries, @orta. Thanks for the 👍

PR #205, opened 2017-03-30
Fix Lodash isArray deprecation

damassi:
When installing the latest danger, I noticed this in the console output: 

screen shot 2017-03-29 at 5 15 32 pm

Replaces `lodash.isarray` with native implementation.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,10 +4,11 @@
 
 ### 0.14.2
 
-* Moved `@types/chalk` from dependencies to devDependencies - orta
+* Replaced deprecated `lodash.isarray` package with `Array.isArray` - damassi
 
 ### 0.14.1
 
+* Moved `@types/chalk` from dependencies to devDependencies - orta
 * Killed some stray console logs - orta
 * Updated the danger.d.ts - orta
 
@@ -18,7 +19,7 @@
   We use TypeScript in Danger, and a lot of my work in Artsy now uses TypeScript (see: [JS2017 at Artsy](http://artsy.github.io/blog/2017/02/05/Front-end-JavaScript-at-Artsy-2017/#TypeScrip1t)), so I wanted to
   explore using TypeScript in Dangerfiles.
 
-  This is built on top of Jest's custom transformers, so if you are already using Jest with TypeScript, then 
+  This is built on top of Jest's custom transformers, so if you are already using Jest with TypeScript, then
   you can change the `dangerfile.js` to `dangerfile.ts` and nothing should need changing ( except that you might have
   new warnings/errors ) (_note:_ in changing this for Danger, I had to also add the `dangerfile.ts` to the `"exclude"`
   section of the `tsconfig.json` so that it didn't change the project's root folder.)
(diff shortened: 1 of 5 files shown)


source/platforms/github/GitHubGit.ts
```diff
@@ -114,7 +113,7 @@ export default async function gitDSLForGitHub(api: GitHubAPI): Promise<GitDSL> {
       // added or removed. This makes it really easy to act on specific
       // changes to JSON DSLs
 
-      if (isarray(diff.after) && isarray(diff.before)) {
+      if (Array.isArray(diff.after) && Array.isArray(diff.before)) {
```
macklinu: I think we should be safe using the native `Array.isArray()` instead of that deprecated lodash package. 👍

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "danger",
-  "version": "0.14.1",
+  "version": "0.14.2",
```
macklinu: Could you revert this change please? Unless we are good to ship this in a patch release - cc @orta

changelog.md
```diff
@@ -4,10 +4,11 @@
 
 ### 0.14.2
 
-* Moved `@types/chalk` from dependencies to devDependencies - orta
+* Replaced deprecated `lodash.isarray` package with `Array.isArray` - damassi
 
 ### 0.14.1
 
+* Moved `@types/chalk` from dependencies to devDependencies - orta
```
macklinu: Good call, I think this was shipped in 0.14.1.

changelog.md
```diff
@@ -4,10 +4,11 @@
 
 ### 0.14.2
 
-* Moved `@types/chalk` from dependencies to devDependencies - orta
+* Replaced deprecated `lodash.isarray` package with `Array.isArray` - damassi
```
macklinu: Looking for confirmation from @orta on shipping 0.14.2 or holding off on it for anything else to keep this changelog message here. Otherwise we usually add changelog messages right below the `//  Add your own contribution below` line and then add the version number we are shipping to the changelog as we package up a release.

macklinu:
Forgot to add, Danger commented on the PR requesting that `yarn install` is run (for dependency additions/removals to keep `yarn.lock` in sync with `package.json`). Could you please run that locally and push up a commit with the changes to `yarn.lock`? 🙂

changelog.md
damassi: Ah, I see -- `danger` said to update the changelog and so I followed the format as it was there. It might be good to provide a little more direction in terms of version numbers, changelog, etc.

damassi:
Updated and rebased. It's odd -- I ran a `yarn install` first thing after pulling down the package, but I had to `rm` it and reinstall packages before it would update.

orta:
Ah yeah, the version/changelog thing - again sorry about that, :D

I got stuck on a particularly long PR, which I thought I could get finished by the weekend ( https://github.com/danger/danger-js/pull/194 ) but inevitably I couldn't.

IMO it makes sense to be more verbose in the changelog about saying above the latest released version. 👍

orta:
Shipping `0.14.2` sounds 👍

orta:
@damassi with respect to the lockfile,  you would have needed to run `yarn remove lodash-isarray` this would have removed the module then only re-created the dependency locks for anything related to it.

In deleting and re-creating it from scratch you've updated hundreds of dependencies in the stack. 

I'm ok with this, after eyeing it over, but ideally next time we should be a bit more strict about smaller lock file changes.

orta:
Alright, I've shipped that - and all other PRs as 0.14.2 

I've also updated the notice on the CHANGELOG to be more specific in https://github.com/danger/danger-js/commit/fde4c54d4e01d4f34921aec43624780b2410458e

damassi:
Ah sorry about that! I wasn't aware of `yarn remove <package>`. Noted 👍

Issue #185, opened 2017-03-21
If Dangerfile fails to eval, post a message on the PR with a stack trace.


orta:
Just because the eval part fails doesn't mean we can't provide meaningful errors.

E.g. a Dangerfile like this:

```js
const a = 1
a = 2
```

should already be failing the build, but we are probably able to post to the CI provider, so we can provide feedback (and recommend `danger pr [pr_url]` at the end too.

macklinu:
That's a great idea - definitely worth adding. 👍

PR #210, opened 2017-04-07
Post to the platform when evaling a Dangerfile fails

orta:
Fixes https://github.com/danger/danger-js/issues/185


.vscode/settings.json
```diff
@@ -15,5 +15,22 @@
     },
     "files.associations": {
         "**/*.js": "javascriptreact"
-     }
+    },
```
orta: my spellchecker got an update

PR #213, opened 2017-04-09
Initial work at creating docs

orta:
I want to keep the guides inside this repo, then have danger.systems import them.

This PR does two things:

* Creates a structure for guides in markdown
* Creates a proof-of-concept for generating the DSL as a JSON file via typedoc


orta:
Paired with https://gitlab.com/danger-systems/danger.systems/merge_requests/141

orta:
OK, this isn't a one big PR job. So, this is good to go as is.

PR #214, opened 2017-04-12
Improve Github dsl test

alex3165:
This PR improve the way `GitHub.ts` is tested by mocking `GithubAPI` class so that it is using mock data instead of doing call to API.

What do you think about it ? If the approach seems good for you I can write the missing tests, it should be straightforward now to add any test on top of this.

### DIFF
--- a/source/platforms/_tests/_github.test.ts
+++ b/source/platforms/_tests/_github.test.ts
@@ -47,7 +47,6 @@ describe("getPlatformDSLRepresentation", () => {
 
   it("should get the issue label", async() => {
     const issue = await github.getIssue()
-    console.log(issue)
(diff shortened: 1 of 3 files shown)


source/platforms/_tests/_github.test.ts
```diff
@@ -1,35 +1,56 @@
+import { resolve } from "path"
+import { readFileSync } from "fs"
+
+const fixtures = resolve(__dirname, "fixtures")
+
+/** Returns JSON from the fixtured dir */
+export const requestWithFixturedJSON = async (path: string): Promise<() => Promise<any>> => () => (
```
orta: hah, this type annotation at the end! :D

PR #215, opened 2017-04-12
Add filter per type to gitdiff

alex3165:
On my project I only want to check for the changes that have been added to the project, this support any kind of specified type (normal, add, delete)


orta:
Yeah, this makes sense. I'll add some docs.

PR #223, opened 2017-04-13
Add convenience info to `danger.git.diffForFile`

### LINKED ISSUE #221: Use a structured format for textual diffs?
Currently when using `danger.git.JSONDiffForFile`, you get a data structure like this:

```ts
/** All JSON diff values will be this shape */
export interface JSONDiffValue {
  /** The value before the PR's applied changes */
  before: any
  /** The value after the PR's applied changes */
  after: any
  /** If both before & after are arrays, then you optionally get what is added. Emprty is no additional objects. */
  added?: any[],
  /** If both before & after are arrays, then you optionally get what is removed. Emprty is no removed objects. */
  removed?: any[]
}
```

However, `danger.git.diffForFile` returns a simple `diff` string:

```diff
     "chai": "3.5.0",
     "chokidar": "1.6.0",
     "chromedriver": "2.28.0",
+    "danger": "0.16.0",
     "docker-lambda": "0.9.1",
     "electron": "1.4.4",
     "enzyme": "2.8.0",
```

I'd propose changing the result to match the shape of `JSONDiffValue`:

```ts
/** All Text diff values will be this shape */
export interface TextDiffValue {
  /** The value before the PR's applied changes */
  before: string
  /** The value after the PR's applied changes */
  after: string,
  added?: string[],
  removed?: string[],
  normal?: string[],
  diff: string,
}
```

As it is currently, you'd need to `fs.readFile` the contents of the changed file, and apply the reverse of the `diff` provided by `danger.git.diffForFile` to get the `before` and `after` values of a file. It's doable, but not pretty.

Thoughts?

namuol:
Fixes #221

Changes the signature for `danger.git.diffForFile` to the following:

```ts
/** All Text diff values will be this shape */
export interface TextDiff {
  /** The value before the PR's applied changes */
  before: string
  /** The value after the PR's applied changes */
  after: string,
  /** A string containing the full set of changes */
  diff: string,
  /** A string containing just the added lines */
  added: string,
  /** A string containing just the removed lines */
  removed: string
}

diffForFile(filename: string): Promise<TextDiff | null>
```

### TODO

- [x] Update docs
- [x] Update changelog (include dangerfile migration instructions)
- [x] Run `yarn declarations`
- [x] Add tests
  - [x] `before`
  - [x] `after`
  - [x] `added`
  - [x] `removed`
- [x] Fix snapshot issues ([?](https://github.com/namuol/danger-js/blob/82b089c980f213c874642a1b2ce2c530a31a2bfd/source/platforms/github/_tests/_github_git.test.ts#L22-L29))


source/dsl/GitDSL.ts
```diff
@@ -64,8 +76,7 @@ export interface GitDSL {
    * @param {string} filename the path to the json file
    * @param {string[]} diffTypes An array containing: "add", "del" or "normal" - then you can only get these changes
    */
-
-  diffForFile(filename: string, diffTypes?: string[]): string | null,
+  diffForFile(filename: string, diffTypes?: string[]): Promise<TextDiff | null>,
```
orta: think you can kill `diffTypes` in favour of this /cc @alex3165

source/platforms/github/GitHubGit.ts
```diff
@@ -138,22 +138,33 @@ export default async function gitDSLForGitHub(api: GitHubAPI): Promise<GitDSL> {
...
+
+    return {
+      before: baseFile,
+      after: headFile,
+      added: changes.filter(({type}: {type: string}) => type === "add"),
+      removed: changes.filter(({type}: {type: string}) => type === "del"),
+      normal: changes.filter(({type}: {type: string}) => type === "normal"),
+      diff: lines.join(os.EOL)
+    }
   }
```
orta: 👍

source/dsl/GitDSL.ts
namuol: Should we still keep the `added`/`removed`/`normal` changes? They're not really the same as `JSONDiff`.

orta: Can't  they be the raw string content of all the hunks combined?

source/platforms/github/GitHubGit.ts
```diff
@@ -138,22 +138,33 @@ export default async function gitDSLForGitHub(api: GitHubAPI): Promise<GitDSL> {
...
-                             .reduce((a: any, b: any) => a.concat(b), [])
+    let changes = structuredDiff.chunks.map((c: any) => c.changes)
+                                .reduce((a: any, b: any) => a.concat(b), [])
 
     // Filter the diffs by type
     if (diffTypes && diffTypes.length > 0) {
       changes = changes.filter((c: any) => diffTypes.indexOf(c.type) !== -1)
     }
 
+    const baseFile = await api.fileContents(filename, pr.base.repo.full_name, pr.base.sha)
```
namuol: Note: I'm using `sha` here instead of `ref`, mainly because the example PR used for the fixture seemed to have deleted the original branch.

We may also want to use `sha` above in the implementation of `JSONPatchForFile`.

source/dsl/GitDSL.ts
namuol: Ah, yes that does make sense. @alex3165 does this satisfy your use-case?

i.e. instead of calling `diffForFile('foo.js', ['add'])`, you'd call `diffForFile('foo.js').added` which would return the same sort of string.

source/platforms/github/GitHubGit.ts
orta: no difference IMO 👍

orta: `she` may be better incase a potential branch has had new commits since the CI was started, and then when danger runs

source/dsl/GitDSL.ts
namuol: I honestly am confused about the meaning of `normal` in this context -- I _think_ it refers to "contextual" lines (i.e. non-changed lines that sandwich an addition/deletion). I have a hard time finding why it'd be useful to have these separately without the benefit of combining them with either `added` or `removed`.

Maybe we could just provide the raw `changes` structure from `parseDiff`'s output as an escape hatch for folks who need finer control.

orta: maybe then added and removed should be `add + normal` and ``del`?

namuol: Here's what I think would be more useful for most folks:

```ts
/** All Text diff values will be this shape */

/** All Text diff values will be this shape */
export interface TextDiff {
  /** The value before the PR's applied changes */
  before: string
  /** The value after the PR's applied changes */
  after: string,
  /** A string containing the full set of changes */
  diff: string,
  /** A string containing just the added lines */
  added: string,
  /** A string containing just the removed lines */
  removed: string,
}
```

This covers all cases except for the odd one where you only want `normal`/contextual lines, which seems quite fringe to me.

orta: Yeah, that feels 👍 to me

namuol:
Hmm... I must be missing something. The tests are passing on my machine but AppVeyor doesn't seem to like it. 🤷‍♂️

source/platforms/github/_tests/_github_git.test.ts
```diff
@@ -20,15 +20,19 @@ export const requestWithFixturedContent = async (path: string): Promise<() => Pr
   Promise.resolve(readFileSync(`${fixtures}/${path}`, {}).toString())
 )
 
+const pullRequestInfoFilename = "github_pr.json"
+const masterSHA = JSON.parse(readFileSync(`${fixtures}/${pullRequestInfoFilename}`, {}).toString()).base.sha
```
orta: try using the path API for this string - that might be what's breaking on windows, can't remember its name yet

namuol: This could possibly be a bug with Jest's snapshot testing. :-/ Their Windows support is still pretty unstable AFAIK.

namuol: I thought maybe it had something to do with line endings, but that shouldn't be an issue: https://github.com/facebook/jest/pull/1879

orta:
These failing snapshots feel backwards, like it's inverting the before and after

namuol:
> These failing snapshots feel backwards, like it's inverting the before and after

Yeah, the Jest output doesn't really make sense. It treats every line but the last as different, even though if you copy/paste the terminal output and do a diff after cleaning out the additional `+`/`-`s, they're identical.

orta:
yeah, must be some invisible characters - can we strip them for the test?

namuol:
Yup. Stripping the whitespace seemed to fix the issue. I'll make this more explicit and add a `TODO` to the test file.

orta:
👍 Perfect

namuol:
I'd wait for @alex3165's response, but otherwise does this look good?

orta:
Yep, everything looks 👍

orta:
will merge and release EOD tomorrow if we don't hear anything back

orta:
also, I've invited you to the org, as a part of [Moya Community Continuity](https://github.com/Moya/contributors/issues/6) - you're welcome to participate (or not, we're chill) in PRs and help out at a level that works for you.

PR #227, opened 2017-04-17
Exposing a GitHub API to users of Danger

### LINKED ISSUE #219: Add an API for writing to labels (/GitHub API DSL)
We should add the ability to write to the labels of an issue

```js
if (danger.pr.body.includes("JIRA:") {
  danger.github.api.add_label("has-ticket")
}
```

I have been a bit hesitant to add a GitHub API client, but maybe this is the time to do that. I've been using the  [github](https://www.npmjs.com/package/github) module in my scripts and it's well made with good types. We could expose this (separate from the internal GitHub API client) .

orta:
Fixes #219 

In #219 I wondered whether we should just expose an API instead of writing one-off APIs like write labels. I've decided that it makes sense to do the full API. I used this module for some scripting and it felt pretty good - well typed, and comprehensively put together. Will make life easier in Peril too.

That said, I have _no idea_ why this is failing, but figured I'd ship it for now.

---

Also, as I now have a new dependency, I'll clean up the dependency checker in the current Dangerfile.

### DIFF
--- a/source/platforms/_tests/_github.test.ts
+++ b/source/platforms/_tests/_github.test.ts
@@ -34,6 +34,10 @@ jest.mock("../github/GitHubAPI", () => {
       const fixtures = await requestWithFixturedJSON("requested_reviewers.json")
       return await fixtures()
     }
+
+    getExternalAPI() {
+      return {}
+    }
   }
 
   return { GitHubAPI }
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,7 @@
 
 ### Master
 
+* Adds `github.api` more to come - @orta
 * [Enhancements to `danger.git.diffForFile()`](https://github.com/danger/danger-js/pull/223) - @namuol
 
   * Removed `diffTypes` second argument in favor of `result.added` and `result.removed`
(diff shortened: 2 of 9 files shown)


source/platforms/_tests/_github.test.ts
```diff
@@ -34,6 +34,10 @@ jest.mock("../github/GitHubAPI", () => {
       const fixtures = await requestWithFixturedJSON("requested_reviewers.json")
       return await fixtures()
     }
+
+    getExternalAPI() {
```
macklinu: @orta since our tests mock the `GitHubAPI` file, we need to supply the `getExternalAPI()` function here for the tests to pass. 👍

macklinu:
Pushed up a merge commit to resolve the conflict in the Changelog. @orta, this PR looks good to me - anything else to do before merging? Will squash and merge into one commit if that's cool.

image

source/platforms/_tests/_github.test.ts
orta: :D

orta:
Nah - this was all. I'll do a docs run separately 👍

orta:
You're welcome to merge - thanks

orta:
Been focusing on the site

Issue #224, opened 2017-04-14
danger pr failing in windows


tivac:
```bash
> node -v
v7.7.3
> npm ls danger
mithril@1.0.1 C:\Users\Patrick\Documents\Code\mithril.js
`-- danger@0.16.0
> set DEBUG=*
> npm-run danger pr --verbose --repl https://github.com/tivac/mithril.js/pull/3

 danger:runner argv: C:\Program Files\nodejs\node.exe,C:\Users\Patrick\Documents\Code\mithril.js\node_modules\danger\distribution\commands\danger.js,pr,--verbose,--repl,https://github.com/tivac/mithril.js/pull/3 +0ms
  danger:pr executing dangerfile at dangerfile.js +0ms
```

and then it never goes anywhere. I traced it through the code as far as a I could, it makes it to [this line](https://github.com/danger/danger-js/blob/e670d43c29994da05f57260a9b8d3d30ee31935e/source/runner/DangerfileRunner.ts#L35) in `DangerfileRunner.ts` and never continues. The process just stops after a bit.

`--verbose` and `--repl` do nothing since it presumably is dying without continuing.

orta:
ouch, interesting, this came from an update to the jest dependencies https://github.com/danger/danger-js/commit/aea102d7706e26952b6d18e4e024aa3175a3952f ( in 0.14.2 I started seeing issues with the tests around this, so it can be re-produced ) 

Perhaps it could be this? https://github.com/facebook/jest/issues/3079

I wonder if we can try set the no watchman

tivac:
I'll try to dig into jest and see where it falls apart I guess.

tivac:
I changed my local copy of `./node_modules/danger` to set the `useWatchman` option to false when creating the `HasteMap`, no luck.

As far as I can tell the transpiled async function never returns the constructed `HasteMap`, for some reason. I verified that the `HasteMap` constructor *is* running, but output gets lost after that.

It's been very difficult to get a debugger able to actually step through any of this though 😞

tivac:
This is still broken in `0.18.0`, fwiw.

orta:
Aye, aye, definitely still a blocker on 1.0 - current guess is for me to go through the exact options Jest sets to the haste map and see where danger differentiates.

orta:
Also, it's possible that jest 20 comes out and fixes it :D

PR #248, opened 2017-05-07
Check windows support

orta:
This runs the formatter on `source/runner/_tests/_danger_runner.test.ts` as well as undoing the Windows check for the danger eval tests.

### DIFF
--- a/source/runner/DangerfileRunner.ts
+++ b/source/runner/DangerfileRunner.ts
@@ -31,11 +31,14 @@ export async function createDangerfileRuntimeEnvironment(dangerfileContext: Dang
   }
 
   // Setup a runtime environment
-  const hasteConfig = { automock: false, maxWorkers: 1, resetCache: false }
+  const hasteConfig = { maxWorkers: 4, resetCache: false, watch: false, watchman: false }
+  console.log(0)
   const hasteMap = await Runtime.createHasteMap(config, hasteConfig).build()
+  console.log(1)
   const resolver = Runtime.createResolver(config, hasteMap.moduleMap)
+  console.log(2)
   const runtime = new Runtime(config, environment, resolver)
-
+  console.log(3)
   return {
     context,
     environment,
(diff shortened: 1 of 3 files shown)


macklinu:
Is this related to #224?

orta:
Yes, I know that updating Jest broke our windows support (something about the haste map generation) and so maybe updating it one more time would fix it - it didn't, now it requires more digging 👍

tivac:
I'm not really able to dig into this too much more, but I'm happy to help QA it once you think you've got a fix.

It's something, hopefully. 🤔

orta:
Aye, right now my plan is to try understand what options go into the map when jest does it, vs when Danger does

macklinu:
I'm not familiar with the inner workings of Jest, but I was poking around the Jest codebase to see how they create a runtime vs. how Danger does it. I came across [these lines](https://github.com/facebook/jest/blob/9b157c3a7c325c3971b2aabbe4c8ab4ce0b0c56d/packages/jest-runtime/src/cli/index.js#L66-L82) in the `jest.run()` function, which is a different way to create a runtime environment than what [we are currently doing](https://github.com/danger/danger-js/blob/0.19.0/source/runner/DangerfileRunner.ts#L18-L44) in the `DangerfileRunner.ts` file.

To translate what happens in `jest.run()` to `DangerfileRunner.ts`, I think it would look like this:

```diff
diff --git a/source/runner/DangerfileRunner.ts b/source/runner/DangerfileRunner.ts
index a4467e8..c8985fe 100644
--- a/source/runner/DangerfileRunner.ts
+++ b/source/runner/DangerfileRunner.ts
@@ -31,10 +31,10 @@ export async function createDangerfileRuntimeEnvironment(dangerfileContext: Dang
   }
 
   // Setup a runtime environment
-  const hasteConfig = { automock: false, maxWorkers: 1, resetCache: false }
-  const hasteMap = await Runtime.createHasteMap(config, hasteConfig).build()
-  const resolver = Runtime.createResolver(config, hasteMap.moduleMap)
-  const runtime = new Runtime(config, environment, resolver)
+  const hasteMap = await Runtime.createContext(config, {
+    maxWorkers: os.cpus().length - 1,
+  })
+  const runtime = new Runtime(config, environment, hasteMap.resolver)
 
   return {
     context,
```

I tried this out locally on my Macbook, and the tests pass, but I'm not sure what the ramifications would be for Danger or if it would help with fixing #224.

I don't have a Windows machine to test on, but I'm glad to help with this however else I can. 👍

orta:
I don't have a windows machine, but the appveyor tests have been enough to ensure that it continues working - but that doesn't seem to have been enough alas

orta:
Deprecated by https://github.com/danger/danger-js/issues/273

PR #270, opened 2017-06-07
Fix danger to lookup comment for its own only

### LINKED ISSUE #269: unique identifier for danger-written comments?
Due to private repo's limitation, I am currently using myself's github token to run danger instance (which post comments under my user). Which makes if I leave a comment before danger finishes run, danger will remove those comment and replace it into danger's results. 

Is there way to make danger validate comments (like having some prefix in comments?) to avoid these cases?

kwonoj:
- closes #269 .

This PR updates logics of `getDangerCommentID`, in addition to look up userid of given danger token but also looks for comment body if it has comment signature written via danger. For cases of user shares token for danger runner, it'll prevent comment is being deleted by danger looks up comment not written by itself.

Instead of introducing new hidden comment identifier, I took simple paths of looking signature directly (including HTML tags), which is sufficiently unique that individual user won't write those in their own comment.


orta:
This looks 👍 to me too - thanks @kwonoj 

Bit surprised by the red, just re-ran it - should be fine if not, I'll throw an hour or two at danger tomorrow morning to try get these builds determinate

kwonoj:
@orta seems it's continuously failing on travis (https://travis-ci.org/danger/danger-js/builds) by looking history, and log says `jest-cli` is missing, yarn does install pkg some way unexpected?

orta:
Alright, well I'll merge this for now as it's not due to your PR 👍

PR #273, opened 2017-06-09
Supply empty haste modulemap for jest runtime resolver

### LINKED ISSUE #224: danger pr failing in windows

kwonoj:
- closes #224.

This PR updates bootstrap logic for jest runtime, instead of acquiring haste module map by building whole just supply empty modulemap directly. This approach works by danger config doesn't provide `config.roots` , so there's no hastemap's actually generated anyway.

For original reason why it wasn't work in windows platform, refer https://github.com/facebook/jest/pull/3776 - if those PR's accepted and if you prefer to preserve existing logic, bump up jest would resolve issue as well.


orta:
damn, the solution was to remove a bunch of code? 🍡 

Thanks a lot for digging in here @kwonoj - I can't imagine this was a quick one to. understand, thanks so much!

PR #275, opened 2017-06-09
Check windows support on CI

orta:
With #273 we should have working Windows support, this will allow appveyor to run the tests to verify. 👍 

#trivial


orta:
🆗 
💯 
🥇

PR #277, opened 2017-06-09
Adds prettier to the codebase

orta:
Adds prettier to the codebase. This wasn't too tricky, as I've already done this on https://github.com/artsy/emission/pull/567

The settings in prettier didn't require any changes to our tslint setup, which is nice. 

@vjeux - another easy integration WRT the TypeScript support. Thanks! I know it's a long running discussion, but I'd love to be able to put settings in one place ( I love how jest does it in the package.json ;) )


vjeux:
Yeah, I'm going to add support for a centralized settings. I just need to actually sit down and spec it out correctly.

PR #278, opened 2017-06-10
Adds danger-yarn plugin

orta:
This is a bit of a dupe right now, as the contents of that plugin came from this dangerfile - but I'm more interested in overall integration. If this works then I'll modify the Dangerfile inside this PR.


orta:
I feel like there's something uniquely wrong here, maybe it's my plugin? Danger has no problem with other external modules inside the dangerfile.

```ts
import * as fs from "fs"
import * as child_process from "child_process"
import fetch from "node-fetch"
import { distanceInWords } from "date-fns"

// For some reason we're getting type errors on this includes module?
// Wonder if we could move to the includes function in ES2015?
import * as includesOriginal from "lodash.includes"

```

orta:
It is - lols

orta:
Just went to check the files:

screen shot 2017-06-10 at 11 27 25

orta:
OK, this works, I'm re-writing danger-plugin-yarn now - so may have improvements

orta:
oh wrong place, this was meant to go https://github.com/macklinu/generator-danger-plugin/pull/16 - ah well @macklinu will still see it

orta:
Doh - spent all this time thinking things were broken, but really the check only works for `dependencies`, not `devDependencies`.

source/runner/_tests/fixtures/__DangerfilePlugin.js
```diff
@@ -0,0 +1,7 @@
+import { checkForTypesInDeps } from "danger-plugin-yarn"
+const deps = {
+  dependencies: {
+    added: ["@types/danger"],
+  },
+}
+checkForTypesInDeps(deps)
```
orta: this is a weird-ish test, as it's making assertions about how `danger-plugin-yarn` works, but I'd rather have a test for something, and this seems reasonable enough

source/dsl/GitDSL.ts
```diff
@@ -40,9 +40,9 @@ export interface JSONDiffValue {
   before: any
   /** The value after the PR's applied changes */
   after: any
-  /** If both before & after are arrays, then you optionally get what is added. Emprty is no additional objects. */
+  /** If both before & after are arrays, then you optionally get what is added. Empty if no additional objects. */
   added?: any[]
-  /** If both before & after are arrays, then you optionally get what is removed. Emprty is no removed objects. */
+  /** If both before & after are arrays, then you optionally get what is removed. Empty ig no removed objects. */
```
orta: typos

.travis-just-danger.yml
```diff
@@ -0,0 +1,10 @@
+language: node_js
+cache:
+  yarn: true
+  directories:
+  - node_modules
+
+node_js: node
+script:
+- yarn run link
+- danger run --verbose
```
orta: this isn't run by default, but it's nice when you're working on the dangerfile and want to test quickly

PR #280, opened 2017-06-12
Adds support for posting statuses

orta:
Re: #279 

This has no tests, but shouldn't be too hard to add them pretty soon.

### DIFF
--- a/source/platforms/platform.ts
+++ b/source/platforms/platform.ts
@@ -43,19 +43,9 @@ export interface Platform {
   editMainComment: (newComment: string) => Promise<any>
   /** Replace the main Danger comment */
   updateOrCreateComment: (newComment: string) => Promise<any>
+  /** Sets the current PR's status */
+  updateStatus: (passed: boolean, message: string) => Promise<boolean>
 }
-
-//     /** Download all the comments in a PR */
-//     async downloadComments: (id: string) => Promise<Comment[]>;
-
-//     /** Create a comment on a PR */
-//     async  createComment: (body: string) => Promise<?Comment>;
-
-//     /** Delete comments on a PR */
-//     async deleteComment: (env: any) => Promise<booleanean>;
-
-//     /** Edit an existing comment */
-//     async editComment: (comment: Comment, newBody: string) => Promise<booleanean>;
 
 /**
  * Pulls out a platform for Danger to communicate on based on the environment
--- a/source/runner/Executor.ts
+++ b/source/runner/Executor.ts
@@ -145,6 +145,9 @@ export class Executor {
 
     this.d(results)
 
+    const failed = fails.length > 0
+    const successPosting = await this.platform.updateStatus(!failed, messageForResults(results))
+
     if (failureCount + messageCount === 0) {
       console.log("No issues or messages were sent. Removing any existing messages.")
       await this.platform.deleteMainComment()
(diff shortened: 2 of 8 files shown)


source/runner/Executor.ts
```diff
@@ -185,3 +190,28 @@ export class Executor {
...
+const compliment = () => {
+  const compliments = ["Well done.", "Congrats.", "Woo!", "Yay.", "Jolly good show.", "Good on 'ya.", "Nice work."]
+  return compliments[Math.floor(Math.random() * compliments.length)]
+}
+
+const pluralize = (word: string, array: any[]) => `${array.length} ${word}${array.length ? "s" : ""}.`
+
+const messageForResults = (results: DangerResults) => {
+  if (results.fails.length && results.warnings.length) {
+    return `All green. ${compliment()}`
```
macklinu: This is a nice touch. 😄

source/runner/Executor.ts
```diff
@@ -185,3 +190,28 @@ export class Executor {
...
+  }
+
+  let message = "⚠️ "
+  if (results.fails.length) {
+    message += pluralize("Errors", results.fails)
```
macklinu: Possible bug (if I'm reading this function correctly): when there are 0 errors, `pluralize()` will return "0 Errors"; when there is more than 1 error, `pluralize()` will return "n Errorss" with two **s**'s.

Looking at some example success and failure statuses I see on GitHub:

image

image

Thoughts on stating "⚠️  Danger found some issues. Don't worry, everything is fixable." for any warnings or errors? I feel like that wording is generic enough for all cases and simplifies this function.

source/platforms/platform.ts
```diff
@@ -43,20 +43,10 @@ export interface Platform {
   editMainComment: (newComment: string) => Promise<any>
   /** Replace the main Danger comment */
   updateOrCreateComment: (newComment: string) => Promise<any>
+  /** Fails the current PR */
```
macklinu: Thoughts on changing this doc comment to something like "Sets the current PR's status. See the [GitHub Statuses API documentation](https://developer.github.com/v3/repos/statuses/#create-a-status)." to document that this could either set the PR status as success or failure?

source/runner/Executor.ts
orta: Yeah, that seems fine 👍

source/platforms/platform.ts
orta: aye

orta:
So the Danger token on danger doesn't support the status API, but I trust that it's OK

orta:
ill try write some tests tomorrrow

orta:
Tomorrow comes today - @macklinu if you're up to it, you're welcome to merge + ship this as a release, I'll update the Artsy danger instance to verify it all etc in the UK morning

macklinu:
Yeah I can release a minor update this afternoon/evening - will test at my work as well. I'm still not super clear on how you ship updates, like this commit - https://github.com/danger/danger-js/commit/d8d1326807815b032d865a7afdd04b2c9e54d9ea. Tried reading through #117 but I'm still not 100% on it:

* modify the changelog and package.json with the new version
* commit
* run `npm publish`? (I feel like a `git tag` or `npm version` is missing from this list)

Thanks @orta 😄

orta:
No worries, yeah, I make those commits then push a tag - that's it

orta:
The tag triggers a deploy on travis, I'm still new to this idea of not deploying locally - in ruby I make these PRs, https://github.com/danger/danger/pull/836 then do `rake deploy` which is `yarn publish` and then it runs https://github.com/mattbrictson/chandler

orta:
I wanna try this semanticc versioning biz, but not got futher

macklinu:
That's funny; I actually like deploying from CI more than I do locally. 😄 But I'm going to merge this and ship 0.21.0 using the current deployment process. Thanks for clarifying!

PR #282, opened 2017-06-16
fix(github): use HTTP for status check URL

macklinu:
When I click the **Details** link in the GitHub status checks of a pull request:

image

I am taken to https://danger.systems/js - however, I see this **Your connection is not private** message:

image

I've updated the danger.systems URL to use HTTP, which seems consistent with the rest of the codebase. 👍 

image

Also, running `yarn` updated the yarn lockfile.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,8 @@
 
 ### Master
 
+-   Use HTTP for the GitHub status check target URL - macklinu
+
 ### 0.21.0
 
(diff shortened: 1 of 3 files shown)


orta:
Yep 👍

macklinu:
Ooo let me update the changelog (why do I always forget)

orta:
Your commit style allows you to auto-generate them - it makes sense that you'd forget

macklinu:
Good point! 😄

PR #284, opened 2017-06-16
Correct some examples in node-app

clintam:
* need to use github property to access the pr
* the length check does not need a negation


orta:
Awesome - thanks for this @clintam

PR #286, opened 2017-06-19
Add support for buddybuild CI

clintam:


orta:
Yep,  looks good - I want to ship a release today/tomorrow - so this should get in pretty quick 👍

PR #287, opened 2017-06-19
Support running on a Github App

clintam:
In this case the get /user API is not avaliable

See: https://platform.github.community/t/obtaining-the-id-of-the-bot-user/2076


orta:
This is great - in Peril I had this problem and just stubbed it out somewhere - think this is a better call 👍

orta:
Yep 👍

PR #290, opened 2017-06-23
Use absolute urls for the node_modules folder to improve module resolution for Peril

orta:
I'm seeing issues with parsing typescript files in Peril. Think this should get it.

### DIFF
--- a/source/runner/DangerfileRunner.ts
+++ b/source/runner/DangerfileRunner.ts
@@ -4,6 +4,7 @@ import { readConfig } from "jest-config"
 import * as NodeEnvironment from "jest-environment-node"
 import * as os from "os"
 import * as fs from "fs"
+import * as path from "path"
 
 import { DangerResults } from "../dsl/DangerResults"
 import { DangerContext } from "../runner/Dangerfile"
@@ -66,7 +67,7 @@ export async function dangerJestConfig() {
       defaultPlatform: "danger-js",
     },
     moduleNameMapper: [],
-    moduleDirectories: ["node_modules"],
+    moduleDirectories: [path.resolve(process.cwd(), "node_modules")],
     moduleFileExtensions: ["js", ...jestConfig.config.moduleFileExtensions],
     transform: [["js$", "babel-jest"], ...jestConfig.config.transform],
     testPathIgnorePatterns: jestConfig.config.testPathIgnorePatterns,
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,8 @@
 
 ### Master
 
+-   Use an absolute url for the jest-resolver config moduleDirs option so that Peril can run dangerfiles out of process - orta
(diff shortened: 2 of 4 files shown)


orta:
Interesting

source/runner/DangerfileRunner.ts
```diff
@@ -66,7 +67,7 @@ export async function dangerJestConfig() {
       defaultPlatform: "danger-js",
     },
     moduleNameMapper: [],
-    moduleDirectories: ["node_modules"],
+    moduleDirectories: [path.join(process.cwd(), "node_modules")],
```
macklinu: I wonder if [`path.resolve()`](https://nodejs.org/api/path.html#path_path_resolve_paths) is more what we're looking for here. From the docs:

> [it] resolves a sequence of paths or path segments into an absolute path.

orta: Let's give it a shot 👍

orta: Looks like it doesn't - at least the results are actually the problem that I was having on Peril.

orta:
For myself, errors look like this:

```
  ● with fixtures › can handle a plugin (which is already used in Danger)
    Cannot find module 'danger-plugin-yarn' from '__DangerfilePlugin.js'
      
      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:172:17)
      at Object.<anonymous> (source/runner/_tests/fixtures/__DangerfilePlugin.js:1:65)
      at source/runner/file:/C:/projects/danger-js/source/runner/DangerfileRunner.ts:97:13
      at ensureCleanDangerfile (source/runner/file:/C:/projects/danger-js/source/runner/DangerfileRunner.ts:136:3)
      at Object.<anonymous> (source/runner/file:/C:/projects/danger-js/source/runner/DangerfileRunner.ts:96:3)
      at step (source/runner/DangerfileRunner.ts:32:23)
      at Object.next (source/runner/DangerfileRunner.ts:13:53)
      at source/runner/DangerfileRunner.ts:7:71
      at Object.<anonymous>.__awaiter (source/runner/DangerfileRunner.ts:3:12)
      at Object.runDangerfileEnvironment (source/runner/DangerfileRunner.ts:126:12)
      at Object.<anonymous> (source/runner/_tests/file:/C:/projects/danger-js/source/runner/_tests/_danger_runner.test.ts:153:27)
      at step (source/runner/_tests/_danger_runner.test.ts:32:23)
      at Object.next (source/runner/_tests/_danger_runner.test.ts:13:53)
      at fulfilled (source/runner/_tests/_danger_runner.test.ts:4:58)
  √ creates a working jest config (15ms)
```

Effectively the dependency can't be resolved from inside the eval context

orta:
This isn't used in the post-jest, vm2 world

PR #297, opened 2017-06-28
More guides updates

orta:
Removes a guide that I needed to move to slim as it was getting too complex to keep in markdown. Fixes my markdown linter from breaking the layout on one page too.


hansemannn:
Heyho! The [guide](http://danger.systems/js/guides/getting_started.html) has an invalid link you might check out - near to `For a deeper understanding of using a Dangerfile, see the guide [The Dangerfile][dangerfile]`.

dkhamsing:
Looks like danger.systems moved to gitlab https://gitlab.com/danger-systems/danger.systems

PR #299, opened 2017-06-30
Make danger report

orta:
I want a screenshot


tivac:
> New module Danger

Where did this info come from? I don't see anything in https://github.com/danger/danger-js/blob/master/dangerfile.ts to implement that behavior but it's pretty neat!

orta:
It's here: https://github.com/danger/danger-js/blob/master/dangerfile.ts#L42-L43

tivac:
Duh, thanks @orta. I saw those lines and skipped right past 'em 😞

PR #303, opened 2017-07-05
feat(getPullRequestCommits): Support retrive paginated pull request c…

kwonoj:
This PR updates behavior of `getPullRequestCommits` to support retrieve possible commit list instead of single page. Still api itself has hard limits of 250commit in singe PR, but I think this is suffecient for most of cases. 

Updated behavior try to naively retrieve available list, will bail out if one of attempt to get page fails and return known-latest, instead of keep trying to recover full list.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,7 @@
 //  Please add your own contribution below inside the Master section, ideally with a consumer's perspective in mind.
 
 ### Master
+-   Support retrive paginated pull request commit list - kwonoj
 
 ### 1.0.0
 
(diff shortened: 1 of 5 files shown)


orta:
Ah sorry for the late review, I get really itchy when new dependencies are added - as we're now adding two more dependencies to everyone using Danger. Any chance you can add me to your repo and I'll convert it to not rely on `parse-link-header`?

orta:
The idea and implementation are 👍 though

kwonoj:
@orta you can update this PR as I enabled options 
(
image
)

feel freely edit PR as necessary. I picked external module as I found parsing github's link header needes some edge case handling, choose module already does it well.

orta:
Thanks - I totally understand 👍

orta:
OK, looks like I didn't get around to this, so I'm merging 💃 - sorry for the delay @kwonoj

Issue #308, opened 2017-07-21
Running locally without any integration


adam-moss:
Apologies if this is obvious, or has been asked before - I have had an admittedly quick look but can't see anything.

Is there any way to run danger-js locally without trying to "phone home" to github?  I.e. as a simple cli tool that dumps to the console, exit 1 on error type of thing?

I've seen the `DANGER_FAKE_CI` env var which I'd hoped might do this but sadly not 😭 

Longer term Peril is probably the better answer for my particular use case, but again without the github bit.

Thanks

orta:
No afraid not, Danger JS's original vision was to  be entirely API based (so that peril could be built.)  Which has been my main focus when working on it.

I'm definitely open to a new command like `danger local` that uses git commands to generate the DSL. It could pass `null` in for `danger.github` but you can have the entire API for `git` and it's functions working by generating the same diff that github provides. 

I'm assuming the use case would be for something like pre-commit hooks, so I think you'd also need to be able to state the branch to compare against.

adam-moss:
Hmmm... pre-commit wasn't the use case I actually have, but that would be a fair side affect of the same I think.  Downside of that though is you'd probably then want to support multiple dangerfiles depending on how many/complex/timeconsuming rules there were and whether they were relevant at the individual commit stage.

I'll have a gander through the code and see if a direct git api is something I could do 👍

orta:
Aye, there's some discussion on multiple run Danger comments in #196 - some of my apps use a before and after Dangerfile in Ruby due to iOS builds taking forever to compile and run

adam-moss:
If you get a moment could you give me some pointers on what would need to change - can't see the wood for the trees at the moment!  Thanks

orta:
Roughly, you'd need to

* create a [new command](https://github.com/danger/danger-js/blob/master/source/commands/danger-pr.ts) 
* which works with [a new GitCLI Platform](https://github.com/danger/danger-js/blob/master/source/platforms/platform.ts#L31) object to [re-create the git DSL](https://github.com/danger/danger-js/blob/master/source/runner/Executor.ts#L78) using Git
* (you might find something useful in how the [ruby git works](https://github.com/danger/danger/blob/master/lib/danger/scm_source/git_repo.rb))
* You'd need to have a console output version of the commenting part of the Platform interface

orta:
[...]

Issue #261, opened 2017-05-24
danger pr throwing "Cannot find module 'package.json'"


kanaabe:
👋  I'm currently trying to add danger to my [project](https://github.com/artsy/positron). I've set up my dangerfile.js as so:

```
const { danger, fail } = require('danger')

const someoneAssigned = danger.github.pr.assignee
if (someoneAssigned === null) {
  fail("Please assign someone to merge this PR, and optionally include people who should review.")
}

```
and if I run `danger pr` I get back an error:
```
> danger pr https://github.com/artsy/positron/pull/1
(node:97283) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3): Error: Cannot find module 'package.json'
(node:97283) DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

I have danger installed both globally and to my project repo. My local version of Node is v7.10.0. Any thoughts?

frozegnome:
I am also having the same issue during my CI build on my application after greenkeeper updated danger to the latest version (0.19)

+1

kwonoj:
check https://github.com/facebook/jest/pull/3723, this seems `babel-jest` side issue.

orta:
Thanks @kwonoj :D

threeve:
This really bummed me out just now trying to set up danger on our project finally. Until the new version of Jest is released it seems that a simple workaround is to ensure you have a `.babelrc` file in your project containing simply an empty json object `{}`. With that in place I'm able to get danger working both on jenkins and locally with `danger pr` 👍

orta:
We ended up working around this in Artsy by moving to Peril - https://github.com/danger/peril

hramos:
I've been running into this as well. Looks like this is fixed in jest#master, and the fix is available in Jest 20.1.0-alpha.3. I'm guessing we won't want to move Danger to an alpha release just yet, so we'll need to wait for the final Jest 20.1.0 release to become available.

hawkrives:
I'm running in to this error when I'm just running danger in TravisCI – should I just wait for a release that includes Jest 20.1?

kwonoj:
https://github.com/danger/danger-js/issues/261#issuecomment-308561728 is well working workaround. example : https://github.com/ReactiveX/rxjs/blob/d2a32f9a18ebbf65bea798f558364571c91a9d79/.travis.yml#L32

[...]

PR #309, opened 2017-07-22
WIP: support standalone local execution

adam-moss:
Working in progress for #308 ... adding so you can see if this is going along the right track.

I did initially create a `local` command but it seems if no env vars are specified it defaults to it anyway with the `run` method, so I took it back out.

Comments/modifications appreciated.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,8 @@
 
 ### Master
 
+- Add support for running Danger against a local git repo - adam-moss
+
 ### 1.0.0
 
 Hello readers! This represents a general stability for Danger. It is mainly a documentation release, as
--- a/source/ci_source/providers/index.ts
+++ b/source/ci_source/providers/index.ts
@@ -9,6 +9,7 @@ import { Codeship } from "./Codeship"
 import { Drone } from "./Drone"
 import { Buildkite } from "./Buildkite"
 import { BuddyBuild } from "./BuddyBuild"
+import { LocalRepo } from "./local-repo"
 
 const providers = [
   Travis,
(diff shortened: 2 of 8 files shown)


orta:
This definitely is the right way to get started 👍 

I think I'd rather have a specific command for doing this, the version without env vars is really so you can fake your CI and still work against GitHub/GitLab, this aims to work without that completely.

adam-moss:
Ok, I've added the `local` command back in a removed `LocalRepo` as a valid CI source so `run` can't use it.

Now I just need to work out why it is dying with a "can't find package.json" when executing the dangerfile.  I presume this is because I haven't actually populated any of the arrays yet...

orta:
I think the package.json issue is https://github.com/danger/danger-js/issues/261 - which has a quick fix workaround

adam-moss:
Yeah, it is 🤦‍♂️

zdenektopic:
Is there any progress on this? I would really appreciate this as we are using Bitbucket, but is easier for us to use our own method of passing the markdown results to PR.

orta:
@zdenektopic you should build out bitbucket support for DangerJS, we have it in Danger Ruby - this PR doesn't really affect that

zdenektopic:
@orta I am aware. But I don't really need bitbucket support as I don't need any info basically. Just to be able to compare 2 branches and work with that. And then get markdown table with results which I can store to file or export to env temporarily and then read from when composing comment response from CI.

adam-moss:
Been on vacation so haven't done much with this, will look again at some point this week.

zdenektopic:
Cool, because I would really appreciate it! :)

orta:
Thanks @adam-moss - I've taken your work and mostly wrapped it up today, see #473

PR #310, opened 2017-07-24
Update to latest TypeScript

orta:
#202 can't run because it requires `ts-node` to update at the same time.


orta:
Also does Madge
Fixes #296 #306

PR #311, opened 2017-07-26
Add VSTS CI provider

mlabrum:
This adds https://www.visualstudio.com/team-services/ support as a CI.


orta:
Awesome, looks good, let's get this shipped

orta:
Shipped in 1.1.0

lumaxis:
@mlabrum Thanks a bunch for this, I wanted to look into adding VSTS support for quite some time but never got around to it 😊
Had a question: Do you see any reason why this wouldn't work with code hosted on GitHub Enterprise or generic git for that matter? We have a somewhat complicated setup where we trigger CI builds for GitHub Enterprise code manually where the values would be: `BUILD_REPOSITORY_PROVIDER == "Git"`and `BUILD_REASON == "Manual"`. I'm wondering if we could loosen the checks a bit? 🤔

mlabrum:
@lumaxis I set it pretty strict just to make sure that users were using VSTS with github and not with the VSTS code backend. 

I thought github enterprise links with VSTS the same as github hosted? If you're triggering builds manually you would need to make sure that `BUILD_SOURCEBRANCH` contained the github merge branch (with PR number)

I see no reason why you couldn't loosen the restriction....

PR #312, opened 2017-07-26
Fix unable to checkout files with : in the filename

mlabrum:
Hey,

Windows doesn't like `:` in filenames, so I've replaced `static_file:<numbers>.json` to `static_file.<numbers.json`

Cheers,


orta:
Sure thing - thanks!

Issue #316, opened 2017-07-27
Expose a function on the DSL for getting a file via the GitHub API


orta:
This is pretty essential in plugins and for peril support.

orta:
This is in `danger.github.utils.getFile`

PR #318, opened 2017-07-28
Update the DSL to expose a file contents API

orta:
I needed this in [danger-plugin-spellcheck](https://github.com/orta/danger-plugin-spellcheck) as it [had to be re-implmented](https://github.com/orta/danger-plugin-spellcheck/blob/dd1b388621a9dd1f6babe360fd38d6ec71d9a870/src/index.ts#L51-L65) and it is definitely useful for any upcoming plugins I'd like to run.

Also updates the docs with some of the techniques I've used with integration testing plugins.


orta:
Fixes https://github.com/danger/danger-js/issues/316

PR #320, opened 2017-07-31
Move dtslint to devDependencies

### LINKED ISSUE #319: dtslint shouldn’t be in runtime dependencies
Because it depends on tsutils which requires a TypeScript peer dependency:

> npm WARN tsutils@1.9.1 requires a peer of typescript@>=2.0.0 || >=2.0.0-dev || >=2.1.0-dev || >=2.2.0-dev || >=2.3.0-dev || >= 2.4.0-dev but none was installed.

sapegin:
Fixes #319

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "babel-preset-stage-3": "^6.22.0",
     "danger-plugin-yarn": "^0.2.9",
     "date-fns": "^1.28.3",
+    "dtslint": "^0.1.2",
     "husky": "^0.14.0",
     "in-publish": "^2.0.0",
     "jest": "^20.0.0",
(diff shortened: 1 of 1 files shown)


orta:
Hrm interesting, maybe I should add this at runtime on the CI test instead

orta:
interesting, the code already does `- if [[ $TRAVIS_NODE_VERSION == "8" ]]; then yarn add dtslint; yarn dts-lint; fi`

sapegin:
So we can just remove from package.json?

orta:
That's what I'm trying in #323 - but now basically every PR is red, something must have changed on travis side.

orta:
Fixed in #323 - thanks!

PR #322, opened 2017-08-05
Update VS Code recommendations

orta:
Flow isn't used, and the rest are now built into vscode.


orta:
Master is now green, so this is fine.

PR #324, opened 2017-08-07
ci(Travis): deploy on Node 6

macklinu:
#trivial

A [recent deployment from npm](https://travis-ci.org/danger/danger-js/jobs/261614689#L961) failed because it attempted to deploy on Node 8, which has some breaking changes regarding the prepublish npm script.

Based on [these docs](https://docs.travis-ci.com/user/deployment#Conditional-Releases-with-on%3A), I think this will force deployments on Node 6 for the time being.


orta:
It's good to have it deploy only once anyway, it was really confusing which version build would have deploy'd it 👍

PR #328, opened 2017-08-08
Use a vm2 based Dangerfile runner

orta:
Support [a vm2](https://github.com/patriksimek/vm2) based Dangerfile runner as an alternative to the jest infrastructure. There are a few main reasons for this:

* I haven't been able to completely understand how Jest's internals work around all of the code-eval and pre-requisite setup, which has made it hard to work on some more complex Peril features.

* Jest releases are every few months, personally I love this as a user of Jest, as an API consumer it can be difficult to get changes shipped.

* The fact that both Danger/Jest make runtime changes means that you need to update them together

* I have commit access to vm2, so getting changes done is easy

I like to think of it as having gone from Jest's runner which is a massive toolbox, to vm2 which is a tiny  toolbox where I'll have to add a bunch of new tools to get everything working.

The *massive downside* to this is that Danger now has to have support for transpiling via Babel, or 
from TypeScript unlike before, where it was a freebie inside Jest. Jest handled this so well. This means that a Dangerfile which used to "just work" with no config may not. Thus, IMO, this is a breaking major semver.

Is it likely that you need to make any changes? So far, it seems probably not. At least all of the tests with Dangerfiles original from the older Jest runner pass with the new version.

This is an alpha release, because it's knowingly shipped with some breakages around babel support, specifically:

* Babel parsing of relative imports in a Dangerfile aren't working
* Some of the features which require the `regeneratorRuntime` to be set up aren't working yet

Those are blockers on a 2.0.0 release.


orta:
This is close now to feature parity with the main runner:

screen shot 2017-08-09 at 08 13 30

Just this tricky one:

```
    TypeError: 'set' on proxy: trap returned falsish for property '_transform'

      at node_modules/encoding/node_modules/iconv-lite/lib/streams.js:36:45
      at Object..js (node_modules/vm2/lib/sandbox.js:72:5)
      at _require (node_modules/vm2/lib/sandbox.js:253:24)
      at node_modules/encoding/node_modules/iconv-lite/lib/index.js:135:9
      at Object..js (node_modules/vm2/lib/sandbox.js:72:5)
      at _require (node_modules/vm2/lib/sandbox.js:253:24)
      at node_modules/encoding/lib/encoding.js:3:17
      at Object..js (node_modules/vm2/lib/sandbox.js:72:5)
      at _require (node_modules/vm2/lib/sandbox.js:253:24)
      at node_modules/danger-plugin-yarn/node_modules/node-fetch/lib/body.js:8:15
```

macklinu:
How will we expect users to communicate what their dangerfile syntax is? For example, if the dangerfile should be transpiled with babel, would its file name be `dangerfile.babel.js`? Would the user include a [register](https://babeljs.io/docs/usage/babel-register/)?

orta:
If babel is in the project, it will transpile with the default settings for the project for any `*.js` file for Dangerfile + any file the Dangerfile references, and will skip node_modules. Same for TypeScript, but for `*.ts`. 

_So far_ I don't think it will need any config, still not 100% of the way though.

source/runner/DangerfileRunnerTwo.ts
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Updates a Dangerfile to remove the import for Danger
+ * @param {string} contents the file path for the dangerfile
+ * @returns {string} the revised Dangerfile
+ */
+export function cleanDangerfile(contents: string): string {
+  return contents.replace(es6Pattern, "// Removed import").replace(requirePattern, "// Removed require")
+}
+
+const typescriptify = (content: string): string => {
```
macklinu: Ah, I missed these functions - makes sense 👍

source/runner/DangerfileRunnerTwo.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  const sandbox = {
+    ...dangerfileContext,
+  }
+  return {
+    require: {
+      external: true,
+    },
+    sandbox,
+    compiler: (code, filename) => {
```
macklinu: This is cool!

orta:
OK, this is green with a few caveats around babel support which I think is ok for making an alpha release 👍

orta:
This is how ts-jest handles the custom work on a require: https://github.com/TypeStrong/ts-node/blob/master/src/index.ts#L392-L415

orta:
OK, left notes on the blockers: https://github.com/danger/danger-js/issues/330

PR #334, opened 2017-08-13
Support passing in a file's contents to the Danger VM

orta:
This allows Peril to not have to write to disk, but to run a Dangerfile from memory.


source/ambient.d.ts
```diff
@@ -13,132 +13,3 @@ declare module "jsonpointer"
 declare module "parse-link-header"
 
 declare module "*/package.json"
-
```
orta: https://github.com/patriksimek/vm2/pull/91

Issue #340, opened 2017-08-17
Support a multi-process Danger with custom runners


orta:
OK, now bear with me. 

1st, consider this. With Peril I created the ability to see the Danger DSL _for any PR_:

- https://peril-staging.herokuapp.com/api/v1/pr/dsl?owner=artsy&repo=emission&number=681
- https://peril-staging.herokuapp.com/api/v1/pr/dsl?owner=danger&repo=danger-js&number=338
- https://peril-staging.herokuapp.com/api/v1/pr/dsl?owner=artsy&repo=metaphysics&number=715

I did this so I could make better documentation inside danger.systems, but what would happen if this were formalised a bit, and inside _danger_ instead of Peril?

Imagine that we add a new command:`danger process`.

e.g. `danger process` - which would output the above JSON to STDOUT.

Thennnn, we also make a standard in terms of the results. Today, Danger expects results in [this form](https://github.com/danger/danger-js/blob/1d5c2f7de7004039f10528bfe0fafd166e8c0cbc/source/dsl/DangerResults.ts#L4-L30):

```json
{
  "fails": [{ "message": "Failing message Failing message" }],
  "warnings": [{ "message": "Warning message Warning message" }],
  "messages": [{ "message": "message" }],
  "markdowns": ["markdown"],
}
```

So what about if `danger process` could also take the above as STDIN. Now you can make a pipeline.

```sh
danger process > [ x ] > danger process
```

It leaves the really gnarly work of handling different CIs, getting all the useful data from GitHub/GitLab/BitBucket/etc, amending/creating/deleting Dangerfiles to DangerJS. A Danger runner  just need to take in a DSL as JSON, and output a results JSON to STDOUT.

This means, anyone, in any language could make a Danger runner. So, there can be Swift/Rust/Go/Whatever Dangerfiles in the future.

orta:
This is in.

PR #341, opened 2017-08-18
Adds a Danger Process command that outputs the DSL as a JSON object

orta:
Re: #340 

screen shot 2017-08-18 at 07 32 44

This will now take the results of the Danger PR and pass that into the STDIN of a child process. For example this Ruby Script:

```ruby
#!/usr/bin/env ruby

puts 'Hello world'

str = STDIN.tty? ? 'Cannot read from STDIN' : $stdin.read
puts str
```

With this command `danger process ./scripts/danger_runner.rb` would echo first "Hello world" then output the JSON in the above screenshot. You could make a trivial version of Ruby Danger with this:

```ruby
#!/usr/bin/env ruby

require 'json'
dsl_json = STDIN.tty? ? 'Cannot read from STDIN' : $stdin.read
danger = JSON.parse(dsl_json)

if danger.github.pr.body.include? "Hello world"
  # TODO: sending messages back
end
```

Now I need to handle sending messages back to Danger JS.


zdenektopic:
Is it somehow possible to do this for danger Ruby version and then use the output of Ruby version be parsed and worked with in danger-js?

orta:
It's possible, as with all programming anything is possible but it's not something I plan on doing maintaining three processes that communicate with each other for this feature sounds complex. If you're looking to make Bitbucket work with Danger Process, Bitbucket support will need to be added to Danger JS.

changelog.md
```diff
@@ -4,6 +4,42 @@
 
 // TODO
 
+* Adds a `danger process` command, this command takes amn argument of a process to run which expects the Danger DSL as JSON in STDIN,
+  and will post a DangerResults object to it's STDOUT. This frees up another process to do whatever they want. So, others
+  can make their own Danger runner.
+
+  An example of this is [Danger Swift][danger-swift]. It takes a [JSON][swift-json] file via [STDIN][swift-stdin], [compiles
```
bdotdub: Minor clarification: `JSON file` may be more accurate as `JSON document` if I'm understanding correctly.

orta: You're correct, I've just updated the changelog: https://github.com/danger/danger-js/commit/d0fc1eb6037feb68508c18e7bc0da0a603a282fb

PR #343, opened 2017-08-19
Support using Babel 7 for internal code compilation. 

orta:
Well, this was simple. I updated a bunch of our versions, both 6 + 7 exists in Danger's dev dependency stack (as Jest brought 6 in, which is fine.)

This PR means Babel handles both JavaScript, and TypeScript. It's relatively trivial in terms of danger's implementation. Kinda amazing that something that has taken so much time is so trivial to work with as a consumer.

I'm not 100% that I'm getting the default options, don't want to accidentally kick off a project's plugins just to ensure flow/ts is added. However, all tests are passing locally. 👍


dangerfile.ts
```diff
@@ -18,26 +18,26 @@ import { distanceInWords } from "date-fns"
 
 // For some reason we're getting type errors on this includes module?
 // Wonder if we could move to the includes function in ES2015?
-import * as includesOriginal from "lodash.includes"
-const includes = includesOriginal as Function
-
+import includes from "lodash.includes"
 const sentence = danger.utils.sentence
```
orta: The above was a hack because the inferred type of `lodash.includes` was `object: collection, target: any, fromIndex?:any, somethingElse:any` having a extra mandatory 4th param. By adding `@types/lodash.include` it went to ``object: collection, target: any, fromIndex?:any`.

dangerfile.ts
```diff
@@ -18,26 +18,26 @@ import { distanceInWords } from "date-fns"
...
+  if (!hasChangelog && !isTrivial && !isGreenkeeper) {
+    warn("Please add a changelog entry for your changes.")
 
-  // Politely ask for their name on the entry too
-  const changelogDiff = danger.git.diffForFile("changelog.md")
-  const contributorName = danger.github.pr.user.login
-  if (changelogDiff && !includes(changelogDiff, contributorName)) {
-    warn("Please add your GitHub name to the changelog entry, so we can attribute you correctly.")
+    // Politely ask for their name on the entry too
+    const changelogDiff = await danger.git.diffForFile("changelog.md")
```
orta: Fixing the types above from `includes` highlighted that `danger.git.diffForFile("changelog.md")` actually returns a Promise. An older pre-release version of the API was probably synchronous. A good catch 👍

source/runner/DangerfileRunner.ts
```diff
@@ -7,20 +7,28 @@ import { DangerContext } from "../runner/Dangerfile"
...
-  hasTypeScript = true
-} catch (e) {} // tslint:disable-line
+
 try {
   require.resolve("babel-core") // tslint:disable-line
   require("babel-polyfill") // tslint:disable-line
   hasBabel = true
+
+  try {
+    require.resolve("babel-plugin-transform-typescript") // tslint:disable-line
```
macklinu: Will this require `dangerfile.ts` users to include a `.babelrc` with this transform in order for the TypeScript dangerfile to be evaluated?

macklinu: Before this PR, all that was needed was a project to include the typescript dependency, and the dangerfile.ts could be parsed, right? My question is, is this no longer the case?

orta: Yeh, I thought about this for a bit, and came to the same conclusion. So I started work on bringing back the original typescript optional dep also - so you can have TypeScript via `tsc`, or Babel, and then optionally babel for JS. All depending on if the deps exist.

.babelrc
```diff
@@ -1,11 +1,4 @@
...
-        "es2015",
-        "stage-3"
-    ],
-    "plugins": [
-        "transform-flow-strip-types",
-        "syntax-async-functions",
-        "transform-regenerator"
-    ]
+  "presets": ["es2015", "stage-3"],
+  "plugins": ["syntax-async-functions", "transform-regenerator"]
```
hzoo: Aside: you can use `preset-env` and remove es2015, syntax,transform-regenerator

orta:
Thanks @hzoo - going to call this atomic for now, I've made an issue about updating to the env preset ( I gave it a shot and it didn't pass tests, so I'll give myself a bit more time to figure it out later. )

PR #347, opened 2017-08-19
Use the head instead of the base for thisPR

orta:
It's causing this fail in Peril - https://github.com/CocoaPods/CocoaPods/pull/6968


source/platforms/_tests/fixtures/github_pr.json
```diff
@@ -127,21 +127,21 @@
...
+      "hooks_url": "https://api.github.com/repos/orta/emission/hooks",
+      "issue_events_url": "https://api.github.com/repos/orta/emission/issues/events{/number}",
+      "events_url": "https://api.github.com/repos/orta/emission/events",
+      "assignees_url": "https://api.github.com/repos/orta/emission/assignees{/user}",
+      "branches_url": "https://api.github.com/repos/orta/emission/branches{/branch}",
+      "tags_url": "https://api.github.com/repos/orta/emission/tags",
+      "blobs_url": "https://api.github.com/repos/orta/emission/git/blobs{/sha}",
       "git_tags_url": "https://api.github.com/repos/artsy/emission/git/tags{/sha}",
       "git_refs_url": "https://api.github.com/repos/artsy/emission/git/refs{/sha}",
       "trees_url": "https://api.github.com/repos/artsy/emission/git/trees{/sha}",
```
orta: Just arbitrary changed the head branch to be different from the base branch so that tests can provide which one they use, TBH - should have called it 'head' - but I need to get all these balls in the air - so no time this second.

Issue #348, opened 2017-08-22
FetchError: invalid json response body


macklinu:
I updated to 2.0.0-alpha.9 at work, and the build is now failing. 2.0.0-alpha.7 succeeded without issues. Here is the output of Danger on CI:

```
+ yarn run danger
OK, everything looks good: Jenkins on GitHub
Error:  { FetchError: invalid json response body at https://github.com/redacted/redacted/pull/64.diff reason: Unexpected token N in JSON at position 0
    at /var/lib/jenkins/jobs/redacted/workspace/node_modules/node-fetch/lib/body.js:48:31
    at process._tickCallback (internal/process/next_tick.js:109:7)
  name: 'FetchError',
  message: 'invalid json response body at https://github.com/redacted/redacted/pull/64.diff reason: Unexpected token N in JSON at position 0',
  type: 'invalid-json' }
error Command failed with exit code 1.
```

Here are the contents of `https://github.com/redacted/redacted/pull/64.diff`:

```diff
diff --git a/package.json b/package.json
index 49a2222..bd51aa7 100644
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "connect-history-api-fallback": "1.3.0",
     "cross-spawn": "4.0.2",
     "css-loader": "0.26.1",
-    "danger": "^1.2.0",
+    "danger": "^2.0.0-alpha.9",
     "danger-plugin-jest": "^1.0.1",
     "danger-plugin-jira-issue": "^1.1.0",
     "danger-plugin-no-test-shortcuts": "^2.0.0",
diff --git a/yarn.lock b/yarn.lock
index a8b98b6..7d6d698 100644
--- a/yarn.lock
+++ b/yarn.lock
@@ -1053,11 +1053,10 @@ babel-plugin-transform-strict-mode@^6.24.1:
     babel-runtime "^6.22.0"
     babel-types "^6.24.1"
 
-babel-polyfill@^6.20.0:
-  version "6.23.0"
-  resolved "https://registry.yarnpkg.com/babel-polyfill/-/babel-polyfill-6.23.0.tgz#8364ca62df8eafb830499f699177466c3b03499d"
+babel-polyfill@7.0.0-alpha.19:
+  version "7.0.0-alpha.19"
+  resolved "https://registry.yarnpkg.com/babel-polyfill/-/babel-polyfill-7.0.0-alpha.19.tgz#546aabf80c92f4e9314176a64458a6258e5970de"
   dependencies:
[...]

PR #349, opened 2017-08-25
Bugfix - FetchError: invalid json response body

felipesabino:
Issue: #348

A lot of uncertainties on my part here, so please review carefully 😬 

It seems that the main reason the problem was occurring was that the `api` function tried to always parse the response as `JSON`, even when the error response wasn't a valid JSON string.

I did not want to change too much on how dependencies are being handled and as `fetch.ts` file directly imports `node-fetch`, the only way I could think of testing it was to create a mock http server. It is not the most elegant solution and I guess I could use `jest` mock to do so, I tried that route a bit but it got so complex that I just gave up for now. Anyway, having the http server allowed testing without changing all dependency management for now and it should be versatile if future improvement on tests are needed.

Something else I stumbled upon was that calling either `.text()` or `.json()` on `node-fetch` are not idempotent actions, so in order to not mess with future usages of the response instance I cloned the object before messing with the object to retrieve the body for logging 😬 

I wasn't very sure either how to automatically test `getPullRequestDiff`, so for now I just added a test to make sure the correct headers are actually being sent. They were removed recently and I wasn't sure if it was on purpose, let me know if I need to rollback this change 🤔

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,7 @@
 
 ### Master
 
-// TODO
+- Fixed #348 invalid json response body error - felipesabino
 
 ### 2.0.0-alpha.9
 
--- a/source/api/_tests/fetch.test.ts
+++ b/source/api/_tests/fetch.test.ts
@@ -1,0 +1,90 @@
+// import * as jest from 'jest';
+import * as http from "http"
+
+import { api } from "../fetch"
+
+interface ResponseMock {
+  body?: any
+  statusCode?: number
+  contentType?: string
+}
+
+class TestServer {
+  private port = 30001
+  private hostname = "localhost"
+  private response: ResponseMock = null
+  private router = (req, res) => {
+    res.statusCode = this.response && this.response.statusCode ? this.response.statusCode : 200
+    res.setHeader(
+      "Content-Type",
+      this.response && this.response.contentType ? this.response.contentType : "application/json"
+    )
+    res.end(this.response ? this.response.body : null)
+  }
+  private server = http.createServer(this.router)
+
+  start = async (response: ResponseMock): Promise<void> => {
+    this.response = response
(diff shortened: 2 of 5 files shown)


orta:
Hrm, interesting.

So in this case I switched it out because you can't use this API with Peril ( GitHub Apps [have](https://developer.github.com/v3/apps/available-endpoints/#available-endpoints) to set their Accept to `application/vnd.github.machine-man-preview+json`)

screen shot 2017-08-25 at 15 34 24

So - this could mean that Peril will have to handle this specific API route differently?

felipesabino:
Oh snap, so that's the error's real source

We could [list all files for a pull request and get the _"patch"_](https://developer.github.com/v3/pulls/#list-pull-requests-files) but I [tested with this request diff but parse-diff throws an error](https://runkit.com/embed/783pm46paqxf) and just changing that won't be enough...

Don't know exactly how to proceed now, will have some time over the weekend and will get back to it if I end up having some insight on how to do it without having to change too many things

orta:
Can you explain what you're seeing in terms of symptoms? Is it #348 ?

orta:
OK, I've gone and done some digging. This PR is fine to go as-is. It'll break peril, so I'll try get that fixed before a release.

The key here is that GitHub accepts multiple media types [inside the Accept](https://developer.github.com/v3/media/) header

> This is done by adding one or more of the following types to the Accept header when you make a request

My job will be to figure out _what_ that looks like for the diff API.

felipesabino:
The wording from Github API is somewhat misleading, isn't it? Saying `you must provide` and then pasting a random `Accept` Header example, it looked like that was the only option, but now I think I got it.

But cool, that's was actually very proactive of you, thanks for taking a look.

PR #353, opened 2017-08-27
Merge accept requests 

orta:
This expands on #348 and adds support [for merging accept headers](https://developer.github.com/v3/media/). The docs on the github API don't actually say what they'll expect, so I'm kinda hoping this'll just work given the spec and the ordering.

Converts "accept" to "Accept" as that's what's in the docs too.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,10 @@
 
 ### Master
 
-- Fixed #348 invalid json response body error - felipesabino
+### 2.0.0-alpha.12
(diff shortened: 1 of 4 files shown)


orta:
Yep, this looks good ( we'd see missing modified_files here: https://peril-staging.herokuapp.com/api/v1/pr/dsl?owner=artsy&repo=eidolon&number=609 if it was broken)

PR #359, opened 2017-08-28
Bring back relative files

orta:
Again, as with https://github.com/danger/danger-js/pull/355 I'm a bit confused that this seemed to just work...

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,8 @@
 
 ### Master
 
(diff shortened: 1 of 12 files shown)


orta:
Interesting, the tests pass, but the real life example fails - I bet this is due to jest and t's normal typescript support.

orta:
Fixed in #378

PR #362, opened 2017-09-01
Add some docs for the danger process command

orta:
Improves the docs for danger process 👍


orta:
and by improves, I mean actually adds them

Issue #368, opened 2017-09-05
GitDSL: Diffs for added and removed files


bdotdub:
Calling `diffForFile` unexpectedly returned `null` for me for a file that was added. It looks like expected behavior per these lines:

https://github.com/danger/danger-js/blob/72b7c6c7f85672d1982347dd83b8b4a5e48b17d5/source/platforms/github/GitHubGit.ts#L50
https://github.com/danger/danger-js/blob/72b7c6c7f85672d1982347dd83b8b4a5e48b17d5/source/platforms/github/GitHubGit.ts#L150

----------------

It was initially unclear to me why `diffForFile` would strictly return the diffs for modified files only. I can understand why we would want this conceptually (diffs are a byproduct of a _change_ of a file), but there is definitely utility in being able to get a diff for created and removed files.

I'd like to propose that `diffForFile` return a `TextDiff` object for added & removed files. For added files, `added` will be the whole file and removed would be empty, and vice versa for removed files.

It would make getting all changes in the whole PR much easier, rather than calling diff on some, and `fileContents` on others:

```js
danger.git.created_files
      .concat(danger.git.modified_files)
      .concat(danger.git.deleted_files)
      .map(p => danger.git.diffForFile(p))

// vs.

danger.git.modified_files
      .map(p => danger.git.diffForFile(p))

const pr = danger.github.pr;
danger.git.created_files
      .map(p => danger.github.utils.fileContents(p, pr.head.repo.full_name, pr.head.sha));
danger.git.deleted_files
      .map(p => danger.github.utils.fileContents(p, pr.base.repo.full_name, pr.base.sha));
```

(The above code also isn't working for me because of #367)

This change would extend to `JSONDiffForFile` and `JSONPatchForFile`.

----------

Barring that, the documentation should be more explicit in communicating this.

Happy to do the work but wanted to get some thoughts!

orta:
Yeah, I think this behaviour is a great call 👍  - I think your expectations there are what I'd imagine too

orta:
This is in 👍

PR #369, opened 2017-09-07
Include created and removed files for `diffForFile`, `JSONDiffForFile`, `JSONPatchForFile`

bdotdub:
* Implementation for: danger/danger-js#368

**Functions**

* [x] `diffForFile`
* [x] `JSONPatchForFile`
* [x] `JSONDiffForFile`

### DIFF
--- a/source/platforms/github/_tests/_github_git.test.ts
+++ b/source/platforms/github/_tests/_github_git.test.ts
@@ -146,6 +146,62 @@ describe("the dangerfile gitDSL", async () => {
       expect(empty).toEqual(null)
     })
 
+    it("handles showing a patch for a created file", async () => {
+      const before = ""
+
+      const after = {
+        a: "o, world",
+        b: 3,
+        c: ["one", "two", "three", "four"],
+      }
+
+      github.api.fileContents = async (path, repo, ref) => {
+        const obj = ref === masterSHA ? before : after
+        return obj === "" ? "" : JSON.stringify(obj)
+      }
+
+      const gitDSL = await github.getPlatformGitRepresentation()
+      const empty = await gitDSL.JSONPatchForFile("data/schema.json")
+      expect(empty).toEqual({
+        before: null,
+        after: after,
+        diff: [
+          {
+            op: "replace",
+            path: "",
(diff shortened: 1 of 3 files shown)


orta:
This is looking sweeeeeeet 👍

source/platforms/github/_tests/_github_git.test.ts
```diff
@@ -146,6 +146,62 @@ describe("the dangerfile gitDSL", async () => {
...
+        const obj = ref === masterSHA ? before : after
+        return obj === "" ? "" : JSON.stringify(obj)
+      }
+
+      const gitDSL = await github.getPlatformGitRepresentation()
+      const empty = await gitDSL.JSONPatchForFile("data/schema.json")
+      expect(empty).toEqual({
+        before: null,
+        after: after,
+        diff: [
```
bdotdub: @orta Would love to get your thoughts here.

For created files, I think it makes sense to have `before` return `null` since the `before` is the absence of the file. The `diff` is _technically_ correct to, but feel unexpected and/or inconsistent with other behavior.

What do you think about `JSONPatchForFile` returning `null` for `before`, but the `diff` as the result between `{}` and `after`?

So instead of this:

```json
{
  "op": "replace",
  "path": "",
  "value": {
    "a": "o, world",
    "b": 3,
    "c": ["one", "two", "three", "four"],
  }
}
```

it becomes this:

```json
[
  {
    "op": "add",
    "path": "/a",
    "value": "o, world"
  },
  {
    "op": "add",
    "path": "/b",
    "value": 3
  },
  {
    "op": "add",
    "path": "/c",
    "value": [
      "one",
      "two",
      "three",
      "four"
    ]
  }
]
```

While not technically correct, I think it makes more sense semantically.

orta: I agree 👍

bdotdub:
@orta I think this is ready! 👌

orta:
Awesome, amazing and beautifully tested 👍

orta:
Shipped as Danger 2.0a15

PR #378, opened 2017-09-19
Support running via eval/require as well as vm2

orta:
Gets relative paths working correctly via vm2 - deprecates https://github.com/danger/danger-js/pull/359

Updates us to a vm2 which has an index.d.ts shipped https://github.com/patriksimek/vm2/pull/92

Also makes the dangerfile use the custom tsconfig - which means it gets the include function.


orta:
I was testing by using ts-node, which I bet was pre-compiling it in some form for me, will need to test this by using the built versions. Can't write a test with Jest either as it will compile also. 🍡

orta:
OK, I have a clue 

screen shot 2017-10-14 at 12 15 30 pm

When running via the eval, the _file_  for which all relative imports are working from is _of course_ the runner. Not the cwd for the dangerfile.

PR #395, opened 2017-10-16
Separate Danger Run into two processes

orta:
Moves the JavaScript runner to also work in the same method as `danger process`. The idea is that we can use process separation even inside the JavaScript client built into Danger JS. ( see [the docs](http://danger.systems/js/usage/danger-process) if this is new to you ) which means new features to DangerJS will definitely improve anyone using Danger in their own language.  

This process separation will work with https://github.com/danger/peril/issues/159 to solidify the idea of separation for Peril and the runner (on lambda.) 

## TODO:

- [x] Convert `danger` to actually be `danger process "danger runner"` basically
- [x] The DSL needs to have functions added back into it, e.g. `git.utils`, `github.utils`, `github.api`
- [x] Use the runner to run fixtures DSL data agains the JS files inside the runner's fixtures


orta:
OK, this is now green in terms of the biggest refactorings 👍

orta:
CI now runs every example dangerfile through `danger-runner` - see travis' node 8.4

screen shot 2017-10-21 at 5 02 02 pm

orta:
Only the conversion of `danger [args]` to `danger process 'danger runner [args]'` to go

PR #402, opened 2017-10-23
Update JSON RFCwhatever dep

orta:
Closes #390


orta:
The merge should fix this.

PR #429, opened 2017-11-28
Fix gloabally typo

dfalling:
gloabally -> globally


orta:
Thanks for the typo fix! 👍

PR #430, opened 2017-11-30
Danger Init

orta:
A bunch of words need to change (Ruby -> JS, remove anthropomorphizing)  but the majority of the grunt work is done


orta:
These are actually green-ing on windows
screen shot 2017-12-05 at 6 10 25 pm

PR #432, opened 2017-12-04
Fix double negative

dfalling:
Small typo fix for a double negative:

> due to ~~not~~ this run not executing


orta:
Hah- yep. Thanks!

Issue #409, opened 2017-10-24
Support not adding status line


keith:
We run danger in our Jenkins CI setup, from there we determine if it failed based on Danger's output / status code, and fail the jenkins job. It would be nice if Danger didn't also add a custom status line because clicking details there just takes users to Danger's site, where clicking details for our jenkins build would take them to the log instead.

Thoughts?

orta:
I was thinking earlier that it might be better to make it link to the comment that it just posted instead, it will probably just jump up the page but I know I've seen people miss it and say "why did danger failed this?"

orta:
Fixed in 2.1.5

PR #446, opened 2017-12-17
Improve status updates

orta:
* Removes the error log if it fails
* Gives a better URL 

I swear that @keith once mentioned that the status URL could be the CI build URL, but couldn't find the reference. So, I've built that out, as that's useful.

The key is the error messages, that request will fail on most OSS repos, so it shouldn't show a message.


keith:
Nice! For reference: https://github.com/danger/danger-js/issues/409

Issue #177, opened 2017-03-19
Arguments aren't passed to run when using just 'danger'


orta:
https://circleci.com/gh/danger/danger-js/324
https://circleci.com/gh/danger/danger-js/325

After doing 

```js
- yarn
- yarn add danger
```
 I would have expected

`yarn danger -- run --dangerfile dangerfile.circle.js` or `yarn danger -- -d dangerfile.circle.js`  to work. As that's what's in the code.

```js
program
  .option("-v, --verbose", "Verbose output of files")
  .option("-c, --external-ci-provider [modulePath]", "Specify custom CI provider")
  .option("-d, --dangerfile [filePath]", "Specify custom dangerfile other than default dangerfile.js")
  .parse(process.argv)
```

but it didn't work:

```
yarn danger v0.18.1
$ "/home/ubuntu/danger-js/node_modules/.bin/danger" -d dangerfile.circle.js

  error: unknown option `-d'
```

weird

alex3165:
When I install the latest version of danger, the executable JS file look like this:
```
program
    .version(package_json_1.version)
    .command("run", "Runs danger on your local system", { isDefault: true })
    .command("pr", "Runs your changes against an existing PR")
    .parse(process.argv);
```

There is nothing defined for a `-d` option in  `node_modules/danger/distribution/commands/danger.js`

orta:
Yeah, the debugging options have always been an ENV var

macklinu:
It is strange that the TypeScript source shows the `-d` option in `source/commands/danger.ts`, but that line is not present in `distribution/commands/danger.js`, as stated in https://github.com/danger/danger-js/issues/177#issuecomment-287784035. I'm going to take a look at this and see if I can resolve this.

macklinu:
Oops, I was looking at the wrong file. 😛  Still going to take a look.

macklinu:
I wonder if this is related to https://github.com/tj/commander.js/issues/469?

paulmelnikow:
I just ran into this setting up Danger for [Shields](https://github.com/badges/shields).

Would you be up for switching to [argparse](https://github.com/nodeca/argparse)? It's a conscientious port of [Python's absolutely delightful argument parser](https://docs.python.org/3/library/argparse.html). It has a nice API, provides excellent error messages, and most importantly, does not allow errors to pass silently. I've used it in Node projects here and there and had good luck with it.

If not, no worries. I'm just not that invested in fixing commander.

orta:
[...]

PR #447, opened 2017-12-17
Look into handling the the default `danger` command better

orta:
Fixes https://github.com/danger/danger-js/issues/177 without us having to move of commander

Also should fix the commands like `danger-pr`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ matrix:
 script:
 - yarn lint
 - yarn add jest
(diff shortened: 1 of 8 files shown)


orta:
Hrm, somethings up with that Dangerfile but otherwise, seems fine

screen shot 2017-12-17 at 6 46 15 pm

orta:
👍 Makes this easier to test too

PR #450, opened 2017-12-19
Updates typedoc to the latest (to get it to compile with a later TS)

orta:
https://gitlab.com/danger-systems/danger.systems/-/jobs/45249795

> $ /builds/danger-systems/danger.systems/danger-js/node_modules/.bin/typedoc --json docs/js_ref_dsl_docs.json --includeDeclarations
>
> Using TypeScript 2.2.2 from /builds/danger-systems/danger.systems/danger-js/node_modules/typedoc/node_modules/typescript/lib

Basically I can update typedoc and it will move to 2.4 - which is an improvement 👍

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -16,7 +16,7 @@
     "pretty": true,
     "target": "es5",
     "outDir": "distribution",
-    "lib": ["es2017"]
+    "lib": ["dom", "es2017"]
   },
   "formatCodeOptions": {
     "indentSize": 2,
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -16,7 +16,7 @@
     "pretty": true,
     "target": "es5",
     "outDir": "distribution",
-    "lib": ["es2017"]
+    "lib": ["dom", "es2017"]
```
orta: not super pleased on this, but typedoc depends on highlightjs which depends on the dom.

PR #452, opened 2017-12-24
Fixes for `danger pr` vs `danger` WRT async code

orta:
In https://github.com/facebook/react/pull/11865 I spotted that running `danger pr` doesn't actually have the same behavior as `danger` - this is because `danger pr` used the vm2 runner from Peril, but `danger` used the inline process separated runner. 

The current tradeoff is that this means I removed the REPL. TBH, I wanted that feature because of how well it works in Ruby. JS has `--inspect` and I should write up some docs on how to use that instead.


orta:
Looking like the config for danger-run isn't aligned, but this is fine, I can fix that before shipping

PR #455, opened 2017-12-24
Only run the default runner when there are no args (e.g. no calls to the subcommands)

orta:
I think Danger is always being ran, on every run because it's happening for `danger` and for `danger pr` for example.


GantMan:
I believe I saw this issue, too.  I was able to circumvent it by editing node_modules pull of an what's on NPM.

orta:
Bah, I can't figure this one out - gonna give it tomorrow

orta:
the warning above come from the last stable release, so fixing the problem here wouldn't actually fix it there

PR #460, opened 2017-12-29
Add a note about setting up a token in danger pr

orta:
* also fixes release bug where the npmignore would ignore a file called `dangerfile.ts` that was nested deep, I guess at some point the ignore was improved in npm and I updated


CHANGELOG.md
```diff
@@ -1,8 +1,14 @@
+<!--
```
orta: thanks @seanpoulter

PR #461, opened 2017-12-29
Updates some docs for post-2.0 - #trivial

orta:
Adds more more docs


orta:
The fail is fine, sigh

PR #462, opened 2017-12-29
RFC: Refactor CLI into a single program, using argparse

paulmelnikow:
See https://github.com/danger/danger-js/issues/177#issuecomment-349828564.

### DIFF
--- a/source/commands/danger.ts
+++ b/source/commands/danger.ts
@@ -1,7 +1,7 @@
 #! /usr/bin/env node
 
 import { version } from "../../package.json"
-import * as program from "commander"
+import { ArgumentParser } from "argparse"
 import * as debug from "debug"
 import * as chalk from "chalk"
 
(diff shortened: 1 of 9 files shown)


source/commands/danger.ts
```diff
@@ -13,12 +13,47 @@ process.on("unhandledRejection", function(reason: string, _p: any) {
...
+dangerRunner.createParser(subparsers).setDefaults({
+  entryPoint: (args: any) => {
+    dangerRunner.main(args as dangerRunner.App)
+  },
+})
+dangerRun.createParser(subparsers).setDefaults({
+  entryPoint: (args: any) => {
+    dangerRun.main(args as dangerRun.App)
+  },
+})
```
paulmelnikow: There's a lot of repetition here which could be tightened up.

orta:
Want to leave this with me to wrap up?

orta:
I took a look into argparse, it has the exact same flaw as commander: no support for the subparsers being optional ( https://github.com/nodeca/argparse/issues/58 https://github.com/nodeca/argparse/issues/99 ) so I threw a few hours last night and this morning at adding it and couldn't get very far ( current WIP is in https://github.com/danger/danger-js/tree/paulmelnikow-refactor-cli )

paulmelnikow:
Lemme give that a poke. If it comes down to it I'd rather hack on argparse than on commander; the behavior is better documented and the code is in better shape. Worst case, we could wrap a try-catch around the `parseArgs` bit.

paulmelnikow:
There's a `debug` option which does not really debug, but throws an exception when arguments are invalid. Creating a parser to handle the default invocation of `danger run` and trying that first, then falling back to the hybrid command: this seems to work.

orta:
OK, so I gave it a look yesterday also and figured it was probably time to take a clean look at all the CLI and ended up in a different place, I simplified Danger's CLI options and added a lot more documentation in #465 

Docs: https://github.com/danger/danger-js/pull/465/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR11

paulmelnikow:
Alright, seems like I should close this. Maybe I can jump in some other place.

orta:
Peril still gave you invite for helping - that counts to me 👍 

Thanks @paulmelnikow - dunno if this would have got fixed had you not started this

Issue #463, opened 2017-12-31
Passed 2 releases broken (2.1.7 and 2.1.8)


GantMan:
```bash
gantman@Gants-iMac: ~/Documents/Projects/playground/solidarity danger_changer!
 $ npm i danger@2.1.6 -g                                                                                                                       [13:18:05]
/usr/local/bin/danger-process -> /usr/local/lib/node_modules/danger/distribution/commands/danger-process.js
/usr/local/bin/danger-pr -> /usr/local/lib/node_modules/danger/distribution/commands/danger-pr.js
/usr/local/bin/danger-run -> /usr/local/lib/node_modules/danger/distribution/commands/danger-run.js
/usr/local/bin/danger-init -> /usr/local/lib/node_modules/danger/distribution/commands/danger-init.js
/usr/local/bin/danger -> /usr/local/lib/node_modules/danger/distribution/commands/danger.js
+ danger@2.1.6
added 6 packages, removed 8 packages and updated 6 packages in 4.233s

gantman@Gants-iMac: ~/Documents/Projects/playground/solidarity danger_changer!
 $ danger pr https://github.com/infinitered/solidarity/pull/154 -d dangerfile.js                                                               [13:18:14]
{
  fails: [],
  warnings: [],
  messages: [
    {
      message: ":tada:, this worked @GantMan"
    }
  ],
  markdowns: [],
  scheduled: []
}

gantman@Gants-iMac: ~/Documents/Projects/playground/solidarity danger_changer!
 $ npm i danger@2.1.7 -g                                                                                                                       [13:18:19]
/usr/local/bin/danger -> /usr/local/lib/node_modules/danger/distribution/commands/danger.js
/usr/local/bin/danger-pr -> /usr/local/lib/node_modules/danger/distribution/commands/danger-pr.js
/usr/local/bin/danger-process -> /usr/local/lib/node_modules/danger/distribution/commands/danger-process.js
/usr/local/bin/danger-run -> /usr/local/lib/node_modules/danger/distribution/commands/danger-run.js
/usr/local/bin/danger-init -> /usr/local/lib/node_modules/danger/distribution/commands/danger-init.js
+ danger@2.1.7
added 1 package and updated 1 package in 3.443s

gantman@Gants-iMac: ~/Documents/Projects/playground/solidarity danger_changer!
 $ danger pr https://github.com/infinitered/solidarity/pull/154 -d dangerfile.js                                                               [13:18:28]
module.js:529
    throw err;
    ^

[...]

PR #465, opened 2018-01-01
Try fix the CLI

orta:
While work is ongoing in #462 I've done some work to make `danger pr` work, and hopefully undo the damage done in the last two releases #463


orta:
Also run `danger pr` during the tests, so we know for sure if it's broken

orta:
I've shipped this as `3.0.0.beta.1`

paulmelnikow:
Hey, just a heads up, I'm seeing Danger 3.0 run in CI, invoked using plain `danger`. I'm getting a zero exit status though it really should be a failure since we haven't switched to `danger ci`. Also it seems like there's supposed to be some output warning that the invocation needs to change, though it's not being printed:

https://circleci.com/gh/badges/shields/936?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

hawkrives:
I was noticing the "you may not have switched to danger ci" warning even though I have switched… maybe the condition for printing is flipped? (I should really report these in a more timely fashion, sorry.)

Issue #467, opened 2018-01-03
Tests do not work on Windows due to CRLF issues.


jchv:
Hello. Today, I tried to run Danger tests for the first time. My Git installation is very stock and has `core.autocrlf` set to `true` (as is default on Windows.) 

These problems can reliably be resolved by switching off `autocrlf` like this:

```sh
git config core.autocrlf false
git rm --cached -r .
git reset --hard
```

The failures mainly seem to result from code that does not strip `\r`. Therefore errant `\r`s appear at the beginning of lines sometimes, causing diffs to mismatch.

The problem can be solved from two angles:

  1. Handling the carriage return more gracefully where possible. In this case, the bug is specifically from loading a fixture, not something that would ever come up in practice, but the test could be fixed to strip carriage returns.
  2. Forcing `danger-js` to always checkout as UNIX line-endings by specifying it should be so in the `.gitattributes` file:
     ```
     * text eol=lf
     ```

The latter would normalize things better across platforms, but personally seems like a bad idea.

The tests failed with the following results:

```
 FAIL  source\platforms\github\_tests\_github_git.test.ts
  ● the dangerfile gitDSL › sets the modified/created/deleted

    Error

      69 |   it("sets the modified/created/deleted", async () => {
      70 |     const gitJSONDSL = await github.getPlatformGitRepresentation()
    > 71 |     expect(gitJSONDSL.modified_files).toEqual([
      72 |       "CHANGELOG.md",
      73 |       "data/schema.graphql",
      74 |       "data/schema.json",
      Error: expect(received).toEqual(expected)

      Expected value to equal:
        ["CHANGELOG.md", "data/schema.graphql", "data/schema.json", "externals/metaphysics", "lib/__mocks__/react-relay.js", "lib/components/artist/about.js", "lib/components/gene/header.js", "lib/containers/__tests__/__snapshots__/gene-tests.js.snap", "lib/containers/__tests__/gene-tests.js", "lib/containers/gene.js", "tsconfig.json"]
      Received:
[...]

PR #468, opened 2018-01-03
Fix test errors on Windows relating to line-endings.

jchv:
This attempts to resolve an issue with running tests on Windows machines that have `core.autocrlf` set to `true` (default.)

Issue: #467


orta:
Awesome! Thanks!

Issue #464, opened 2017-12-31
Broken link in Docs


GantMan:
Docs refer to example for dangerfile
http://danger.systems/js/guides/the_dangerfile.html#examples

The link to metaphysics is 404 and no dangerfile is found manually searching.

orta:
Man, we move so fast, all of that Dangerfile has moved entirely to Peril

PR #469, opened 2018-01-03
Improve docs #trivial

orta:
Fixes https://github.com/danger/danger-js/issues/464 and adds some more notes


source/runner/Executor.ts
```diff
@@ -111,8 +111,6 @@ export class Executor {
    */
   async handleResultsPostingToSTDOUT(results: DangerResults) {
     const { fails, warnings, messages, markdowns } = results
-    process.exitCode = fails.length > 0 ? 1 : 0
```
orta: This is down in STDOUT already

source/commands/utils/runDangerSubprocess.ts
```diff
@@ -36,32 +37,35 @@ const runDangerSubprocess = (subprocessName: string[], dslJSONString: string, ex
...
       process.exitCode = code
-      const results = resultsWithFailure(`${subprocessName}\` failed.`, "### Log\n\n" + markdownCode(allLogs))
-      await exec.handleResults(results)
+
+      const failResults = resultsWithFailure(`${processDisplayName}\` failed.`, "### Log\n\n" + markdownCode(allLogs))
+      if (results) {
+        results = mergeResults(results, failResults)
+      } else {
+        results = failResults
+      }
```
orta: Avoid sending multiple copies of the results back to the host process

source/commands/danger.ts
```diff
@@ -32,9 +32,13 @@ program
     console.log("       http://danger.systems/js/reference.html")
   })
 
+// Commander mutates process.argv
+const originalProcessArgV = process.argv
```
hawkrives: Don't we need to actually copy the array here, not just grab a reference to it?

Like `[...process.argv]` or something. Otherwise we've just got a second reference to the mutated array, I'd think.

Issue #470, opened 2018-01-04
Danger 3 Warning Message (why?)


GantMan:
run `yarn danger ci`
warning comes out telling me I didn't run it right.

![image](https://user-images.githubusercontent.com/997157/34568097-552caa2e-f12a-11e7-89b0-818f937287a7.png)

Travis report:  https://travis-ci.org/infinitered/solidarity/jobs/324899875

Not sure why Travis got this message, but it's off.

hawkrives:
See also https://github.com/danger/danger-js/pull/465#issuecomment-355191964; I've got a hunch that the condition for printing that message is flipped?

hawkrives:
https://github.com/danger/danger-js/blob/aabb0069e5b3cf91d11fdc7b9dd38225ffb95c39/source/commands/danger.ts#L37

So… the condition's not flipped, but for some reason it's still being printed even when invoked as `danger ci`?

orta:
It likely needs a check if `process.args` does not include the strings: `init, ci, process, pr or --help`

orta:
I've not had much time for Danger during the work week, so will take a look over the weekend

fbartho:
I'm getting this error, even though I'm running danger 3.0.3?

```shell
$ /Users/greenhouse/tmpXwQA6s/node_modules/.bin/danger ci
You may have updated from Danger 2.x -> 3.x without updating from `danger` to `danger ci`.
Skipping Danger due to this run not executing on a PR.
Done in 1.08s.
```

Invoked via a shell script that calls `yarn danger ci`

Note: I'm using the new Nevercode.io CISource, so I don't know if that's related to the issue.

orta:
this is fixed in https://github.com/danger/danger-js/pull/473

but it isn't shipped yet

fbartho:
Ahhhh @orta, thanks, I missed that that PR existed and hadn't shipped yet.

fbartho:
@orta --
1. Is there any way that we can help with your work in #473? (The TODO list looks slightly less clear than I need to jump right in?)
2. Alternately, can we back-port the fix for this issue back into mainline?

This issue is blocking me from running danger right now as my old CI is gone & I'm leaning towards using Nevercode, but the lack of danger is one of my final concerns on selecting that CI provider.

orta:
I'm not sure how this is blocking you, this is just a console.error:

[...]

PR #472, opened 2018-01-05
Check that the danger invocation isn't straight `danger`

hawkrives:
Addresses https://github.com/danger/danger-js/issues/470

Checks that the `danger` command was invoked by way of some subcommand.

### DIFF
--- a/source/commands/danger.ts
+++ b/source/commands/danger.ts
@@ -34,6 +34,7 @@ program
 
 program.parse(process.argv)
 
-if (process.env.CI) {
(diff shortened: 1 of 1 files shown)


orta:
I did not know about the sum function, damn, I could have wrote a lot of code better with that 👍

orta:
Thanks @hawkrives

orta:
Tested this locally, because commander mutates `process.env` this didn't work - fixed it in #469

hawkrives:
Oh right, you mentioned that. Thanks.

Issue #477, opened 2018-01-14
Danger should be able to perform actions on Github


deecewan:
Danger's purpose is 'stop saying you forgot to...'.
In general, this is by automating the 'you forgot to...' message.

But if the item in question is resolvable programmatically, I think Danger should have the capability to fix this.

For instance, if under certain conditions in a pull request, a certain label needs to apply, then Danger should be able to add the label. If the body of a pull request is missing something and it can be determined programatically, then Danger should be able to add it.

tl;dr Danger should be able to 'stop saying you forgot to...' by just fixing the issue, instead of making a comment saying 'you forgot to...'

I propose making the danger Github API have some write functionality (adding labels, modifying the PR body, assigning people, etc).

Edit: this is all already possible using the `danger.github.api` item. My bad 👎

orta:
All of that is available via `danger.github.api` today

orta:
Perhaps the reference docs for it could be improved with a link to the node docs? http://danger.systems/js/reference.html#GitHub

deecewan:
Yeah, I realised this as I was writing my Dangerfile after having just added the `github` module to my project for this purpose.

Yeah, the GitHub reference doesn't actually point anywhere for me. I know now that it's an instance of the `github` npm package.

PR #478, opened 2018-01-14
Improves API docs - #trivial

orta:
Just updates the reference for the GH node API


orta:
Fixes https://github.com/danger/danger-js/issues/477

Issue #493, opened 2018-01-23
`danger local` unknown option `--base`


adam-moss:
Hi,

For `danger local --base develop --dangerfile dangerfile.js` I am getting:

```
error: unknown option `--base'

Error:  TypeError: Cannot read property 'concat' of undefined
    at Object.exports.mergeResults (/Users/adam.moss/projects/project_name/node_modules/danger/distribution/commands/utils/reporting.js:15:33)
    at Object.<anonymous> (/Users/adam.moss/projects/project_name/node_modules/danger/distribution/commands/utils/runDangerSubprocess.js:97:51)
    at step (/Users/adam.moss/projects/project_name/node_modules/danger/distribution/commands/utils/runDangerSubprocess.js:32:23)
    at Object.next (/Users/adam.moss/projects/project_name/node_modules/danger/distribution/commands/utils/runDangerSubprocess.js:13:53)
    at /Users/adam.moss/projects/project_name/node_modules/danger/distribution/commands/utils/runDangerSubprocess.js:7:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/adam.moss/projects/project_name/node_modules/danger/distribution/commands/utils/runDangerSubprocess.js:3:12)
    at ChildProcess.<anonymous> (/Users/adam.moss/projects/project_name/node_modules/danger/distribution/commands/utils/runDangerSubprocess.js:85:48)
    at ChildProcess.emit (events.js:160:13)
    at maybeClose (internal/child_process.js:943:16)
```

For branch structure:

master -> develop -> build/danger

adam-moss:
Ok, so I think this is being caused by `--base` not being a valid option in [sharedDangerfileArgs.ts](source/commands/utils/sharedDangerfileArgs.ts)

Happy to submit a PR to include it if you want?

orta:
Hrm, that's a not ideal - yeah, I'm surprised it should be available because I added it before the shared args:

https://github.com/danger/danger-js/blob/d89be2ad968a0098f5c855931858771f71125369/source/commands/danger-local.ts#L19

adam-moss:
I think this is what is causing it: 
https://github.com/danger/danger-js/blob/d89be2ad968a0098f5c855931858771f71125369/source/commands/danger-runner.ts#L33

orta:
Yes! 

The sub-process part, I see where you're coming from. Great thinking, I think there's a commander option to allow a command to ignore unknown input and this is a perfect case

orta:
[...]

PR #494, opened 2018-01-23
Allows any args on the runner command

orta:
Makes it possible to use the `base` argument in `danger local`. Will ship a release right after this.

### DIFF
--- a/source/commands/danger-runner.ts
+++ b/source/commands/danger-runner.ts
@@ -19,6 +19,7 @@ const d = debug("danger:runner")
 // Given the nature of this command, it can be tricky to test, so I use a command like this:
 //
 // tslint:disable-next-line:max-line-length
+//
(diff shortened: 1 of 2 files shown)


source/commands/danger-runner.ts
```diff
@@ -28,6 +29,9 @@ program
   .description(
     "Handles running the Dangerfile, expects a DSL from STDIN, which should be passed from `danger` or danger run`. You probably don't need to use this command."
   )
+  // Because other calls will trigger this one,
+  // and we don't want to keep a white/blacklist
+  .allowUnknownOption(true)
```
orta: can't really write tests for the commands themselves, so I'm letting myself off here

orta:
Fixes #493

adam-moss:
You ok with that build failure?

orta:
Let's go with yes for now, it likely is the same in the latest release.

PR #499, opened 2018-01-30
Correct section about showing plugins on homepage

hanneskaeufler:
We already display plugins on the website ("AWESOME PLUGINS") so reword the section to make that clear.


orta:
Yeah, great point 👍

PR #525, opened 2018-03-01
Don't check for same user ID on comment when running as a GitHub App

tibdex:
I think the line was wrongly changed [here](https://github.com/danger/danger-js/commit/a3befe7cef72ce6e3d2d3907f126b47c0a413d3c#diff-b77e13a15617a77791367e0df9de18c7R82).
The consequence is that when running as a GitHub App, Danger will keep creating new comments instead of updating the existing one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 ## Master
 
 * Update the PR DSL to include bots. [@orta][]
+* Don't check for same user ID on comment when running as a GitHub App. [@tibdex][]
 
 ## 3.1.7
 
--- a/source/platforms/github/GitHubAPI.ts
+++ b/source/platforms/github/GitHubAPI.ts
@@ -83,7 +83,11 @@ export class GitHubAPI {
 
     return allComments
       .filter(comment => v.includes(comment.body, dangerIDMessage))
-      .filter(comment => userID || comment.user.id === userID)
+      .filter(
+        comment =>
+          // When running as a GitHub App, the user ID is not accessible so we skip the check.
+          userID === undefined || comment.user.id === userID
+      )
       .filter(comment => v.includes(comment.body, dangerSignaturePostfix))
       .map(comment => comment.id)
   }


orta:
For peril? You should be setting [the user ID](https://github.com/danger/peril/blob/master/source/globals.ts#L77) otherwise it'll delete Danger messages that came from the CI as well

This should be pretty easy to fix in Peril - just need to spend some time on https://github.com/danger/peril/issues/179

tibdex:
I'm not using Peril.

I use Danger as part as our CircleCI build.
But I'm working in an organization and we prefer using GitHub Apps instead of personal access tokens or bot accounts. We find it cleaner as it's easy to install the GitHub App on a set of internal repositories and it doesn't take an additional seat on our organization like a bot account would.

tibdex:
> I think the line was wrongly changed [here](https://github.com/danger/danger-js/commit/a3befe7cef72ce6e3d2d3907f126b47c0a413d3c#diff-b77e13a15617a77791367e0df9de18c7R82).

Hi @orta, as I said in my first message, I think this PR is valuable even for non-Peril users.

orta:
Oh yeah, sorry, yeah - I think this makes sense 👍

orta:
OK, that green'd

ggajews:
@tibdex can you write how did you guys manage to run danger with the Github App? I'm struggling with that for a while. Do you get your app token before initializing danger?
Sorry for commenting on this old issue.

tibdex:
We do it like this:

```js
// Danger requires a DANGER_GITHUB_API_TOKEN environment property to work.
// Since we use a GitHub App to interact with the repository, there is an intermediate
// step which is to convert the GitHub App private key to an access token.
// We use fetch-github-app for that.

'use strict';

require('loud-rejection/register');

const envalid = require('envalid');
const execa = require('execa');
const fetchGithubApp = require('fetch-github-app');
const findYarnWorkspaceRoot = require('find-yarn-workspace-root');

const env = envalid.cleanEnv(
  process.env, // eslint-disable-line  no-process-env
  {
    DANGER_GITHUB_APP_ID: envalid.num({
      desc:
        'Can be found in the "GitHub Apps" section of the organization settings.',
    }),
    // Using base64 encoding in order not to deal with cumbersome EOL escaping.
    // eslint-disable-next-line id-match
    DANGER_GITHUB_APP_PEM_BASE64_ENCODED: envalid.makeValidator(str => {
      const privateKey = Buffer.from(str, 'base64').toString('utf8');
      if (
        /-----BEGIN RSA PRIVATE KEY-----[\s\S]+-----END RSA PRIVATE KEY-----/m.test(
          privateKey,
        )
      ) {
        return privateKey;
      }
      throw new Error('invalid GitHub App RSA private key');
    })({
      docs:
        'https://developer.github.com/apps/building-integrations/setting-up-and-registering-github-apps/registering-github-apps/#generating-a-private-key',
    }),
    DANGER_GITHUB_INSTALLATION_ID: envalid.num({
      desc:
        'Can be found in the "Installed GitHub Apps" section of the organization settings',
    }),
  },
  {strict: true},
);

fetchGithubApp({
  appId: env.DANGER_GITHUB_APP_ID,
  // Using base64 encoding in order not to deal with cumbersome EOL escaping.
  appPrivateKey: env.DANGER_GITHUB_APP_PEM_BASE64_ENCODED,
  installationId: env.DANGER_GITHUB_INSTALLATION_ID,
  userAgent: 'activeviam',
}).then(({installationAccessToken}) =>
  execa(
    'danger',
    ['--dangerfile', 'danger/dangerfile.js', ...process.argv.slice(2)],
    {
      // Extend the environment of the child process with the access token.
      env: {DANGER_GITHUB_API_TOKEN: installationAccessToken},
      // To display Danger output in the console.
      stdio: 'inherit',
      cwd: findYarnWorkspaceRoot(),
    },
  ),
);
```

Instead of `fetch-github-app` you should be able to use https://www.npmjs.com/package/@octokit/app

ggajews:
Sweet, thanks a lot!

orta:
I wonder if that could be moved into a custom command in danger-js?

hemp:
Thanks @tibdex , that code above was useful.

Currently using `@octokit/auth-app` <https://www.npmjs.com/package/@octokit/auth-app> in place of `fetch-github-app` since it has been since archived.

PR #538, opened 2018-03-19
Add `reset-status` command

mxstbr:
Adds a new command to set the PR status to "pending" at the beginning of a CI run.

Note: This is completely untested as I have no clue how I'd even start testing this, any hints appreciated. At least the code compiles I guess...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 
 ## Master
 
-* []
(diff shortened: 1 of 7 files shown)


orta:
Yeah, TBH, I try to write tests for these files by moving everything out of a command as quickly as possible. This feels pretty much like I'd expect, and given it's simplicity and the fact that everything else is covered, I'm fine to take it like this 👍

PR #555, opened 2018-04-03
Don't print error for commit status when there was no error.

### LINKED ISSUE #554: Danger printed mistaken error message
In [danger-js/source/runner/Executor.ts, line 200](https://github.com/danger/danger-js/blob/2418999ccc596460ccf64321cf5adce87f1b5bd9/source/runner/Executor.ts#L200), it checks for verbose in the options but logs an error whether or not the commit status was actually set. 

This was confusing during our set up, since the PR passed the dangerfile, but we saw this message in the logs so assumed it had failed.

```
    if (this.options.verbose) {
      console.log("Could not add a commit status, the GitHub token for Danger does not have access rights.")
      console.log("If the build fails, then danger will use a failing exit code.")
    }
```

sunshinejr:
Fixes #554.

### DIFF
--- a/source/runner/Executor.ts
+++ b/source/runner/Executor.ts
@@ -197,7 +197,7 @@ export class Executor {
     const dangerID = this.options.dangerID
     const failed = fails.length > 0
     const successPosting = await this.platform.updateStatus(!failed, messageForResults(results), this.ciSource.ciRunURL)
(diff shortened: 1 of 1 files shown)


orta:
Hah!

orta:
Yeah, that makes more sense

PR #560, opened 2018-04-13
Bitbucket Server Inline mode

codestergit:
### Todo:
- [x]  Proof of Concept
- [x]   Bitbucket Inline API implementation
- [x]  Convert BitBucketServerDiff to  GitStructuredDiff
- [x]  Platform changes
- [x]  Inline Template changes
- [x]  fix test cases
- [x]   Add test cases
- [x]   Add Changelog entry

PR #564, opened 2018-04-17
[inline mode] fix empty comment

### LINKED ISSUE #563: Danger writes empty comment if it have inline comment but no main commet.
Danger writes an empty comment when it has one or more inline comments but no Main comment.

To reproduce you can use below danger file.
```
declare function warn(message: string, file?: string, line?: number): void

warn("Testing", "source/danger.d.ts", 174)
```

codestergit:
Fixes #563

### DIFF
--- a/source/dsl/DangerResults.ts
+++ b/source/dsl/DangerResults.ts
@@ -208,7 +208,7 @@ export function sortResults(results: DangerResults): DangerResults {
   }
 }
 
-export function isEmpty(results: DangerResults): boolean {
+export function isEmptyResults(results: DangerResults): boolean {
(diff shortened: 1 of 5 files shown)


source/dsl/DangerResults.ts
```diff
@@ -208,6 +208,10 @@ export function sortResults(results: DangerResults): DangerResults {
   }
 }
 
+export function isEmpty(results: DangerResults): boolean {
```
orta: Any chance this could mention results? `isEmptyResults`? Then this is good to go!

orta:
Nice one!

orta:
Congrats

codestergit:
Adding these changes to #560.

PR #566, opened 2018-04-17
fix bug when commiting .png files on BitBucket

ionutmiftode:


orta:
Thanks, will probably merge #560 today

PR #569, opened 2018-04-20
Makes the Danger title come first, not last, in the std out

orta:
screen shot 2018-04-20 at 9 33 34 am

felt backwards


sunshinejr:
Hmm, theoretically yes, but right now when you have a long build log it is a lot easier (at least for me) to find the summary - it's near end so you can quickly see the results.

orta:
I agree, was thinking about this afterward,  probably worth doing a check how many lines of text are in the violations, and if it's over ~10 have the message at the bottom

sunshinejr:
Yeah, I like it more 👍

source/runner/Executor.ts
```diff
@@ -148,11 +148,40 @@ export class Executor {
...
+        output = `Danger: ${tick} found only warnings, ${message}`
+      } else if (messages.length > 0) {
+        output = `Danger: ${tick} passed, found only messages.`
+      } else if (!messages.length && !fails.length && !messages.length && !warnings.length) {
+        output = `Danger: ${tick} passed review, received no feedback.`
+      }
+
+      const allMessages = [...fails, ...warnings, ...messages, ...markdowns].map(m => m.message)
+      const oneMessage = allMessages.join("\n")
+      const longMessage = oneMessage.split("\n").length > 30
```
orta: My laptop full screen is 40 lines, so this is 3/4 of that

sunshinejr: cool 👍

PR #573, opened 2018-04-24
Posts the comment URL to the CLI

### LINKED ISSUE #572: When danger posts a message to a service include the link in the CLI
So that it’s easy to cross reference

orta:
Fixes #572  - also sets the clicky link in the status to be that comment

### DIFF
--- a/source/platforms/BitBucketServer.ts
+++ b/source/platforms/BitBucketServer.ts
@@ -206,15 +206,17 @@ export class BitBucketServer implements Platform {
   /**
    * Either updates an existing comment, or makes a new one
    *
+   * @param {string} dangerID the UUID for the run
    * @param {string} newComment string value of comment
-   * @returns {Promise<boolean>} success of posting comment
+   * @returns {Promise<string>} the URL for the comment
    */
-  async updateOrCreateComment(dangerID: string, newComment: string): Promise<boolean> {
+  async updateOrCreateComment(dangerID: string, newComment: string): Promise<string | undefined> {
     const comments = await this.api.getDangerComments(dangerID)
+    let issue = null
 
     if (comments.length) {
       // Edit the first comment
-      await this.api.updateComment(comments[0], newComment)
+      issue = await this.api.updateComment(comments[0], newComment)
 
       // Delete any dupes
       for (let comment of comments) {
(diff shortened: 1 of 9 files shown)


orta:
Works correct: 
screen shot 2018-04-24 at 10 02 39 am

source/platforms/BitBucketServer.ts
```diff
@@ -223,10 +225,15 @@ export class BitBucketServer implements Platform {
...
       }
     } else {
-      await this.createComment(newComment)
+      issue = await this.createComment(newComment)
     }
 
-    return true
+    // We want to make a URL like:
+    // https://bitbucket.org/atlassian/jiraconnect-ios/pull-requests/4/crash-groups/diff#comment-426
+    // But I don't see
```
sunshinejr: Did you forget to add the rest of the comment? 😄

source/platforms/GitHub.ts
```diff
@@ -244,10 +245,10 @@ export class GitHub implements Platform {
       }
     } else {
       this.d(`Creating new comment`)
-      await this.createComment(newComment)
+      issue = await this.createComment(newComment)
     }
 
-    return true
+    return issue && issue.html_url
```
sunshinejr: this is a really smart way of avoiding accessing a property on `null`, although it wasn't very clear to me at first so I had to play with it to confirm that it does what I think it does 😄

source/platforms/GitHub.ts
```diff
@@ -227,13 +227,14 @@ export class GitHub implements Platform {
    * @param {string} newComment string value of comment
    * @returns {Promise<boolean>} success of posting comment
    */
-  async updateOrCreateComment(dangerID: string, newComment: string): Promise<boolean> {
+  async updateOrCreateComment(dangerID: string, newComment: string): Promise<string | undefined> {
```
sunshinejr: Sorry if obvious, but why is there an `undefined` type here instead of `null`?

source/platforms/GitHub.ts
orta: Yeah, this is definitely a JS-ism

source/platforms/GitHub.ts
orta: There's very little difference between undefined and null, but they are semantically different:`x = {}` and `x = {a:undefined}` respond the same to `x.a` but `{a: null}` is different. So I use null to define a conscious missing value - vs undefined which is kinda "not there", if you're doing an `if(x.a)` they're all the same, but some libs will treat null different from undefind

source/platforms/BitBucketServer.ts
orta: lol yep, took a while to figure out how to make the link

PR #574, opened 2018-04-25
Allow returning a value and having Danger execute the code

orta:
Re: https://github.com/danger/peril/issues/240

I don't think this is really something people will need at the danger-js level of abstraction, but I think it makes a lot of sense for Peril when working with tests, events or PRs in general


orta:
wonder if this is a node 10 issue

orta:
Yep, https://github.com/evanw/node-source-map-support/pull/212

PR #579, opened 2018-04-29
Add support for Screwdriver CI

dbgrandi:
Hi! 👋 

This adds support for [Screwwdriver CI](http://screwdriver.cd). It's an open source, self hosted CI that we use at Yahoo. It's based on the Environment Variables described in https://docs.screwdriver.cd/user-guide/environment-variables


orta:
Yep 👍

Issue #592, opened 2018-05-08
Potential use of GitHub checks and statuses.


sunshinejr:
Both https://developer.github.com/v3/checks/ and https://developer.github.com/v3/repos/statuses/ are exciting, might be worth looking into, this screenshot below looks promising:<br />
![image](https://user-images.githubusercontent.com/5232779/39742913-8525dca6-529f-11e8-992b-2afaefd0216b.png)

orta:
This may end up a Peril only thing due to it requiring a GitHub app, but support for the API calls just got merged in the GH api client, https://github.com/octokit/rest.js/issues/852

PR #594, opened 2018-05-11
Support for the new GitHub Checks API

orta:
WIP on support for using GH checks - #592  https://github.com/danger/danger/issues/984

For a normal Danger (only) run, then Danger will need to know the following:

* GitHub App ID
* GitHub Installation ID for the Org
* GitHub App Private Key

Definitely not a blocker, but certainly a bit non-trivial. This is intended to be opt-in for Danger, and probably the default for Peril.

I intend do to it by refactoring the commenting aspect of the GitHub platform to either be the issue comment we have now, or via the checks API exclusively.


orta:
Sidenote:  it's *possible* that this may break using the API with enterprise apps

orta:
Somewhat blocked by https://github.com/octokit/rest.js/pull/866 and https://github.com/octokit/rest.js/pull/864 for the mo'

sunshinejr:
This is great @orta, thanks for working on this so quickly! Really excited about that one, the integration looks awesome

orta:
Got my first check working 👍 

Was trivial work with octokit @gr2m

.travis.yml
```diff
@@ -34,8 +34,20 @@ matrix:
     script:
     - yarn jest --outputFile test-results.json --json --runInBand
     - yarn run link
-    - echo "This is the real `danger ci` run on this repo"
-    - DEBUG="*" danger ci --verbose
+
+    # If the PR is odd then run using issue commenting, else use checks
```
natecook1000: ✨✨ a/b testing ✨✨

fbartho:
This is really exciting. Too bad it's going to require such a tricky integration process :(

orta:
Given the security model of GitHub Apps I can just give out both an app ID, and a private key with access to checks, people will only need to install it on their org and give danger the install ID. Could be less work in general.

.travis.yml
orta: Thanks man, yeah, I was thinking for a while on how I can ensure this is regularly integrated properly - love this hack, I've not yet got it working but I will get it eventually

orta:
I went one better, it's way simpler to set up Danger now:

* Go to https://github.com/apps/danger-js
* Add it to your org
* Get the install ID
* Set that in your rnv
* Add danger to CI

No faffing with making a new user, or new app.

gr2m:
> Somewhat blocked by octokit/rest.js#866 and octokit/rest.js#864 for the mo'

No mo': https://github.com/octokit/rest.js/releases/tag/v15.6.0

orta:
Alright, I'm shipping this so I can try it out in Peril.

PR #596, opened 2018-05-13
Adds support for mutliple danger file runs in a single execution

orta:
Allows Peril to only make one message instead of many.

### DIFF
--- a/source/runner/Executor.ts
+++ b/source/runner/Executor.ts
@@ -93,7 +93,7 @@ export class Executor {
     // If an eval of the Dangerfile fails, we should generate a
     // message that can go back to the CI
     try {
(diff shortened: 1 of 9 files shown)


source/runner/Executor.ts
```diff
@@ -352,12 +352,12 @@ export class Executor {
   }
 
   /**
-   * Takes an error (maybe a bad eval) and provides a DangerResults compatible object3ehguh.l;/////////////
```
sunshinejr: Oh my, that was probably my cat, Ada! 🤣
img_2134

orta: Yep, those are the right keys 👍

PR #599, opened 2018-05-19
Improve peril checks

orta:
Let's `process.env.LOG_FETCH_REQUESTS` include octokit info, so  I can verify https://github.com/octokit/rest.js/pull/878


orta:
Merge on green

PR #601, opened 2018-05-21
Add update issue

orta:
There's a whole lot of extra code from by #600 (which 500s, which I reported to GH) but we'll see how it turns out


orta:
Creating and updating an issue works: https://github.com/PerilTest/sandbox/issues/24

Just figuring out some checks issues now 👍

PR #602, opened 2018-05-21
Use top-level exports in '.d.ts' instead of a module augmentation.

DanielRosenwasser:
Addresses issues raised in https://github.com/Microsoft/TypeScript/issues/24267. You probably want to view with whitespace diffing disabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 
 ## Master
 
+* Updates type declarations to use top-level exports instead of a module augmentation - [@DanielRosenwasser][]
+
 # 3.7.2-11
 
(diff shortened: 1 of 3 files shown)


orta:
Thanks! This is a generated interface, so I'll replicate your changes and fix it in this PR

DanielRosenwasser:
Duh, missed that. :sad:

orta:
Thanks!

Merge on green

orta:
Interesting, looks like the checks API means I can't do merge on green because I had a warning give a neutral result - will see what that looks like when CI is done

orta:
screen shot 2018-05-21 at 1 33 43 pm

Followed the instructions and it looks 👍

PR #605, opened 2018-05-27
Fix big JSON passed from subprocess being cut short

### LINKED ISSUE #603: runDangerSubprocess implementation is incorrect (?)
I encountered an issue, when danger fails on the `validateResults` function (`DangerResults.ts`) although the result is actually correct. It throws `Results passed to Danger JS did not include fails`. The result argument passed here is indeed an empty object, however the result has a correct shape at an earlier point, as it's printed to the console.

I noticed this only happens if I include markdown to the result, and only if the markdown is bigger than a certain size. This makes me think the subprocess stream is not correctly read here: https://github.com/danger/danger-js/blob/master/source/commands/utils/runDangerSubprocess.ts#L38

Because this code assumes the data will arrive in a single chunk (L43), and if it's not a json string (L41), it won't process the data ever.

Am I thinking right?

LarsJK:
Fixes #603

### DIFF
--- a/source/commands/danger-runner.ts
+++ b/source/commands/danger-runner.ts
@@ -48,6 +48,8 @@ const run = async (jsonString: string) => {
   runtimeEnv = await inline.createDangerfileRuntimeEnvironment(context)
   d(`Evaluating ${dangerFile}`)
   await inline.runDangerfileEnvironment([dangerFile], [undefined], runtimeEnv)
+
+  process.stdout.write(JSON.stringify(runtimeEnv.results, null, 2))
 }
 
 // Wait till the end of the process to print out the results. Will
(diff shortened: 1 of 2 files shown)


LarsJK:
Looks like I broke async running.. Will have another go at it..

orta:
The node 9 runner is the runner which runs Danger like a consumer CI, it looks like the JSON it thinks is the response from the subprocess is `{}` ([logs](https://travis-ci.org/danger/danger-js/jobs/384528573#L793-L805)).

screen shot 2018-05-28 at 8 04 47 am
 

Could it be that more than one JSON object is being sent over, and getting to the results processor here: https://github.com/danger/danger-js/blob/master/source/runner/Executor.ts#L122-L125

PR #613, opened 2018-06-13
Use `pullRequestParser` in TeamCity provider

markelog:
If one uses Bitbucket Server + TeamCity, parsing of the
PULL_REQUEST_URL environment variable will be incorrect.

It seems `pullRequestParser` module already does a pretty good
job in figuring out what's what, so use it in TeamCity provider implementation


markelog:
Oh and btw, should we add info about `PULL_REQUEST_URL` variable here – https://danger.systems/js/usage/bitbucket_server.html?

markelog:
Also added proper handling of `-` symbols in BitBucker server repo names

orta:
Nice one, yeah, this looks good to me 👍

markelog:
Cool thanks!

PR #619, opened 2018-06-22
Use Exec instead of spawn in danger local

orta:
WIP on  #571 

got a crash though:

```
~/dev/projects/danger/danger-js exec_to_spawn*
❯ yarn tsc -p tsconfig.production.json; node distribution/commands/danger.js local --dangerfile dangerfile.lite.ts --base 7cf25a13730309fe0773c597f2ae0f1e22a1d4ab
yarn run v1.7.0
$ /Users/orta/dev/projects/danger/danger-js/node_modules/.bin/tsc -p tsconfig.production.json
✨  Done in 5.29s.
Error:  { SyntaxError: JSON5: invalid character 'A' at 9:317
    at syntaxError (/Users/orta/dev/projects/danger/danger-js/node_modules/json5/lib/parse.js:1:12963)
    at invalidChar (/Users/orta/dev/projects/danger/danger-js/node_modules/json5/lib/parse.js:1:12177)
    at Object.afterPropertyValue (/Users/orta/dev/projects/danger/danger-js/node_modules/json5/lib/parse.js:1:8133)
    at Object._default [as default] (/Users/orta/dev/projects/danger/danger-js/node_modules/json5/lib/parse.js:1:2255)
    at lex (/Users/orta/dev/projects/danger/danger-js/node_modules/json5/lib/parse.js:1:1680)
    at Object.parse (/Users/orta/dev/projects/danger/danger-js/node_modules/json5/lib/parse.js:1:992)
    at Object.<anonymous> (/Users/orta/dev/projects/danger/danger-js/distribution/platforms/git/localGetCommits.js:73:33)
    at step (/Users/orta/dev/projects/danger/danger-js/distribution/platforms/git/localGetCommits.js:40:23)
    at Object.next (/Users/orta/dev/projects/danger/danger-js/distribution/platforms/git/localGetCommits.js:21:53)
    at /Users/orta/dev/projects/danger/danger-js/distribution/platforms/git/localGetCommits.js:15:71 lineNumber: 9, columnNumber: 317 }
Error:  { SyntaxError: JSON5: invalid character 'A' at 15:317
    at syntaxError (/Users/orta/dev/projects/danger/danger-js/node_modules/json5/lib/parse.js:1:12963)
    at invalidChar (/Users/orta/dev/projects/danger/danger-js/node_modules/json5/lib/parse.js:1:12177)
    at Object.afterPropertyValue (/Users/orta/dev/projects/danger/danger-js/node_modules/json5/lib/parse.js:1:8133)
[...]

PR #621, opened 2018-06-27
Exec to spawn

joshacheson:
WIP, Branched from https://github.com/danger/danger-js/pull/619

### DIFF
--- a/source/platforms/git/localGetCommits.ts
+++ b/source/platforms/git/localGetCommits.ts
@@ -14,7 +14,7 @@ const authorDate = "%ai"
 const committerName = "%cn"
 const committerEmail = "%ce"
 const committerDate = "%ci"
-const message = "%s" // this is subject, not message, so it'll only be one line
+const message = "%f" // this is subject, not message, so it'll only be one line
 
 const author = `"author": {"name": "${authorName}", "email": "${authorEmail}", "date": "${authorDate}" }`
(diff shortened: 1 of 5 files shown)


source/platforms/git/localGetCommits.ts
```diff
@@ -14,30 +14,33 @@ const authorDate = "%ai"
 const committerName = "%cn"
 const committerEmail = "%ce"
 const committerDate = "%ci"
-const message = "%s" // this is subject, not message, so it'll only be one line
+const message = "%f" // this is subject, not message, so it'll only be one line
```
joshacheson: @orta this is the crux of it - curious if you're comfortable with this solution?

Relevant doc for reference: https://git-scm.com/docs/pretty-formats

orta: Yep, seems totally fine to me!

orta:
Test fails are legit 👍

joshacheson:
@orta fixed what I could, seems that the last remaining failure might be... not that legit?

orta:
Agreed - rerunning that test, looks like doing `create-react-app init` failed which is definitely unrelated 👍

orta:
merge on green

orta:
OK, this got deployed in the latest release 👍

joshacheson:
Thank you! Hope it solves the problem I was having :D

PR #626, opened 2018-07-02
add option to localGetCommits to support timeAgo and filePath fix #625

### LINKED ISSUE #625: Want to get recent commits within x number of weeks/months/years etc.
## Background
I work on a team of 10+ individuals in the same code base and we want a way to identify risky PRs, one of the suggestions was to write a danger rule to flag a file if there are multiple or more commits within a given period (say 3 weeks)

## Proposal
I think we should add something similar to
```
export const localGetCommits = (base: string, head: string) =>
  new Promise<GitCommit[]>(done => {
    const args = ["log", `${base}...${head}`, `--pretty=format:${formatJSON}`]`
   ...
```
but instead allow for the option `log --since=${timeAgo}`. With a default option of 3 weeks

thinking of calling the method `getRecentLocalCommits` or following naming convention here: `localGetRecentCommits = (base: string, head: string, timeAgo: string = 3 weeks) => ...`

then when writing the danger rule, I could get the length of recent commits returned by the promise.

timurista:
add option to localGetCommits to support timeAgo and filePath fix #625.

I decided it was a better approach to tweak this file than to make a new one since the logging information and returned list of commits is basically the same. Just want some more fine grained control over selecting date ranges and limiting to single file.

Note: No Readme exists for localGetCommits so it might be worth adding one in the future to document the api better.

### DIFF
--- a/source/platforms/git/localGetCommits.ts
+++ b/source/platforms/git/localGetCommits.ts
@@ -19,10 +19,17 @@ const message = "%f" // this is subject, not message, so it'll only be one line
 const author = `"author": {"name": "${authorName}", "email": "${authorEmail}", "date": "${authorDate}" }`
 const committer = `"committer": {"name": "${committerName}", "email": "${committerEmail}", "date": "${committerDate}" }`
 export const formatJSON = `{ "sha": "${sha}", "parents": "${parents}", ${author}, ${committer}, "message": "${message}"},`
+export type localCommitOptions = {
+  filePath?: string
+  timeAgo?: string // e.g. 3.weeks
+}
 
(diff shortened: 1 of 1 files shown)


orta:
Yep, cool, this make sense - this function isn't exposed to a user in the DSL, so just adding this will not do anything for users - it would need to be exposed on the [GitDSL](https://github.com/danger/danger-js/blob/master/source/platforms/git/gitJSONToGitDSL.ts#L50)

timurista:
@orta we decided to do something different for our internal review process so we don't need the list of recent commits. Closing this PR. Thank you for your help

PR #630, opened 2018-07-10
Removes a potential crash

### LINKED ISSUE #629: `danger pr` and `danger ci` crash with lodash import error
When I:
1. Install `danger@3.7.20` (using npm 6.0.0 fwiw)
2. Create a Dangerfile
3. Add an npm script `"danger": "danger pr"`,
4. Run `env DANGER_GITHUB_API_TOKEN=<redacted> npm run danger <PR URL>`

danger crashes with the following:

```sh
Error:  TypeError: lodash_1.includes is not a function
    at /node_modules/danger/distribution/platforms/git/diffToGitJSONDSL.js:13:77
```

And, looking at that file, I'm not sure why it _wouldn't_ do that: it [runs](https://github.com/danger/danger-js/blob/f7ecf6fa516831e5e1e0945614e2658cef267e1b/source/platforms/git/diffToGitJSONDSL.ts#L2)
`import { includes } from "lodash"`, but this package does not install `lodash`. Instead it installs [`lodash.includes`](https://github.com/danger/danger-js/blob/f7ecf6fa516831e5e1e0945614e2658cef267e1b/package.json#L126). And it looks like [all other places in the code](https://github.com/danger/danger-js/search?q=lodash&unscoped_q=lodash) `import includes from "lodash.includes"` instead. So, should `diffToGitJSONDSL.js` also do that? And how is danger not crashing for y'all?

orta:
Fixes #629


wearhere:
Yay excited for this! Looks like just have to get the tests passing? I can't quite tell what the failure is about.

PR #638, opened 2018-07-20
Spawn 200kb size limit bug

banzalik:


orta:
OK, yeah, that makes senes 👍

PR #644, opened 2018-07-31
Use the peril bot ID

orta:
This might fix #642 

Think I was hesitant to have danger know about peril too much when I first built this, but we're pretty far from that now.


orta:
merge on green

PR #645, opened 2018-08-01
Adapt emoji for Bitbucket server to fix "Database error has occurred"

acecilia:
Bitbucket server only supports emojis in recent versions. For example, Bitbucket v4.10.1 does not support them, and when commenting with them the server returns an error saying "Database error has occurred".

Implementation similar to https://github.com/danger/danger/commit/638943df2bb19b6ada140acd8c5a72eac2ae5f95#diff-4b908d596ce5ef33cce2a524a5694a4f


CHANGELOG.md
```diff
@@ -789,12 +794,12 @@ If these files are supposed to not exist, please update your PR body to include
 
   ```js
   // In danger 0.16.0:
-  const fullDiff = danger.git.diffForFile('foo.js')
-  const addedLines = danger.git.diffForFile('foo.js', ['add'])
-  const removedLines = danger.git.diffForFile('foo.js', ['del'])
+  const fullDiff = danger.git.diffForFile("foo.js")
```
acecilia: I did not add this changes, I think the hooks that the project has are causing this.

orta: cool 👍

source/runner/templates/bitbucketServerTemplate.ts
```diff
@@ -73,9 +82,9 @@ export function inlineTemplate(dangerID: string, results: DangerResults, file: s
   return `
 [//]: # (${dangerIDToString(dangerID)})
 [//]: # (${fileLineToString(file, line)})
-${results.fails.map(printViolation("🚫")).join("\n")}
-${results.warnings.map(printViolation("⚠️")).join("\n")}
-${results.messages.map(printViolation("📖")).join("\n")}
+${results.fails.map(printViolation(noEntryEmoji)).join("\n")}
+${results.warnings.map(printViolation(warningEmoji)).join("\n")}
+${results.messages.map(printViolation(messageEmoji)).join("\n")}
```
orta: Yep - this all makes sense 👍

CHANGELOG.md
```diff
@@ -13,6 +13,11 @@
 
 ## Master
 
+# 3.8.3
```
orta: Adding this means likely merge conflicts, best to leave them next time - I normally add this version when I create the new release. I should make a danger rule for that 👍

acecilia: I can avoid adding changelog for the next PR. Do you prefer that?

PR #646, opened 2018-08-01
Bitbucket fixes

acecilia:
* Fix error when trying to obtain a response json when the response code is 204 (which means that there is no response)
* Fix link of the PR status, so it opens the web version of the PR, pointing to the Danger comment


CHANGELOG.md
```diff
@@ -13,6 +13,12 @@
 
 ## Master
 
+# 3.8.3
+
+- Fix error when trying to obtain a response json when the response code is 204 (which means that there is no response).
+  Fix link of the PR status, so it opens the web version of the PR, pointing to the Danger comment
+  [646](https://github.com/danger/danger-js/pull/646) - [@acecilia](https://github.com/acecilia)
+
```
acecilia: The hooks that the project has are putting each of the sentences here in a new line

orta: No worries, if they are different fixes then they can be different line items, I’ll fix it pre shipping

orta:
Yep, cool, merge on green

PR #649, opened 2018-08-04
Adds the potential to override the custom module handler in peril

orta:
So I can add local module resolution


orta:
merge on green

PR #658, opened 2018-09-04
Add CI integration for Concourse

Cwright017:
Adds integration for Concourse CI.

### DIFF
--- a/source/ci_source/providers/Concourse.ts
+++ b/source/ci_source/providers/Concourse.ts
@@ -5,6 +5,29 @@ import { ensureEnvKeysExist, ensureEnvKeysAreInt } from "../ci_source_helpers"
  * Concourse CI Integration
  *
  * https://concourse-ci.org/
+ *
+ *  ### CI Setup
+ *
+ *  With Concourse, you run the docker images yourself, so you will want to add `yarn danger ci` within one of your build jobs.
+ *
+ *   ``` shell
+ *    build:
+ *      image: golang
+ *        commands:
+ *          - ...
+ *          - yarn danger ci
+ *   ```
+ *
+ *  ### Environment Variable Setup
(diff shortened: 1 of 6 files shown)


source/ci_source/providers/Concourse.ts
```diff
@@ -0,0 +1,37 @@
+import { Env, CISource } from "../ci_source"
+import { ensureEnvKeysExist, ensureEnvKeysAreInt } from "../ci_source_helpers"
+
+/**
+ * Concourse CI Integration
+ *
+ * https://concourse-ci.org/
+ */
```
orta: This is all great so far - any chance you can add the full template for setting up concourse here? I've never used it - the format should look something like [this travis](https://github.com/danger/danger-js/blob/065b2a5c9ea8518590ba487b6cbdc9e6c4beae89/source/ci_source/providers/Travis.ts#L3-L24) or this [Surf](https://github.com/danger/danger-js/blob/master/source/ci_source/providers/Surf.ts#L4-L14) one. These docs end up on the website being used by anyone wanting to get set up

Cwright017: Sure, no problem. There are two ways to set this up on Concourse, an easy way ( similar to the rest of the providers ' run yarn danger ci as part of a job ', or make a concourse resource ( a docker container specific to a job that concourse uses ). The latter is the more concourse way to do it, but I first have to create the resource. 

My plan was to document the first way, and then amend it to include a reference to the resource once I've created it, is this ok?

Cwright017:
Made a comment RE the documentation - https://github.com/danger/danger-js/pull/658#discussion_r215185974 

Posting here as the comment got minimised with the previous change.

orta:
Awesome, yes!

PR #662, opened 2018-09-13
Update to babel 7 stable

adamnoakes:
Updated to use preset-env with target of node 6 (i assume this is the lowest supported as jest won't run below 6). Ran the tests in a node 6 environment.

Not sure if this is major/ minor please advise and i'll update the changelog


orta:
Aha, I see, so the runtime danger wouldn't support babel7 because of the require paths etc.

I think I'll go with a semantic version major bump here. Existing installs may have different behavior after updating. 

Thanks!

orta:
I'll handle the changelog with the release, don't worry about it 👍

PR #669, opened 2018-09-22
Fix missing PR status bug

johansteffner:
Since v3.6.0 danger doesn't post a updated status if there are no failures or messages. This has broken our build pipeline since we have Danger set as a required step in our repositories.

This update should fix that issue by moving the update status call outside of `failureCount + messageCount === 0` again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 
 ## Master
 
+- Fixed a bug where Danger would fail to update status when there are no failures or messages [@johansteffner][]
+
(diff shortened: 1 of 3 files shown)


fwal:
Nice! 👍 I think we should also update or add unit tests so we have this case covered now and in the future. 🙂

johansteffner:
Added a test case for empty result.

orta:
Oh, Great catch! 

Thanks for the test, I'll ship this now

PR #671, opened 2018-09-26
Add a script to package Danger JS to distribute via Homebrew

### LINKED ISSUE #437: Standalone Danger
We are considering to migrate from the Ruby version of Danger to `danger-js` with `danger-swift` in our iOS projects so we can start writing our Dangerfile in Swift, but having to install Node.js as an extra dependency is holding us back. What do you think about distributing a standalone version of it so it can be installed via Hombrew? 

Using [pkg](https://github.com/zeit/pkg) is a possible solution that I know of. I can throw in a PR if you think this is fine.

thii:
Fixes #437.


orta:
Hi @thii - great work, this is a good start! 

To have this merged, and for me to do it on every deploy, it'll need to fully be automated. E.g. uploading as a GitHub release, and including sending the home-brew PR otherwise support will be sloppy. I've done this before, but [in Ruby here](https://github.com/CocoaPods/CocoaPods-app/blob/master/Rakefile#L1138-L1260) if you want an example to work from

thii:
Thanks @orta! Just one question: Are you using release-it to release a new version? I saw it in the dependencies but the release process in the documentation is different.

orta:
Ah yeah, I'm using release it, only just started, so I've not updated docs yet 👍

thii:
@orta Can you take another look? CI failed on Node 7 because of `create-react-app` now requires Node 8 or later.

thii:
Build passed. Is there anything left that needs to be addressed?

orta:
Validated that it works too:

screen shot 2018-10-10 at 8 51 49 am

orta:
I updated the danger-swift homebrew dependency to include this now, so the setup process on a Mac should now be:

```sh
brew install danger/tap/danger-swift
danger process danger-swift
``` 

nice work 👍

PR #676, opened 2018-09-29
Closes #675: Remove suggestion to display token in travis build log.

### LINKED ISSUE #675: Travis documentation suggests "Display [API token] in build log": seems insecure
I was going through [the getting started guide](https://danger.systems/js/guides/getting_started.html) and saw this section for Travis:

image

In particular:
> If you have an open source project, you should ensure "Display value in build log" enabled, so that PRs from forks work.

If I understand correctly, this will insert your `DANGER_GITHUB_API_TOKEN`'s contents into the build log. For open source projects, build logs are often public, e.g. https://travis-ci.org/mozilla-mobile/focus-android/builds/434818832 This means anyone can find and then use your API token: I would recommend removing these lines.

It looks like this would need to be removed from the CLI init and the Travis documentation: https://github.com/danger/danger-js/search?q=display+value+in+build+log&unscoped_q=display+value+in+build+log

mcomella:
It's insecure to post API tokens to public build logs and most open
source repositories have public build logs. However, I do not
understand the effects not enabling this option would have on users of
danger.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 -->
 
 ## Master
+- Remove suggestion from docs and init tool to display GitHub API token in travis build logs [@mcomella][]
 
 # 4.0.1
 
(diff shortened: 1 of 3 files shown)


orta:
This would mean Danger would only work on local branches, and not on fork PRs. - so danger would only work for people with write access to the repo. 

This isn't really how most people would expect it to work, so I recommended the token be available.

orta:
Happy to change wording though, this is explained in the getting started section of the docs: https://danger.systems/js/guides/getting_started.html#tokens-for-oss-projects

PR #690, opened 2018-10-10
Allow repository name parsing to have extra valid characters

saamorim:
While setting up the danger environment to use bitbucket server Ive notice that url like 
https://someserver.com/projects/PRJ/repos/com.somewhere.library/pull-requests/10/overview
where not parsed. 
The regex for the repo name, in this case com.somewhere.library, was only allowing \w+ and -.

This PR extends to allow _ and .

Tests and linters are all green


orta:
Yep, cool, merge on green

PR #691, opened 2018-10-11
Support Multibranch pipelines in Jenkins

### LINKED ISSUE #379: Getting "Skipping Danger due to not this run not executing on a PR." on PR builds on Jenkins using multi branch pipeline jobs
Looks like the implementation assumes usage of the Github Pull Request Builder, which is a handy plugin for legacy builds. I think the push now is to move to pipeline jobs, specifically to things like Github Organization Folders which scan github orgs for matching repos using a regexp and generating multi branch pipeline jobs underneath it.

In that scheme, there's no `ghprbPullId` and `ghprbGhRepository` env variables as looked for here: https://github.com/danger/danger-js/blob/09dcbd74e1b5df3e1a4a2058b716eb993181a25c/source/ci_source/providers/Jenkins.ts#L40-L42

Personally, I hacked a workaround in my `Jenkinsfile` like so:
```groovy
withEnv(['ghprbGhRepository=organization/reponame',"ghprbPullId=${env.CHANGE_ID}"]) {
  sh 'npx danger'
}
```

So I think `CHANGE_ID` and `ghprbPullId` are equivalents, but I don't think there's an exact equivalent of `ghprbGhRepository ` produced, though I believe you could parse it out of `CHANGE_URL`.

azz:
Jenkins is ❤️ 

Fixes #379


azz:
Strange typescript errors only on Node 9? https://travis-ci.org/danger/danger-js/jobs/440092950

orta:
OK, cool, yeah, I'll get this merged and shipped, and the TS error handled too - I know what's going on there.

PR #696, opened 2018-10-16
use import/require for octokit as types are not provided

flovilmart:
Recent version of typescript broke (somehow) the GitHub exports with prevents a clean upgrade on peril (https://github.com/octokit/rest.js/issues/1041)

This PR attempts to resolve this issue. In a sense, it is better to use this as ts supports it and octokit is not written with imports in mind.


orta:
This doesn't feel right for me - I think it's a tsconfig issue somewhere, I don't have time till tonight to take a look at this though

flovilmart:
Yeah, it’s odd, I can look further. Odd thing, is that it breaks in peril, when importing the modules, not here.

orta:
I wonder if Danger's tsconfig has a different module setting

flovilmart:
@orta found the solution, using [allowSyntheticDefaultImports](https://github.com/danger/peril/pull/391/commits/78202fa177ee2d7d05f6fc90169e97554ab38302#diff-e5e546dd2eb0351f813d63d1b39dbc48R4) in peril to silence the warnings.

https://www.typescriptlang.org/docs/handbook/compiler-options.html

PR #698, opened 2018-10-16
Adds support for using danger-js in an action

orta:
This means you can use danger/danger-js in a GitHub Action


orta:
merge on green

PR #699, opened 2018-10-16
Fix: error thrown undefined comment

flovilmart:
It can occur that the comment is passed as undefined and the method `deleteInlineComment` throws an exception attempting to read `.id` on undefined.

This patch ensure this doesn't happen but I haven't found the root cause. I seems to occur when not responding any message on an issue event.


orta:
👍

orta:
feel safe either way

flovilmart:
I’ll probably dig to the root cause at one point

PR #701, opened 2018-10-18
Add Proxy support

steprescott:
This PR adds the ability to proxy any of the network requests made by danger to be proxied.

It uses the standard `HTTP_PROXY` & `HTTPS_PROXY` environment variables and if either of them is set will then proxy those requests.

To ensure the tests pass a server is created so the request can be handled but dose no assertion to ensure the request was proxied. The only addition I have added is adding the `Agent` to the request and so I have only tested that an `Agent` is set on the request, not that the request can be successfully proxied as that should be covered by the `http-proxy-agent` package.

However when running `danger pr` with either `HTTP_PROXY` or `HTTPS_PROXY` set I can see that the requests are being proxied via Charles.
screenshot 2018-10-18 at 12 36 10

More details can be found [here](https://github.com/danger/peril/issues/364).


package.json
```diff
@@ -129,7 +130,6 @@
     "commander": "^2.18.0",
     "debug": "^4.0.1",
     "get-stdin": "^6.0.0",
-    "https-proxy-agent": "^2.2.1",
```
orta: I think you just removed this dependency by accident

orta: It's needed - the `@types` go in dev dependencies and will only occur *in this repo*. Not in a consumer of danger

orta:
Other than that - looks good to me

steprescott:
Thanks for the review. I've acted on your comments. Any more issue let me know. Once this is in I will look at adding the same change to Peril.

orta:
👍

PR #704, opened 2018-10-23
Fix posting status update on Github PRs

### LINKED ISSUE #614: Upgrade from 2.x to 3.7.17, fails to set Github status
I had a working setup before using Danger.js with Github (using an access token). After upgrading from 2.x to 3.7.17, that setup broke. I modified the build to go from `npx danger` to `npx danger ci`, which I saw in the breaking changes.

However, now when danger runs, it updates the comment on the PR with the fails/warnings/messages, but the status is never updated. There's no indication of an issue in the build logs (from the output of `npx danger ci`).

I think the refactoring that added checks support may have broken the older issue comment/status workflow on Github.

sgtcoolguy:
Fixes #614

- Fix posting status update on Github PRs. 
- Fix handling of 'pending' status on updateStatus(). 
- Add unit tests around GithubAPI's updateStatus() impl

This basically reverts back to delegating to `GithubAPI` to perform the status update on a PR using the issue commenter workflow. (this means that the nice octokit API isn't used, but this does centralize the API operations in the right place, rather than having two implementations - one broken, the other unused).

The underlying issue was that the new implementation in IssueCommenter was using the head she (correct), but also the head repo (should be base). The old impl in GithubAPI was correct here.

Additionally, while the platforms' api states the status could be `"pending"` that was never handled properly here. I'll open a PR to fix that for Bitbucket too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 
 ## Master
 
+- Fix #614 - Posting status updates to Github using issue workflow broken - [@sgtcoolguy][]
+
 # 5.0.1, err. 6.0.0
 
 - Hah, my computer ran out opf power mid-deploy, and now I have to ship another build to make sure the brew versions of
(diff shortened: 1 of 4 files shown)


orta:
Yeah -  this make sense, thanks @sgtcoolguy

PR #705, opened 2018-10-23
Fix handling of 'pending' status for BitBucket's updateStatus()

sgtcoolguy:
While the API declared the status as a boolean or `"pending"`, it didn't actually handle it properly and forward to the API state of `"INPROGRESS"`

(The fix for Github is in #704)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 
 ## Master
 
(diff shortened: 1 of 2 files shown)


orta:
Awesome, thanks, merge on green 👍

orta:
I'm alright with skipping the consolidation for now, its a good call for someone working in the area though

Issue #240, opened 2017-05-05
Render markdown inside violations


orta:
Ideally I should be able to write:

```js
import {message} from "danger"

message("*b* _i_ [link](https://github.com/artsy/emission/pull/505/)") 
```

And get:

> *b* _i_ [link](https://github.com/artsy/emission/pull/505/)

orta:
I think if we leave a space between the table and user's content we get it for free. 

E.g

```
<table>
...
<tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/components/artist/artworks/__tests__/index-tests.tsx`
- `src/lib/components/artwork_grids/__tests__/infinite_scroll_grid-tests.tsx`
- `src/lib/containers/__tests__/works_for_you-tests.tsx`

If these files are supposed to not exist, please update your PR body to include "Skip New Tests".</td>
    </tr>
  </tbody>
</table>
```

<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Fails</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Please assign someone to merge this PR, and optionally include people who should review.</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/components/artist/artworks/__tests__/index-tests.tsx`
- `src/lib/components/artwork_grids/__tests__/infinite_scroll_grid-tests.tsx`
- `src/lib/containers/__tests__/works_for_you-tests.tsx`

If these files are supposed to not exist, please update your PR body to include "Skip New Tests".</td>
    </tr>
  </tbody>
</table>

macklinu:
Would it make sense to use a module like [marked](https://github.com/chjj/marked) to convert each message from markdown to HTML instead of adding spaces? I tried out the spaces on a Danger comment in an internal repo at work, but in order to get the markdown to render, I had to add a couple of new lines before and after the message, which just added unnecessary whitespace to the danger table.

I'm going to open a PR for this for discussion. 👍

macklinu:
I played around with marked for a bit. I think it would work well for most cases expect for highlighted code blocks.

[...]

PR #754, opened 2018-11-05
Prevent GitHub adding a <p> tag to <td> inside issue comment.

patrickkempff:
Thanks for this awesome project. It looks awesome! There was one thing that caught my attention; the vertical alignment of content in the table seemed somewhat off. 

Before:
screen shot 2018-11-05 at 12 44 58

After:
screen shot 2018-11-05 at 12 43 59

This PR fixes that.

After comparing the implementation and the code that GitHub renders. I found that the GitHub markdown implementation adds an <p> tag inside the <td> when it thinks the text should be a text block (because of the linebreak before and after `${message}`). When it appends a `<p>` the user-agent add the default margin at the bottom of the paragraph. (as an example; chrome adds margin-bottom: 16px;) This messes up the vertical alignment.

I removed the linebreaks and github andeverything seems to render everything correctly.

HTML output by github:

Before:
screen shot 2018-11-05 at 13 01 01

After:
screen shot 2018-11-05 at 13 03 22

**Important:** I also updated the `githubIssueTemplates.test.ts.snap` snapshot reflect the changes.

PS. @orta I am sure this is not the right way but I wanted to thank you and the artsy crew for the inspiring talks @facebookHQ in london last July.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 
 ## Master
 
+- Fix vertical alignment in GitHub issue template - [@patrickkempff][]
+
 # 5.0.1, err. 6.0.0
 
 - Hah, my computer ran out opf power mid-deploy, and now I have to ship another build to make sure the brew versions of
--- a/source/runner/templates/_tests/__snapshots__/_githubIssueTemplates.test.ts.snap
+++ b/source/runner/templates/_tests/__snapshots__/_githubIssueTemplates.test.ts.snap
@@ -10,7 +10,7 @@ exports[`generating inline messages Shows correct message for multiple inline vi
   DangerID: danger-id-blankID;
   File: File.swift;
   Line: 10;
--->  
+-->
 - :no_entry_sign: Test message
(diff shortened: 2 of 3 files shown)


orta:
Hrm, part of me feels like this is intentional (otherwise why would it be so weird?) - let's give it a shot anyway, and see what happens.

patrickkempff:
> Hrm, part of me feels like this is intentional (otherwise why would it be so weird?) - let's give it a shot anyway, and see what happens.

After some searching around I found an commit from 2 years ago by @macklinu.
This discussion seems to be related: https://github.com/danger/danger-js/issues/240#issuecomment-299481579. I am not sure if it is still relevant

orta:
It probably is, I wonder if maybe we will need to do multi-line detection or something like that to make sure it handles inline html

patrickkempff:
@orta maybe we should revert this PR until we have a better solution handling both cases. What do you think?

orta:
It's not shipped yet, so maybe we can look to see if the string includes a few particular chars and just include that?

PR #761, opened 2018-11-12
Improve logs and verifies --process on the CI

orta:
I want to know if the danger swift issues - want to be certain about https://github.com/danger/swift/issues/97


source/commands/utils/runDangerSubprocess.ts
```diff
@@ -50,15 +52,15 @@ export const runDangerSubprocess = (
     const maybeJSON = getJSONFromSTDOUT(stdout)
     const maybeJSONURL = getJSONURLFromSTDOUT(stdout)
 
-    if (!results && maybeJSONURL) {
+    if (!maybeJSON && !maybeJSONURL) {
+      // Pass it back to the user
+      console.log(data.toString())
+    } else if (!results && maybeJSONURL) {
```
orta: This orders the logs correctly

PR #768, opened 2018-11-18
Declares the html_url on the pr json

orta:
This always existed, but to use it you had to remove the type system.


orta:
merge on green

Issue #767, opened 2018-11-15
local mode don't work without credentials


langovoi:
When I try to use `danger local` without BitBucket or GitHub credentials i got error on [this line](https://github.com/danger/danger-js/blob/master/source/commands/danger-runner.ts#L51)

```sh
The DANGER_GITHUB_API_TOKEN/DANGER_BITBUCKETSERVER_HOST environmental variable is missing
```

orta:
Ah yeah, this is likely an oversight because I have `DANGER_GITHUB_API_TOKEN` in my ENV, it should be passing in the FakeCI env vars from the command

PR #769, opened 2018-11-20
Allow sub-processes to request DSL JSON

orta:
Ideally should allow a fix for https://github.com/danger/swift/issues/108  to be created on danger-swift side
Also includes a fix for #767


orta:
merge on green

PR #778, opened 2018-11-27
Remove the changelog step from a deploy

orta:
This is mainly to remove the final manual step from a deploy.


orta:
cool, no need for a changelog entry - merging.

PR #779, opened 2018-12-04
Add support for COPY change type to fix BitBucket Server

sebinsua:
I noticed that [since this PR](https://github.com/danger/danger-js/commit/f680b5f65c32f339a21a242462c890e829dde9dc#r31538980) some of our PRs on BitBucket Server cause Danger to throw the error: `"Unhandled change type"`.

This seems to be because BitBucket has an extra change type `COPY` which appears when it detects that a file has been added which is 'similar' to another file that already exists in the repository.

[The `ChangeType`s are defined here](https://docs.atlassian.com/bitbucket-server/javadoc/4.1.0/api/reference/com/atlassian/bitbucket/content/ChangeType.html).

There was also an `UNKNOWN` change type. I don't know when this can occur: I added it for correctness sake and because it allowed me to put the `${value.type}` into the error that will be thrown if a new change type is added by BitBucket Server in future. This should make it slightly easier to debug any future additions of change types.

/cc @langovoi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
 
 ## Master
 
+- Add support for `COPY` change type to fix a BitBucket Server regression in
+  [danger/danger-js#764](https://github.com/danger/danger-js/pull/764) - [@sebinsua][]
+
 # 6.1.8
 
 - Revert removal of implicit `<p>` tag from [danger/danger-js#754](https://github.com/danger/danger-js/pull/754) and add
--- a/source/danger.d.ts
+++ b/source/danger.d.ts
@@ -306,8 +306,8 @@ interface BitBucketServerPagedResponse<T> {
   values: T
 }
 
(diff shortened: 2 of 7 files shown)


sebinsua:
@orta I had to cheat to get the build to pass by using `--no-verify` when commiting the `danger.d.ts` file.

Otherwise, it would fail in the `Danger-JS-Test-App` check telling me I had to run `yarn declarations` and when I did this it would fail in exactly the same way.

I think that what was happening is that whatever Danger rule that is providing that check does not run `prettier` against this file before testing that it is the same?

orta:
Interesting, it should be running prettier against it - but that's my problem not yours, I think this is 👍 
`

sebinsua:
Thanks. This is an amazing piece of open-source btw!

It's unlocking so many possibilities to improve PR processes at the places I consult at.

orta:
Thanks! Yeah, this should be in the latest release

Issue #168, opened 2017-03-14
feat(template): support commit stamp in message


orta:
re:  #97

Basically this:
![1a8e4372-db2e-11e6-94c2-287ed77c0cc5](https://cloud.githubusercontent.com/assets/49038/23926278/a12d3568-0956-11e7-9ee7-05e6f0341657.png)

orta:
@ashfurrow ^ might be worth a shot

PR #783, opened 2018-12-08
Add commit ID to comment footer (#168)

randak:
Addresses #168. This PR reads the latest commit's ID and adds it to the footer of danger comments.

The documentation for both Github and Bitbucket indicate that commit SHAs should be shortened and turned into links automatically, but I haven't had the opportunity to test this in a real CI environment.

This is my first PR for Danger, so all feedback is welcome!


source/platforms/github/GitHubAPI.ts
```diff
@@ -93,7 +93,7 @@ export class GitHubAPI {
     return allComments
       .filter(comment => v.includes(comment.body, dangerIDMessage)) // does it contain the right danger ID?
       .filter(comment => comment.user.id === userID) // Does it have the right user ID?
-      .filter(comment => v.includes(comment.body, dangerSignaturePostfix)) // Does it look like a danger message?
+      .filter(comment => v.includes(comment.body, dangerSignature)) // Does it look like a danger message?
```
randak: I'm wondering if this filter is even necessary. I'm  not a huge  fan of exporting a different version of the signature just to test if it is present or not, but I didn't want to check the commit ID in here either.

randak:
So my understanding from Github is that including a full-length sha will automatically create a short SHA link. It actually appears to automatically turn into `preformatted code` as well, so I'm going to update this to remove the backticks I manually placed.

This is what happens when you put an unformatted full-length SHA in a Github comment: 28bd18ca866e765c985d94d5f9544834b95c24e9

For Bitbucket, I'm not so sure. I know that the commit hash should be autolinked from [this support article](https://confluence.atlassian.com/bitbucket/mark-up-comments-305037452.html), but it's not clear if it'll get shortened or not. I don't have access to any Bitbucket accounts so I'm not sure how to validate this. If you don't have a way to test, I can also just truncate to 8 characters to match the example in the Bitbucket help article.

orta:
Perfect, let's assume the best WRT bitbucket, and it can change if someone says it's verbose 👍

PR #786, opened 2018-12-10
Update GITHUB_EVENT_TYPE to GITHUB_EVENT_NAME for Actions

### LINKED ISSUE #782: DangerJS fails in github actions
Hello!  Trying to figure out the best way to run danger via GitHub actions.  I have gotten to the point of the action running and finding my dangerfile no problem, but the `danger.github` API is coming back `null`.

I have seen this happen when I run the `danger local` command as well, where the other namespaces (eg `danger.git`) are fine, but clearly there is no PR to reference so the `danger.github` response being `null` makes sense to me.

Here is that config:
```
workflow "Dangerfile JS Eval" {
  on = "pull_request"
  resolves = "Danger JS"
}

action "Danger JS" {
  uses = "danger/danger-js@master"
  secrets = ["GITHUB_TOKEN"]
  args = "--dangerfile ./dangerfile.ts"
}
```

I have also tried just running a typical bash script via github actions (just following the docs, with a `Dockerfile` and an `entrypoint.sh`).  I can run the following script and have the same issue:

```bash
#!/bin/sh -l

set -eu

export DANGER_GITHUB_API_TOKEN="$GITHUB_TOKEN"
yarn global add danger

# Run danger
danger --dangerfile="./dangerfile.ts" ci
```
This yields a `null` value for `danger.github` as well.

My last attempt was to use the `--external-ci-provider "GitHubActions"` and I got this response:
`could not load CI provider at GitHubActions due to Error: ENOENT: no such file or directory, stat '/github/workspace/GitHubActions'`

Again, I can run the `danger pr https://github/path/to/pr` and it runs fine, and that bash script currently works on our `CircleCi`.

Any information/links/examples for Github Actions you can provide would be really helpful!  If there is someway I can help document/contribute let me know.

Thanks!

cysp:
Fixes #782 (for me 🙃)

Unfortunately GitHub Actions' runtime environment doesn't provide `git` so I tested this fix by bundling a snapshot of all of `danger-js` in my app. 🙂

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Add support for `COPY` change type to fix a BitBucket Server regression in
   [danger/danger-js#764](https://github.com/danger/danger-js/pull/764) - [@sebinsua][]
 - Add support for older Babel versions (prior 7) [@sajjadzamani][]
+- Fix detection of GitHub Actions event types [@cysp][]
 
 # 6.1.8
 
@@ -1438,3 +1439,4 @@ Not usable for others, only stubs of classes etc. - [@orta][]
 [@langovoi]: https://github.com/langovoi
 [@randak]: https://github.com/randak
 [@sajjadzamani]: https://github.com/sajjadzamani
+[@cysp]: https://github.com/cysp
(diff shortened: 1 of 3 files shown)


orta:
Nice one 👍

orta:
Alright, this is shipped 👍

fbartho:
Thanks @cysp @cysp!

Issue #781, opened 2018-12-06
Danger should support older Babel versions


sajjadzamani:
# Problem
I wanted to import other modules using the es6 `import` in my Dangerfile. I have Babel and Jest setup in my project so I expected the Dangerfile to be transpiled. However, Danger did not transpile my Dangerfile and the `import` statements threw errors.
### Project Dependencies
```JS
"devDependencies": {
    "babel-core": "^6.26.3",
    "babel-eslint": "7.1.1",
    "babel-jest": "^23.6.0",
    "babel-runtime": "^6.26.0",
    "danger": "^6.1.9",
    "danger-plugin-jest": "^1.1.0",
    ...
  }
```
### Debug Output
```
2018-12-06T17:32:02.854Z danger:runner Started Danger runner with

2018-12-06T17:32:02.865Z danger:runner Got STDIN for Danger Run

2018-12-06T17:32:02.867Z danger:jsonToDSL Creating pr DSL from JSON

2018-12-06T17:32:02.924Z danger:localGetDiff > git diff master...HEAD

2018-12-06T17:32:02.939Z danger:localGetDiff > git log master...HEAD --pretty=format:{ "sha": "%H", "parents": "%p", "author": {"name": "%an", "email": "%ae", "date": "%ai" }, "committer": {"name": "%cn", "email": "%ce", "date": "%ci" }, "message": "%f"},

2018-12-06T17:32:02.958Z danger:runner Evaluating dangerfile.js

2018-12-06T17:32:02.959Z danger:transpiler:setup Does not have TypeScript set up

2018-12-06T17:32:02.960Z danger:transpiler:setup Does not have Babel set up

2018-12-06T17:32:02.960Z danger:inline_runner Started parsing Dangerfile:  dangerfile.js

Unable to evaluate the Dangerfile
 dangerfile.js:4
import jest from "danger-plugin-jest";
^^^^^^

SyntaxError: Unexpected token import
    at createScript (vm.js:56:10)
    at Object.runInThisContext (vm.js:97:10)
    at Module._compile (module.js:542:28)
    at Object.requireFromString [as default] (/Users/sajjad/Documents/TapSource/tap-react/node_modules/require-from-string/index.js:28:4)
    at Object.<anonymous> (/Users/sajjad/Documents/TapSource/tap-react/node_modules/danger/distribution/runner/runners/inline.js:144:63)
[...]

PR #788, opened 2018-12-10
Fix babel polyfill import

sajjadzamani:
Relates to #781.
# Issue
The change of #784, introduced an extra dependency on the `babel-polyfill` package, to detect if the dangerfile should be transpiled, if the user had the older version of Babel (`babel-core`) installed.

# Fix
We should always import the version of the polyfill that is installed as part of Danger's dependencies which is`@babel/polyfill`. If the user environment has Babel then this package should be imported no matter the version of Babel installed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,11 @@
 
 <!-- Your comment below this -->
 
+- Fix issue with detecting Babel if `babel-core` is installed [@sajjadzamani][]
+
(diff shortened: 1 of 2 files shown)


orta:
Cool, this works for me (we've got a lot of integration tests about this stuff 👍 )

orta:
Shipped with 6.1.12

PR #789, opened 2018-12-16
Custom runtime naming in danger js sub-processes

orta:
I was looking at this message in Danger Swift https://github.com/danger/swift/pull/140 :

screen shot 2018-12-16 at 11 42 38 am

and thought, hrm "dangerJS, really?" seems a bit of a leaky abstraction. So I've given sub-processes the chance to pass their own name and URL into Danger.


orta:
hah, bad peril - ignoring danger :D

PR #794, opened 2018-12-25
Update octokit

orta:
Hrm, this is a server breaking change I think... So, I think it'll have to happen eventually. Maybe will wait a bit on it.


orta:
merge on green

PR #795, opened 2018-12-26
Fix Danger comment detection

cysp:
The existing fixtures for GitHub API tests don't appear to contain a positive match for this function. Would it be okay to just tweak the existing fixture to include a "Generated by Danger" marker?


source/danger.d.ts
```diff
@@ -323,7 +323,9 @@ interface BitBucketServerChangesValueMove {
   }
 }
 
-type BitBucketServerChangesValue = BitBucketServerChangesValueAddCopyModifyDelete | BitBucketServerChangesValueMove
```
cysp: Danger complained about this file in this PR, but it looks like this formatting is the only thing that changes.

orta:
Yep, this is what happens when I jump between ruby and JS too much. TypeScript is supposed to fix that! Doh well. Thanks @cysp - I've merged master and the current fail is #794 

So I'll merge this and ship it for now.

PR #796, opened 2018-12-26
Fix formatting of danger.d.ts

cysp:
This formatting changed in https://github.com/danger/danger-js/commit/c83dc564a38cd359bf3ddbcd2d17ab34eef18bb3#diff-b80101a8397dd9218bf49711f9f485f0R326 and then it looks like an attempt was made to correct it in https://github.com/danger/danger-js/commit/4a62feaf1b9dbd5a1af8b409a628118836ffcb39 but must have been stymied by one of the git hooks, which also made making this commit a little difficult. 🙃

### DIFF
--- a/source/danger.d.ts
+++ b/source/danger.d.ts
@@ -323,7 +323,9 @@ interface BitBucketServerChangesValueMove {
   }
 }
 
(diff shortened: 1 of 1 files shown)


orta:
This seems to be flip flopping, which  feels weird to me as prettier is determinate but thanks! Yep

PR #798, opened 2018-12-27
Expose option for using GitHub Checks

### LINKED ISSUE #765: Posting GitHub comments instead of populating Checks page?
(I saw https://github.com/danger/danger-js/issues/592, which might actually be closable now. Anyway, I think this is distinct from it.)

Is it possible to flip a switch and have Danger post comments in the PR thread again, instead of hiding them away in the GitHub checks page? I know Danger posts the comment saying "hey, go check out the checks page! You have N warnings and N failures.", but I really liked the old (months-old) behavior of posting in the thread.

(If the solution is "write your own GitHub#comment wrapper", well, I can do that too 😸 – just thought I'd check beforehand.)

cysp:
Closes #765

### DIFF
--- a/source/commands/ci/_tests/runner.test.ts
+++ b/source/commands/ci/_tests/runner.test.ts
@@ -16,6 +16,10 @@ const defaultAppArgs: Partial<SharedCLI> = {
 const source = new FakeCI({})
 const platform = new FakePlatform()
 
+beforeEach(() => {
+  mockRunDangerSubprocess.mockReset()
+})
+
 it("uses the source and platform from config", async () => {
   await runRunner(defaultAppArgs as SharedCLI, { platform, source })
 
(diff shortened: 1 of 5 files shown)


source/commands/ci/_tests/runner.test.ts
```diff
@@ -16,6 +16,10 @@ const defaultAppArgs: Partial<SharedCLI> = {
 const source = new FakeCI({})
 const platform = new FakePlatform()
 
+beforeEach(() => {
+  mockRunDangerSubprocess.mockReset()
```
cysp: Currently each assertion here is actually testing against the arguments passed into the _first_ invocation of the danger runner, not the one that they've just performed.

source/commands/utils/sharedDangerfileArgs.ts
```diff
@@ -38,3 +40,4 @@ export default (command: any) =>
     .option("-c, --external-ci-provider [modulePath]", "Specify custom CI provider")
     .option("-p, --process [command]", "[dev] Runs a custom sub-process instead of the Danger JS runtime")
     .option("-u, --passURLForDSL", "[dev] Use a custom URL to send the Danger DSL into the sub-process")
+    .option("--use-github-checks", "Use GitHub Checks", false)
```
cysp: Whaddayareckon?
I didn't dare try to come up with a single-character option for this.

source/commands/ci/_tests/runner.test.ts
```diff
@@ -25,6 +29,44 @@ it("uses the source and platform from config", async () => {
...
+  await runRunner(customArgs, { platform, source })
+
+  // Pull the executor out of the call to runDangerSubprocess
+  const executor: Executor = mockRunDangerSubprocess.mock.calls[0][2]
+  expect(executor.ciSource).toEqual(source)
+  expect(executor.platform).toEqual(platform)
+  expect(executor.options.disableGitHubChecksSupport).toEqual(true)
+})
+
 // TODO: This occasionally fails!
```
cysp: I suspect that this intermittency was caused by the failure to reset the mock but I didn't want to muddy up this PR with that change. If you concur with that assessment I'll raise a PR unskipping this after the mock resetting has landed. 🙂

source/commands/ci/_tests/runner.test.ts
orta: 👍

source/commands/ci/_tests/runner.test.ts
orta: This makes sense to me - good thinking!

PR #799, opened 2018-12-27
Declare a direct dependency on http-proxy-agent

cysp:
I'm seeing this error when running the latest version of danger-js:
```
internal/modules/cjs/loader.js:582
    throw err;
    ^

Error: Cannot find module 'http-proxy-agent'
```

… it looks like we used to get the package via a transitive dependency via `@octokit/rest`, but that changed with the recent upgrade of that package:
https://github.com/danger/danger-js/commit/085903291bd17107b17a6171ea4919dcdc8e70be#diff-8ee2343978836a779dc9f8d6b794c3b2L719

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,6 +129,7 @@
     "commander": "^2.18.0",
     "debug": "^4.1.1",
     "get-stdin": "^6.0.0",
+    "http-proxy-agent": "^2.1.0",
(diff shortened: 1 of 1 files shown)


orta:
Fixes #801

orta:
Thanks!

PR #800, opened 2018-12-27
Introduce a type for GitHub issue comments

cysp:
Here's some typing for GitHub issue comments, where there used to be a lot of `any` (which was why `.include()` wasn't picked up as an error).

### DIFF
--- a/source/platforms/github/GitHubAPI.ts
+++ b/source/platforms/github/GitHubAPI.ts
@@ -4,11 +4,10 @@ import * as node_fetch from "node-fetch"
 import parse from "parse-link-header"
 import pLimit from "p-limit"
 
-import { GitHubPRDSL, GitHubUser } from "../../dsl/GitHubDSL"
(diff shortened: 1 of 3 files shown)


source/platforms/github/GitHubAPI.ts
```diff
@@ -96,7 +95,7 @@ export class GitHubAPI {
       .map(comment => comment.id) // only return IDs
   }
 
-  updateCommentWithID = async (id: number, comment: string): Promise<any> => {
+  updateCommentWithID = async (id: string, comment: string): Promise<any> => {
```
cysp: The existing `Comment` type has its ids as strings, and I _think_ that these are only internal API, so I figured that this change was okay. Whaddayathink?

orta: That makes sense 👍

orta:
Yep, awesome

Issue #802, opened 2019-01-03
PR Command Failing with GitHub Enterprise


DavidBrunow:
I'm getting the following error message when trying to use Danger Swift with Github Enterprise:

`Could find an arg which mentioned GitHub.`

I can work around this issue by adding `#github` to the end of the PR. I believe this issue was created with this change: https://github.com/danger/danger-js/commit/5b6c2227ed8f11011f76e60659ab6dc76b69a540#diff-5a8723ba15d4301f9d3543b4f434788aR64

orta:
I'm open to having something to indicate that you're definitely using GitHub 👍

DavidBrunow:
I'll look into using the environment variable for `DANGER_GITHUB_HOST` if that exists and if not then fall back to the hard-coded string "github".

orta:
Yeah, that makes sense!

orta:
This got fixed and shipped with #803

PR #803, opened 2019-01-04
Allow PR command to work with GitHub Enterprise

DavidBrunow:
See issue created here: https://github.com/danger/danger-js/issues/802


orta:
Yep, perfect - thanks

Issue #805, opened 2019-01-07
Support GITHUB_URL env variable


Aghassi:
Currently danger has scoped variables:

https://danger.systems/js/guides/getting_started.html#enterprise-github

I think this is fine, but for enterprise use it is a little annoying to have separate GITHUB environment tokens. A number of other repos support the paradigm `GITHUB_URL`. Would it be possible to just add a check that sets the `DANGER_GITHUB_URL` etc internally based on if that `GITHUB_URL` exists?

orta:
Yep, I did this mainly so it is explicit, but I have no problems with also supporting un-prefixed versions 👍

Aghassi:
So looks like there are three files that would require changing https://github.com/danger/danger-js/search?q=%22process.env%5B%22DANGER_GITHUB_API_BASE_URL%22%5D%22&unscoped_q=%22process.env%5B%22DANGER_GITHUB_API_BASE_URL%22%5D%22

Is there anything outside of this repo that would need changing? Otherwise, I can pick this up this weekend most likely.

PR #806, opened 2019-01-12
feat: added GITHUB_URL as possible token

Aghassi:
Closes https://github.com/danger/danger-js/issues/805
Provides the ability to use `GITHUB_URL` as
an environment token


orta:
Yep - merge on green

orta:
Alright, that is in with 7.0.2 👍

PR #818, opened 2019-01-31
Remove 404 styled components link

stevemoser:


orta:
Ah yeah, they all moved to Peril - thanks!

Issue #812, opened 2019-01-25
Unexpected token and octokit deprecation error


kangax:
I'm running a simple example dangerfile.js straight from the docs, using the latest 7.0.2:

```js
import { message, danger } from 'danger';
import { readFileSync } from 'fs';

const modifiedMD = danger.git.modified_files.join('- ');

message(`Changed Files in this PR: \n -  ${modifiedMD}`);
```

and `yarn danger local` throws both syntax error and deprecation notice:

```
> yarn danger local
yarn run v1.13.0
$ <user>/node_modules/.bin/danger local
Error: new Octokit({headers}) is deprecated. Use {userAgent, previews} instead. See https://github.com/octokit/rest.js#client-options
    at parseOptions (<user>/node_modules/@octokit/rest/lib/parse-client-options.js:43:18)
    at new Octokit (<user>/node_modules/@octokit/rest/lib/constructor.js:14:50)
    at apiForDSL (<user>/node_modules/danger/distribution/runner/jsonToDSL.js:114:15)
    at Object.<anonymous> (<user>/node_modules/danger/distribution/runner/jsonToDSL.js:72:23)
    at step (<user>/node_modules/danger/distribution/runner/jsonToDSL.js:43:23)
    at Object.next (<user>/node_modules/danger/distribution/runner/jsonToDSL.js:24:53)
    at <user>/node_modules/danger/distribution/runner/jsonToDSL.js:18:71
    at new Promise (<anonymous>)
    at __awaiter (<user>/node_modules/danger/distribution/runner/jsonToDSL.js:14:12)
    at Object.exports.jsonToDSL (<user>/node_modules/danger/distribution/runner/jsonToDSL.js:65:53)

Error: octokit.authenticate() is deprecated. Use "auth" constructor option instead.
    at authenticate (<user>/node_modules/@octokit/rest/plugins/authentication-deprecated/authenticate.js:4:16)
    at apiForDSL (<user>/node_modules/danger/distribution/runner/jsonToDSL.js:116:13)
    at Object.<anonymous> (<user>/node_modules/danger/distribution/runner/jsonToDSL.js:72:23)
    at step (<user>/node_modules/danger/distribution/runner/jsonToDSL.js:43:23)
    at Object.next (<user>/node_modules/danger/distribution/runner/jsonToDSL.js:24:53)
    at <user>/node_modules/danger/distribution/runner/jsonToDSL.js:18:71
    at new Promise (<anonymous>)
[...]

PR #820, opened 2019-02-02
Updates the usage of octokit to remove warnings

orta:
Re:  https://github.com/danger/danger-js/issues/812

This does seem to only be a warning though, so I'm not sure if it fixes the issue entirely.


orta:
merge on green

PR #823, opened 2019-02-05
Remove vm2 from Danger

orta:
This removes a dependency that has been used for a while, which moves danger to have one consistent runtime environment. This is feasible because Peril only has one runtime env now since https://github.com/danger/peril/commit/ac2b3a57cb1ea57c1aeaf775108ad5e38a638be2 where I consolidated all runtimes.


source/runner/runners/_tests/vm2.test.ts
```diff
@@ -1,338 +0,0 @@
...
-        expect(results).toEqual({
-          fails: [],
-          markdowns: [{ message: "hello" }, { message: "hello2" }],
-          messages: [],
-          warnings: [],
-        })
-      })
-    })
-  })
-})
```
orta: I can't reliably do the inline code evaluation inside Jest, so instead I powered up the fixtures and moved them to be on a push.

PR #827, opened 2019-02-12
Fix meta on prs

f-meloni:
I've got the changes of the other PR here too by mistake :(, but they are just 2 lines

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 <!-- Your comment below this -->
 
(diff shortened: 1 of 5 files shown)


orta:
Great, so it was the results transformations - interesting

f-meloni:
But i need to re record one of the snapshots that i forgot :D

f-meloni:
Actually was failing because there was a place where i didn't handle the meta correctly :)

orta:
I dunno why it's still yellow, it's green on travis. Merging.

Issue #545, opened 2018-03-27
[BitBucketServer] Missing lines_of_code


zdenektopic:
Hello,

one thing I am missing is `lines_of_code` that was available in Ruby version of Danger. It helps to warn reviewers that PR is bigger than is standard or expected.

orta:
Makes sense, I think it should be pretty easy to add this on to the git DSL. It wasn't added because I think GitHub has it in the PR metadata.

PR #834, opened 2019-02-21
Add linesOfCode in GitDSL

ninjaprox:
Address #545.


orta:
Interesting, cool! 

I think this is is going to need docs stating that this way more of an expensive call than someone would think. For example, you're asking the GitHub API for the diff of every file individually - which could quickly blow through the 5k per hour limit on large PRs

ninjaprox:
Make sense. It actually can be improved with one API call per PR for the diffs of all files. `BitBucketServerGit` also needs update to support this. I'll update the PR soon.

ninjaprox:
I made an update to memoize the diff result so that it won't deplete the 5k per hour limit. `BitBucketServerAPI` seems not to work anymore due to the API v1 deprecation, and not sure if any Bitbucket users out there, so I leave `BitBucketServerGit` as-is. We can update it later along with Bitbucket API v2. In v2, [/2.0/repositories/{username}/{repo_slug}/pullrequests/{pull_request_id}/diff](https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Busername%7D/%7Brepo_slug%7D/pullrequests/%7Bpull_request_id%7D/diff) returns a plain text diff which can help remove `getStructuredDiffForFile` in `GitJSONToGitDSLConfig`, and unify usage of `getFullDiff` for both GitHub and BitBucket.

orta:
Yeah, this makes sense - thanks 👍 cool

PR #835, opened 2019-02-22
Not crash for unknown types

f-meloni:
UNKNOWN is a possible change type for the Bitbucket response (https://docs.atlassian.com/bitbucket-server/javadoc/4.1.0/api/reference/com/atlassian/bitbucket/content/ChangeType.html), then Danger shouldn't crash when the response contains it


orta:
Interesting, cool - makes sense to me 👍

orta:
merge on green

PR #840, opened 2019-03-14
If DangerID is defined, use it as the context for platform status

### LINKED ISSUE #839: Running Danger in two separate CI builds only outputs one status
In order to optimize our PR build speeds, we decided to split our builds in two, one which runs our build and one which runs our tests.

Based on past issues, I saw that it should be possible to run danger in each of these builds and have them report back to the PR. I was able to accomplish this by setting the `dangerId` flag.

This was working great for a few days, but today we discovered that while Danger posts two separate comments, it doesn't post two separate build statuses. If my tests run faster, and they have issues, but my build does not, Danger will briefly reflect the right status (failed), and then when the build finishes, it'll update this to a success.

This is obviously not desirable, as it means Danger is now incorrectly indicating that the build is passing. 

I would like the dangerId to be reflected in the name of the Github statuses, or at the very least enable opting out of Danger posting its own status altogether and just letting it fail the build. We run most things in our build through danger anyway, so having the build pass when Danger finds an issue is sort of weird.

I started poking around at Github's status APIs to see if there was an easy way to do this, but I'm not yet familiar enough with either them or Danger's codebase to figure it out tonight. I'm hoping it's as simple as piping the dangerId through to an API somewhere but it wasn't immediately clear to me how Github determines whether to create a new status or update an existing one.

Can you help me figure out how to enable the functionality we need? Thanks in advance!

randak:
Resolves #839.

Currently, if you run two separate danger builds with their own IDs, they will post two separate comments. However, the status is not tied to the ID, so if one danger build fails and the second one passes, your PR will incorrectly be marked as passing.

This change uses the ID as the `context`/`name` for the status, which will allow each build to show up as an independent status, making it possible to have multiple Danger runs on one PR. (We use this to run our build and test in parallel in the CI and report issues from each of them.)

### DIFF
--- a/source/platforms/BitBucketServer.ts
+++ b/source/platforms/BitBucketServer.ts
@@ -39,18 +39,32 @@ export class BitBucketServer implements Platform {
    * Fails the current build, if status setting succeeds
    * then return true.
    */
-  updateStatus = async (passed: boolean | "pending", message: string, url?: string): Promise<boolean> => {
+  updateStatus = async (
+    passed: boolean | "pending",
+    message: string,
+    url?: string,
+    dangerID?: string
+  ): Promise<boolean> => {
     const pr = await this.api.getPullRequestInfo()
     const { latestCommit } = pr.fromRef
+
     let state = passed ? "SUCCESSFUL" : "FAILED"
     if (passed === "pending") {
       state = "INPROGRESS"
     }
+
+    let name = "Danger"
+    if (process.env["PERIL_INTEGRATION_ID"]) {
+      name = "Peril"
+    } else if (dangerID) {
+      name = dangerID
+    }
+
     try {
       await this.api.postBuildStatus(latestCommit, {
         state: state,
         key: "danger.systems",
-        name: process.env["PERIL_INTEGRATION_ID"] ? "Peril" : "Danger",
+        name: name,
         url: url || "http://danger.systems/js",
         description: message,
       })
--- a/source/platforms/github/GitHubAPI.ts
+++ b/source/platforms/github/GitHubAPI.ts
@@ -344,15 +344,29 @@ export class GitHubAPI {
     return res.ok ? res.json() : { labels: [] }
   }
 
-  updateStatus = async (passed: boolean | "pending", message: string, url?: string): Promise<any> => {
+  updateStatus = async (
+    passed: boolean | "pending",
+    message: string,
+    url?: string,
+    dangerID?: string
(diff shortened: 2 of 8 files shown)


source/platforms/BitBucketServer.ts
```diff
@@ -39,18 +39,32 @@ export class BitBucketServer implements Platform {
...
     const { latestCommit } = pr.fromRef
+
     let state = passed ? "SUCCESSFUL" : "FAILED"
     if (passed === "pending") {
       state = "INPROGRESS"
     }
+
+    let name = "Danger"
+    if (process.env["PERIL_INTEGRATION_ID"]) {
+      name = "Peril"
```
randak: Do we need to support this use case in Peril? I'm not familiar enough with it to know if it's possible to run multiple instances.

randak:
@orta Can you take a look at the test failure here? It looks like it might not be related to my changes, but if it is let me know

source/platforms/BitBucketServer.ts
```diff
@@ -39,18 +39,32 @@ export class BitBucketServer implements Platform {
...
+    } else if (dangerID) {
+      name = dangerID
+    }
+
     try {
       await this.api.postBuildStatus(latestCommit, {
         state: state,
         key: "danger.systems",
-        name: process.env["PERIL_INTEGRATION_ID"] ? "Peril" : "Danger",
+        name: name,
```
fbartho:
```suggestion
        name,
```
I don’t remember the style guide here, but in my repo we omit the key when the value name is the same. Feel free to ignore this!

source/platforms/github/GitHubAPI.ts
```diff
@@ -363,7 +377,7 @@ export class GitHubAPI {
         {},
         {
           state: state,
-          context: process.env["PERIL_BOT_USER_ID"] ? "Peril" : "Danger",
+          context: context,
```
fbartho:
```suggestion
          context,
```

source/platforms/BitBucketServer.ts
orta: Nope,  or at least whatever we need to do it'd be a bit of work on Perils side 👍

randak:
Awesome, thanks @orta! Any estimate of when this will be published? It'll help us plan whether to go forward with a workaround or wait 😄

randak:
Oh, nevermind, Peril answered me, haha 👍

PR #846, opened 2019-03-27
 Send different build update keys based on the id on Bitbucket

f-meloni:
Bitbucket requires different keys in order to show different build reports


orta:
👍 merge on green

Issue #706, opened 2018-10-24
Calling `fail()` does not cause `process.exitCode = 1` without `--verbose`.


dtinth:
When there is a failure, Danger says `Failing the build, there is 1 fail.` but the `danger` process completes successfully.

I looked at the code, and found that `process.exitCode = 1` is in `Executor#handleResultsPostingToSTDOUT`, which is only executed when `--verbose` is set. Therefore, if `--verbose` flag is not set, the build will pass despite Danger said it would fail.

In my opinion, this violates the principle of least surprise in 2 ways: 1) Why does the build not fail despite Danger saying it world, and 2) Why passing `--verbose` affects the exit code of Danger.

orta:
It failed your build via the GitHub status API - Oddly enough, most people felt that doing it this way is the least surprise version of Danger (and what has been the default for about 4 years)  which I think I'm pretty happy with.

In the ruby version someone added a flag for always giving a failing process code on fail, but no-one has ported it over to Danger JS. Happy to have that in 👍

dtinth:
I see… 👍 Thanks for the insights.

For our use cases, we want to have failures actually fail the CI build because we’re using CircleCI Workflows. We use Danger to check if a PR is properly labelled and categorized. That means the problem we try to detect using Danger is not code-related (and can be fixed without committing any code to the repository).

However, CircleCI workflow only allows you to rebuild jobs only when it fails. In this default scenario, Danger failed but exits with code 0. Therefore, the CI job passed. After I fix the problem, I only have the option rerun the entire workflow, causing all tests to be run again.

![image](https://user-images.githubusercontent.com/193136/47441238-e4b8ee00-d7d9-11e8-9ee8-8c7e299f20a2.png)

Here, I added `--verbose` to make Danger also crash the CI job. Now CircleCI offers the option to rerun just the failed jobs. This saves 3 minutes of build time.

![image](https://user-images.githubusercontent.com/193136/47441326-1a5dd700-d7da-11e8-9895-1b6afebe3cde.png)

Since we worked around this issue using the `--verbose` flag, I don’t have an immediate need to fix this. But I agree about creating a dedicated flag that causes `process.exitCode = 1` on failure. Maybe put this issue on [Hacktoberfest](https://hacktoberfest.digitalocean.com/)? 😉

f-meloni:
Fixed by https://github.com/danger/danger-js/pull/847

PR #847, opened 2019-03-31
Add a strict option to fail the build if there are error on the danger report

f-meloni:
Is a big change for javascript skills and experience :D
Hope didn't do anything bad :P

### DIFF
--- a/source/commands/utils/sharedDangerfileArgs.ts
+++ b/source/commands/utils/sharedDangerfileArgs.ts
@@ -24,7 +24,7 @@ export interface SharedCLI extends program.CommanderStatic {
   /** Use a URL to send in the DSL instead of just STDIN */
   passURLForDSL?: boolean
   /** Fail if there are fails in the danger report */
-  strict: boolean
+  failOnErrors: boolean
   /** Use GitHub Checks */
   useGithubChecks: boolean
 }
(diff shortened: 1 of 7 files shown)


source/runner/Executor.ts
```diff
@@ -161,7 +172,6 @@ export class Executor {
         const are = fails.length === 1 ? "is" : "are"
         const message = chalk.underline.red("Failing the build")
         output = `Danger: ${cross} ${message}, there ${are} ${fails.length} fail${s}.`
-        process.exitCode = 1
```
f-meloni: I removed this because now this is more generic and is handled by the `strict` option, if this behaviour has to be kept I can add it on the line 143, by adding `this.options.stdoutOnly || this.options.jsonOnly || (this.ciSource && this.ciSource.useEventDSL` to the `if`.
I think is wrong here in any case because this method is called also by `handleResultsPostingToPlatform` just if `verbose` is true, then the build fails if is verbose as side effect also if you are running `danger ci`

source/commands/utils/sharedDangerfileArgs.ts
```diff
@@ -23,6 +23,8 @@ export interface SharedCLI extends program.CommanderStatic {
   process?: string
   /** Use a URL to send in the DSL instead of just STDIN */
   passURLForDSL?: boolean
+  /** Fail if there are fails in the danger report */
+  strict: boolean
```
orta: Sounds good, but can this please be consistent with the Ruby version and be called `failOnErrors`

https://github.com/danger/danger/blob/13060c145f03fc3f29e2d1f15d156945c5aa7197/lib/danger/commands/runner.rb#L62

fbartho:
Fixes #706

orta:
Cool 👍

zArubaru:
This option (`--strict`) has been replaced by the python equivalent: `--failOnErrors`
https://github.com/danger/danger-js/blob/master/source/commands/utils/sharedDangerfileArgs.ts#L27

Issue #431, opened 2017-12-04
Provide primitives for "if this file exists, then so should this" sort of rules


orta:
Basically, provide some kind of way to say that 

```js
// Check that every file touched has a corresponding test file
const correspondingTestsForModifiedFiles = modifiedFiles.map(f => {
  const newPath = path.dirname(f);
  const name = path.basename(f);
  return `${newPath}/__tests__/${name}`;
});

const testFilesThatDontExist = correspondingTestsForModifiedFiles
  .filter(f => !f.includes('__stories__')) // skip stories
  .filter(f => !fs.existsSync(f));

if (testFilesThatDontExist.length > 0 && !skipTests) {
  const output = `Missing Test Files:
${testFilesThatDontExist.map(f => `- \`${f}\``).join('\n')}
If these files are supposed to not exist, please update your PR body to include "Skip New Tests".`;
  warn(output);
}

// Find changed React components
const changedComponents = modifiedFiles.filter(file => {
  const content = fs.readFileSync(file).toString();
  return content.includes('React');
});

// Check for images in PR description if new components added
if (changedComponents.length > 0 && !skipVisualDiff && !hasScreenShots) {
  const output = `Should there be images to represent these components:
  ${changedComponents.map(f => `- \`${f}\``).join('\n')}
  If these changes are not visual, please update your PR body to include "Skip Visual Diff".`;
  warn(output);
}

// Check that every component touched has a corresponding story
const correspondingStoriesForChangedComponents = uniq(
  changedComponents.map(f => {
    const newPath = path.dirname(f);
    return `${newPath}/__stories__/index.js`;
  })
);

const missingStories = correspondingStoriesForChangedComponents.filter(
  f => !fs.existsSync(f)
);
```

and 

```js
// Custom modifiers for people submitting PRs to be able to say "skip this"
const trivialPR = bodyAndTitle.includes("trivial")
const acceptedNoTests = bodyAndTitle.includes("skip new tests")

const typescriptOnly = (file: string) => includes(file, ".ts")
const filesOnly = (file: string) => fs.existsSync(file) && fs.lstatSync(file).isFile()

// Custom subsets of known files
const modifiedAppFiles = modified.filter(p => includes(p, "lib/")).filter(p => filesOnly(p) && typescriptOnly(p))

// Modified or Created can be treated the same a lot of the time
const touchedFiles = modified.concat(danger.git.created_files).filter(p => filesOnly(p))

[...]

PR #850, opened 2019-04-05
Vendor in Chainsmoker for path matching

paulmelnikow:
I've rewritten [chainsmoker](https://github.com/paulmelnikow/chainsmoker) in TypeScript and vendored it in as we discussed at https://github.com/danger/danger-js/issues/431#issuecomment-480269809.

I took a stab at the integration too, though I'm not sure I've added it in all the right places. Tests are passing locally.

Looking forward to feedback!


package.json
```diff
@@ -93,11 +93,13 @@
     "@types/debug": "0.0.30",
     "@types/get-stdin": "^5.0.1",
     "@types/http-proxy-agent": "^2.0.1",
-    "@types/jest": "^24.0.0",
+    "@types/jest": "^24.0.11",
```
paulmelnikow: This version provides a more type-friendly version of `it.each`.

package.json
```diff
@@ -142,8 +144,10 @@
     "lodash.includes": "^4.3.0",
     "lodash.isobject": "^3.0.2",
     "lodash.keys": "^4.0.8",
+    "lodash.mapvalues": "^4.6.0",
```
paulmelnikow: I had a hard time getting types right for my own `mapValues` implementation so used the one from lodash to save time.

orta:
Alright, so I've given this a run through and given it all some some polish - I love it. 

I opted to take `createdAndModified` and call it `edited`. I think with the docs covering it in auto-complete, the website and the CHANGELOG covering it, it shouldn't be hard to figure out how to use it.

If you're happy, we can merge it 👍

package.json
orta: Yeah, seems reasonable to me 👍

paulmelnikow:
This looks awesome! It feels both elegant and friendly, and I hope people enjoy using it. Thanks for the docs – you described it really nicely. 👍 🚀

orta:
Cool, lets get this shipped then

aaronwhite:
This is great; not sure if this is possible, but would love to get the matching filePaths as part of the structure per type as well

orta:
Maybe they could always have a `paths` attribute with that?

aaronwhite:
So jumping right in w/ this Chainsmoker, I'm greatly simplifying the logic `if(appJSFiles.created) ...` but it's hard to report on *which* without duplicating the `fileMatch` pattern as a bunch of filters, now I have two things to maintain. I'm new to Danger, so I probably don't have all the right terms, but if it's paths (for created, modified, createdOrModified, etc) awesome!

paulmelnikow:
> would love to get the matching filePaths as part of the structure per type as well

That makes a lot of sense.

> Maybe they could always have a `paths` attribute with that?

Yea, I think that would work. Lemme work that up!

paulmelnikow:
Seems like this is working for us: badges/shields#3276 👍

PR #851, opened 2019-04-07
Add `danger.git.fileMatch.getKeyedPaths()`, providing more convenient access to paths

paulmelnikow:
This adds a function that provides an object containing arrays of matched paths. I made this a function instead of a property because `getKeyedPaths()` will stand out more in the dangerfile and be easier to search for. The name will also make clear, that this isn't an array of paths, rather an object containing arrays of paths.

Technically a breaking change, though `tap()` and `debug()` had only existed for a day. 🤷‍♂️ 

Ref https://github.com/danger/danger-js/pull/850#issuecomment-480536193

### DIFF
--- a/source/commands/utils/chainsmoker.ts
+++ b/source/commands/utils/chainsmoker.ts
@@ -7,15 +7,14 @@ import mapValues from "lodash.mapvalues"
 
 export type Pattern = string
 export type Path = string
-export type KeyedPatterns<T> = { [K in Extract<keyof T, string>]: Pattern[] }
-export type KeyedPaths<T> = { [K in Extract<keyof T, string>]: Path[] }
-export type MatchResult<T> = { [K in Extract<keyof T, string>]: boolean }
-
-export interface Chainsmoker<T> {
-  (...patterns: Pattern[]): MatchResult<T>
-  debug(...patterns: Pattern[]): MatchResult<T>
-  tap(callback: (keyedPaths: KeyedPaths<T>) => void): (...patterns: Pattern[]) => MatchResult<T>
+export type KeyedPatterns<T> = { readonly [K in keyof T]: Pattern[] }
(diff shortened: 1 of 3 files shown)


source/commands/utils/chainsmoker.ts
```diff
@@ -7,14 +7,16 @@ import mapValues from "lodash.mapvalues"
 
 export type Pattern = string
 export type Path = string
-export type KeyedPatterns<T> = { [K in Extract<keyof T, string>]: Pattern[] }
-export type KeyedPaths<T> = { [K in Extract<keyof T, string>]: Path[] }
-export type MatchResult<T> = { [K in Extract<keyof T, string>]: boolean }
+export type KeyedPatterns<T> = { readonly [K in keyof T]: Pattern[] }
+export type KeyedPaths<T> = { readonly [K in keyof T]: Path[] }
+export type _MatchResult<T> = { readonly [K in keyof T]: boolean }
```
paulmelnikow: Slight improvement on the types here: keys are readonly; and now supports any kind of key, not only string keys. `Extract<keyof T, string>` is something I carried over from an earlier attempt at typing my own `mapValues` but no longer seems needed for this to compile.

`_MatchResult` is used in the tests.

paulmelnikow:
> Happy either way, but did you consider having:
> 
> ```
> {
>   modified: true,
>   modifiedPaths: ["..."],
>   ...
> }
> ```
> Instead? I dig that this is maybe even impossible to generically type - which is a totally acceptable trade-off IMO. But interested in the idea

I quite like that! I was thinking about trying to do something similar like `getModifiedPaths()`. I looked into it a bit, though there didn't seem to be a way to dynamically type function names like that, and I yea, I prefer the tradeoff to get generic types.

paulmelnikow:
@orta Shall I go ahead and squash-merge this?

orta:
Yep, I'll do it and make a release!

PR #862, opened 2019-04-25
Fix nulls in created_files

keith:
https://github.com/danger/danger-js/issues/807

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 
 <!-- Your comment below this -->
 
+- Fix for nulls in changed_files [@keith][]
+
 # 7.1.3
 
 - Cleans up the declarations a little bit - [@orta]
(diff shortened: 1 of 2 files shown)


keith:
Testing this locally now shows `/dev/null` as the file instead of a null path. But maybe it should actually be showing the new file path?

fbartho:
@keith I think I agree that the canonical path should *always* be the "final path" except for "deleted files".

keith:
Yea you're right, I'm not sure why the `to` is null here then, I wonder if this is an issue on the `parse-diff` side

keith:
The object I get back for these files from parse-diff is:

```
  { chunks: [],
    deletions: 0,
    additions: 0,
    from: '/dev/null',
    to: undefined,
    new: true,
    index: [ '0000000000..b4aca5d807' ] },
```

So that's interesting.

orta:
Does that make it deleted?

PR #863, opened 2019-04-26
Fix CHANGELOG.md documentation about "--failOnErrors"

sogame:
Tiny fix, as currently the changelog is a bit confusing about what's the option to use.


orta:
Thanks!

PR #878, opened 2019-06-14
Fix incorrect Codefresh repo slug for fork builds

stevenp:
The repo slug for Codefresh was being calculated using the `CF_REPO_OWNER` environment variable, but in builds for forks, the owner is the owner of the fork, not the owner of the main repo. As a result, the API calls were pointing to the wrong URL and failing.

This change users the `CF_COMMIT_URL` variable instead, which contains the correct slug information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 
 <!-- Your comment below this -->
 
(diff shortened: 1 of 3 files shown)


orta:
👍 Looks good to me

PR #879, opened 2019-06-21
Add GitLab missing states

f-meloni:
When I was implementing the swift GitLab support I had to look at the docs to get all the possible states, in order to avoid decode failures.
I then thought that would have been nice to update danger-js with the correct states too.


orta:
Ace, thanks!

PR #880, opened 2019-06-22
Fix GitHub Actions warning

f-meloni:


orta:
that might get broke next time someone does prettier, would recommend re-ordering it

PR #881, opened 2019-06-23
Support Bitbucket Cloud and Bitbucket Pipelines

HelloCore:
This PR aims to add support for `Bitbucket Cloud` and `Bitbucket Pipelines`

Current Progress
- [x] Add `Bitbucket Pipelines` to CI
- [x] Add `BitbucketCloud` to `Platform`
- [x] Getting PR's information
- [x] Getting Diff Content
- [x] Generated `JSON DSL` from PR's information
- [x] Add to CLI
- [x] Run on Bitbucket Pipelines
- [x] Create comment template for `Bitbucket Cloud`, currently using the one from `Bitbucket Server`, and it doesn't work. 😢 
- [x] Update / Delete Comment
- [x] Support Inline Comments
- [x] Update Documentation - Requires env `DANGER_BITBUCKETCLOUD_USERNAME`, `DANGER_BITBUCKETCLOUD_PASSWORD`, and `DANGER_BITBUCKETCLOUD_UUID` using `App-Passwords` 
- [x] Update Documentation - Bitbucket Pipelines

Maybe later...
- Remove unused fields [doc](https://developer.atlassian.com/bitbucket/api/2/reference/meta/partial-response)


source/platforms/bitbucket_cloud/_tests_/_bitbucket_cloud_git.test.ts
```diff
@@ -0,0 +1,226 @@
...
+
+    expect(gitDSL.commits[0]).toEqual(exampleCommit)
+  })
+
+  // it("writes a JSON DSL fixture", async () => {
+  //   const fakeSource = new FakeCI({})
+  //   const dataSent = await jsonDSLGenerator(bbc, fakeSource, {} as any)
+  //   dataSent.settings.github.accessToken = "12345"
+
+  //   writeFileSync(pathJoin(fixtures, "bbc-dsl-input.json"), JSON.stringify(dataSent, null, "  "), "utf8")
```
HelloCore: This caused `jest --watch` loop infinitely, and I don't know how to fix it. 😢

orta: No worries, you can add it to the [watch ignore](https://jestjs.io/docs/en/configuration#watchpathignorepatterns-array-string) in Jest

orta:
Wow, so when it rains it pours! We just got GitLab support in and now this 🍡

source/dsl/BitBucketCloudDSL.ts
```diff
@@ -0,0 +1,164 @@
+import { RepoMetaData as BitBucketServerRepoMetaData } from "./BitBucketServerDSL"
+
+export interface BitBucketCloudJSONDSL {
```
f-meloni: I wonder how different are the `BitBucketCloud` and `BitBucketServer` DSLs and if makes sense to have just one for both.
At a first look they seem similar to me.
My opinion about that would be that is better have an easier model for the users of this DSL (both `Danger users and the other Danger version), even if it eventually requires a major version bump.
What do you think?

orta: We ended up having to do this split for Danger ruby too - not sure the reasoning but I must have agreed with it then :D

HelloCore: Yeah they both look similar to me too.
I’ve tried to reuse BitBucketServer’s intefaces, but later I found that they were totally different. 🤯

For example, https://github.com/danger/danger-js/blob/master/source/dsl/BitBucketServerDSL.ts#L91
only ‘message’ and ‘parents’ were the same for both of them.

source/platforms/bitbucket_cloud/_tests_/_bitbucket_cloud_git.test.ts
HelloCore: Thank you! 
I found that it happened because `test-results.json` keeps changing, so I added it to `watchPathIgnorePatterns`.

source/platforms/BitBucketCloud.ts
```diff
@@ -0,0 +1,182 @@
...
+  getPlatformReviewDSLRepresentation = async (): Promise<BitBucketCloudJSONDSL> => {
+    let pr: BitBucketCloudPRDSL
+    try {
+      pr = await this.getReviewInfo()
+    } catch {
+      process.exitCode = 1
+      throw `
+          Could not find pull request information,
+          perhaps Danger does not have permission to access the repo.
+        `
```
orta: yeah, this seems like a reasonable call

source/platforms/_tests/fixtures/bbc-dsl-input.json
```diff
@@ -0,0 +1,2549 @@
...
+        },
+        "message": "bump up version\n",
+        "tree": null,
+        "parents": [
+          "25422a6589feee092d7e47f12c4e7a160d086f2b",
+          "d802defdf0e6b4f85e01e48f2576b5f4ade8c77d"
+        ],
+        "url": "https://bitbucket.org/foo/bar/commits/301b16bbd02e1f2107ce3e50812afd584c5a99e0"
+      }
+    ]
```
orta: thanks for all this, it makes it much easier to figure out as someone who hasn't used bitbucket cloud

source/platforms/pullRequestParser.ts
```diff
@@ -24,6 +25,15 @@ export function pullRequestParser(address: string): PullRequestParts | null {
...
     }
 
+    // shape: https://bitbucket.org/proj/repo/pull-requests/1
+    if (includes(components.path, "pull-requests")) {
+      return {
+        platform: BitBucketCloud.name,
+        repo: components.path.split("/pull-requests")[0].slice(1),
+        pullRequestNumber: components.path.split("/pull-requests/")[1].split("/")[0],
+      }
+    }
```
orta: 👍

source/runner/templates/bitbucketCloudTemplate.ts
```diff
@@ -0,0 +1,76 @@
+import { DangerResults } from "../../dsl/DangerResults"
+import { Violation } from "../../dsl/Violation"
+
+// This unicode emojis also work for old versions of bitbucket server, were emojis are not supported
```
orta:
```suggestion
// This unicode emojis also work for old versions of bitbucket server, where emojis are not supported
```

source/dsl/BitBucketCloudDSL.ts
```diff
@@ -0,0 +1,170 @@
...
+}
+
+export type BitBucketCloudLinks<Names extends string> = {
+  [key in Names]: {
+    href: string
+  }
+}
+
+export interface BitBucketCloudPRParticipant {
+  /*The user for  */
```
orta:
```suggestion
  /*The user who participated in this PR  */
```

orta:
Let me know when you think it's good to go

HelloCore:
It shoud be okay now. 🚀

I will create another PR for adding OAuth support later.

orta:
Alright then, lets do it, another major bump - hah, go marketing numbers. Two in a row!

PR #882, opened 2019-06-26
Fix Bitbucket Cloud multiple messages template

HelloCore:
It couldn't render messages correctly if there were 2 or more messages. 

I totally forgot about this. 🙏


orta:
👍

PR #883, opened 2019-06-26
Add support for BitBucket Cloud OAuth

HelloCore:
Just add support for `BitBucket Cloud OAuth`, not sure if it's more secure than `App passwords` or not. 
Anyway, `Danger Ruby` supports this, so someone might need it later.


orta:
Thanks, sorry it took a while!

Issue #884, opened 2019-06-28
Danger crashes on travis ci with error from gitlab


bjankord:
We recently updated to danger v9.0.1 and it was working just fine. However we recently noticed the following error showing up on travis when we run danger.

```
Log:
Creating network "terracore_default" with the default driver
Creating terracore_standalone-chrome_1 ... 
Creating terracore_standalone-chrome_1
> terra-core@0.1.0 danger /opt/module
> danger ci
/opt/module/node_modules/gitlab/dist/index.js:1
(function (exports, require, module, __filename, __dirname) { "use strict";function _interopDefault(e){return e&&"object"==typeof e&&"default"in e?e.default:e}Object.defineProperty(exports,"__esModule",{value:!0});var universalUrl=require("universal-url"),Li=_interopDefault(require("li")),humps=require("humps"),Ky=_interopDefault(require("ky-universal")),FormData=_interopDefault(require("form-data")),queryString=require("query-string"),randomstring=_interopDefault(require("randomstring")),extendStatics=function(e,t){return(extendStatics=Object.setPrototypeOf||{__proto__:[]}instanceof Array&&function(e,t){e.__proto__=t}||function(e,t){for(var r in t)t.hasOwnProperty(r)&&(e[r]=t[r])})(e,t)};function __extends(e,t){function r(){this.constructor=e}extendStatics(e,t),e.prototype=null===t?Object.create(t):(r.prototype=t.prototype,new r)}var __assign=function(){return(__assign=Object.assign||function(e){for(var t,r=1,n=arguments.length;r<n;r++)for(var o in t=argument
TypeError: global$1.XMLHttpRequest is not a constructor
    at checkTypeSupport (/opt/module/node_modules/gitlab/dist/index.js:1:32017)
    at Object.<anonymous> (/opt/module/node_modules/gitlab/dist/index.js:1:32341)
    at Module._compile (module.js:653:30)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/opt/module/node_modules/danger/distribution/platforms/gitlab/GitLabAPI.js:39:16)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! terra-core@0.1.0 danger: `danger ci`
npm ERR! Exit status 1
[...]

PR #885, opened 2019-06-30
Don't automatically download gitlab minors

f-meloni:
Should fix https://github.com/danger/danger-js/issues/884
Not sure if we want to keep it also after they fix the issue, and update gitlab manually.
That would at least avoid that an issue like the one we have happens again

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
     "commander": "^2.18.0",
     "debug": "^4.1.1",
     "get-stdin": "^6.0.0",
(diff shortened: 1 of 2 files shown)


orta:
Restarted CI without cache - merge on green

orta:
low on time atm due to flappy royale and moving house

orta:
OK, I think this is fine for now

f-meloni:
Thank you @orta :)

PR #887, opened 2019-07-05
Allow handleResults to be called without a git object

jtreanor:
Currently, `handleResults` in `Executor.ts` requires that a non-empty `GitDSL` object is passed so that it can add a signature like `Danger.js against COMMIT_HASH` to comments. This behaviour was introduced in https://github.com/danger/danger-js/pull/783.

This is the cause of a crash in Peril, which breaks a common workflow: https://github.com/danger/peril/issues/436.

The updates I have made here are to make `git` an optional parameter in `handleResults` and the templating functions that it calls. If there is no `git` object the signature just doesn't have the `against COMMIT_HASH` part. I have also added a check so that inline comments won't be added if there is no `git` object since it is required for that.


orta:
Nice, looks 👍 

I'll get this in, shipped and Peril updated

PR #889, opened 2019-07-09
Expose BitBucketServerAPI

### LINKED ISSUE #886: Expose BitBucketServerAPI as `danger.bitbucket_server.api`
Hey,

I'd like to ask if it would be desirable to expose [`BitBucketServerAPI`](https://github.com/danger/danger-js/blob/master/source/platforms/bitbucket_server/BitBucketServerAPI.ts) as  `danger.bitbucket_server.api`? This would be similar how we're currently exposing `OctoKit` as `danger.github.api`:

https://github.com/danger/danger-js/blob/9ba46ffc42463e2ca0d6df9ef93fb4118ec7f973/source/runner/jsonToDSL.ts#L33-L34

A very simple way to achieve this would be to change that second line to

```js
bitbucket_server = dsl.bitbucket_server && { ...dsl.bitbucket_server, api }; // might need a type cast here or something.. I don't really use TypeScript lol

// or alternatively just mutate bitbucket_server, maybe TypeScript won't complain?
bitbucket_server = dsl.bitbucket_server
if (bitbucket_server) {
  Object.assign(bitbucket_server, { api })
}
```

Implementation details aside, does this sound like a reasonable change? This would enable Bitbucket Server users to do things like `danger.bitbucket_server.api.getPullRequestInfo()` or `.get()`, `.post()` etc.

NMinhNguyen:
Closes #886

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 
 <!-- Your comment below this -->
 
+- Expose BitBucketServerAPI - [@NMinhNguyen]
+
 # 9.0.3
 
 - Add support OAuth for BitBucket Cloud - [@HelloCore]
--- a/source/ci_source/_tests/_get_ci_source.test.ts
+++ b/source/ci_source/_tests/_get_ci_source.test.ts
@@ -14,7 +14,7 @@ describe(".getCISourceForEnv", () => {
   })
 })
 
-describe(".getCISourceForExternal", async () => {
+describe(".getCISourceForExternal", () => {
   test("should resolve module relatively", async () => {
     const ci = await getCISourceForExternal({}, "./source/ci_source/_tests/fixtures/dummy_ci.js")
     expect(ci).toBeInstanceOf(DummyCI)
--- a/source/dsl/BitBucketServerDSL.ts
+++ b/source/dsl/BitBucketServerDSL.ts
@@ -1,3 +1,5 @@
+import { BitBucketServerAPI } from "../platforms/bitbucket_server/BitBucketServerAPI"
+
 /** Key details about a repo */
 export interface RepoMetaData {
   /** A path like "artsy/eigen" */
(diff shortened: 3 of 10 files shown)


orta:
Cool, yep - looks good

milesj:
This is broken on master when using TS.

```
../../node_modules/danger/distribution/danger.d.ts:209:8 - error TS2304: Cannot find name 'BitBucketServerAPI'.

209   api: BitBucketServerAPI
           ~~~~~~~~~~~~~~~~~~
```

NMinhNguyen:
Thanks for the bug report @milesj! May I suggest you use https://github.com/ds300/patch-package for the time being and remove `api: BitBucketServerAPI` from `danger.d.ts`?

@orta could you point me in the right direction as to how to make `danger.d.ts` import `BitBucketServerApi` when it generates the `.d.ts` file? I believe it's being excluded due to this line https://github.com/danger/danger-js/blob/33c3674c8a0f4d0fffc7a58b958ad5a3f7fcf239/scripts/danger-dts.ts#L84

but I don't quite understand it - removing the `import` guard seems to break quite a lot of other stuff too.

orta:
Step one is I should have added a check that it compiles, which I've now added:

```sh
orta@Perl ~/d/d/danger-js> yarn declarations
yarn run v1.16.0
$ ts-node ./scripts/create-danger-dts.ts
Found an error in the generated DTS file, you probably need to edit scripts/danger-dts.ts
source/danger.d.ts:209:8 - Cannot find name 'BitBucketServerAPI'.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
orta@Perl ~/d/d/danger-js> 
```

Next, the declarations are somewhat hand-built, anything public facing has an interface in the DSL folder, and privately your implementations work to that interface. As this was new code it didn't work that way, I'll switch it up 👍

PR #891, opened 2019-07-15
Add branch reference to Github Workflow docs

ravanscafi:
Otherwise, it doesn't work.
image

See:
https://developer.github.com/actions/managing-workflows/workflow-configuration-options/#using-a-dockerfile-image-in-an-action


orta:
I'm somewhat worried by this, I'll take it (because broken is broken) but maybe it makes sense to add a new branch which gets pushed to on every deploy so that there's a `stable` branch that it can tie to.

PR #892, opened 2019-07-16
Add icon option to message method for table

### LINKED ISSUE #890: Add option for custom icon/emoji for messages
Danger should be able to add custom emoji/icons to the message table. Something like:

image

I think the function could be modified to accept an options object, and add existing opts there?:
```ts
message(message: MarkdownString, opts?: {file?: string, line?: number, emoji?: MarkdownString}): void
```

Willing to work on this, if someone can give me a few pointers :smiley:

rohit-smpx:
Closes #890 

Was not sure how to add icon to `bitbucketServerTemplate` as the icon there is in the heading of the table.

Also `emoji` is current terminology in code everywhere, should I rename `icon` to `emoji` too? I named it `icon` because as was mentioned in https://github.com/danger/danger-js/issues/890#issuecomment-511213623 it could also not be an emoji, could be an image.

Added Tests


orta:
Yep, looks good - get a test or two in to make sure it stays that way and that's a solid contribution. Thanks!

rohit-smpx:
> Yep, looks good - get a test or two in to make sure it stays that way and that's a solid contribution. Thanks!

Will try to finish this today

rohit-smpx:
The `writeResults` option in `run-fixtures` script was not working correctly so fixed that also and made a `yarn test:update-fixtures` option. I hope the name's alright.

source/runner/Dangerfile.ts
```diff
@@ -54,6 +54,18 @@ export interface DangerContext {
...
+   * and warn is the default emoji which shows in the table.
+   * You can also specifiy a custom emoji to show in the table for each message
+   * 
+   * @param {MarkdownString} message the String to output
+   * @param {{file?: string, line?: string, icon?: MarkdownString}} [opts]
+   * @param opts.file a file which this message should be attached to
+   * @param opts.line the line which this message should be attached to
+   * @param opts.icon icon string or image to show in table, take care not to break table formatting
+   */
+  message(message: MarkdownString, opts?: { file?: string, line?: number, icon?: MarkdownString }): void
```
orta: Interesting, yeah, I can see this as a reasonable approach, it is getting pretty unwieldy at that point 👍

package.json
```diff
@@ -54,6 +54,7 @@
     "release": "release-it",
     "test": "jest",
     "test:fixtures": "node ./scripts/run-fixtures.js",
+    "test:update-fixtures": "yarn test:fixtures --update",
```
orta: Yeah, this works for me - nice one

PR #893, opened 2019-07-17
Add Compliment Message to Bitbucket Cloud Comment

HelloCore:
Provide compliment message for a better feeling.

Move `compliment` to `DangerUtils.ts`, so we can reuse it in other places.


orta:
Is it possible that this means Danger won't delete the message when it is green?

HelloCore:
> Is it possible that this means Danger won't delete the message when it is green?

No, this should be the case when there is only message or markdown.
I think it will be useful because Bitbucket Cloud doesn’t show build status message on main screen, I have to click on little green check mark to see the message.

orta:
👍 great- thanks!

PR #894, opened 2019-07-17
Cache Bitbucket Cloud Commits

HelloCore:
We fetch commits twice, so caching it might be better.


orta:
👍  Looks good - merge on green

orta:
I'll give #892 a chance to gett merged and ship all three

PR #895, opened 2019-07-18
Working with two-way authentication in custom Gitlab server

skyway777:
Before this PR, DangerJS was supporting only NODE_EXTRA_CA_CERTS env-variable to set CA certificates.

In my case, I need to use also my SSL Key and my SSL Certificate to connect to Gitlab Api.

When reading node-gitlab docs https://github.com/jdalrymple/node-gitlab I've found only way to do this as using custom requester (requester prop in options)

So, I've made my own KyRequestor realization with supporting of GITLAB_SSL_KEY and GITLAB_SSL_CERT env-variables.


skyway777:
Maybe there is better way to do it?

yarn.lock
```diff
@@ -4067,6 +4093,20 @@ gitconfiglocal@^1.0.0:
...
+  integrity sha512-I8teXj+JzYokyPsoMWRbDKqesNafwRlOqrq7tuNIaeotvsVr+jb9WzEMen+Zy/dHLI6aHN/fp4N5pe3plrdJjQ==
+  dependencies:
+    form-data "^2.3.3"
+    humps "^2.0.1"
+    ky "^0.11.0"
+    ky-universal "^0.2.1"
+    li "^1.3.0"
+    query-string "^6.8.1"
+    randomstring "^1.1.5"
+    universal-url "^2.0.0"
```
orta: I'm worried that we have both gitlab 10, as we have gitlab locked to v6  in the app - have you tried updating gitlab to v10 and seeing if that can fix it in general?

orta: It might be worth giving this another shot now that #903 is in

ludakhris:
I'm curious whats up with this PR, I'm having a similar issue it seems like with a gitlab self hosted instance that has two way auth.

skyway777:
@ludakhris I cannot check if it does work now, since I don't have access to the GitLab instance for which I've created this PR anymore. But if you need this, you can build my fork of DangerJS. It depends on my custom [custom-node-gitlab-requester](https://github.com/skyway777/custom-node-gitlab-requester), which also required building.

Jimimaku:
> Vor dieser PR unterstützte DangerJS nur NODE_EXTRA_CA_CERTS env-Variable zum Setzen von CA-Zertifikaten.
> 
> In meinem Fall muss ich auch meinen SSL-Schlüssel und mein SSL-Zertifikat verwenden, um eine Verbindung zur Gitlab-API herzustellen.
> 
> Beim Lesen von node-gitlab-Dokumentenhttps://github.com/jdalrymple/node-gitlabhabe ich nur einen Weg gefunden, dies als benutzerdefinierten Anforderer zu tun (Requester prop in Optionen)
> 
> Also habe ich meine eigene KyRequestor-Realisierung mit Unterstützung von GITLAB_SSL_KEY und GITLAB_SSL_CERT env-Variablen gemacht.

yarn.lock
Jimimaku: > At first, I've tried to use just node script without dangerjs with last `gitlab` as dependency. This problem was reproduced.
> 
> Gitlab have an option to fix it - not very handy but working - send custom `requester` as parameter. I made the fix with extra options actually, this function in my code fixes my problem:
> 
> ```
> function buildAgent() {
>     const { GITLAB_SSL_KEY, GITLAB_SSL_CERT } = process.env;
> 
>     if (GITLAB_SSL_KEY || GITLAB_SSL_CERT) {
>         const key = readKey(GITLAB_SSL_KEY);
>         const cert = readKey(GITLAB_SSL_CERT);
> 
>         return new https.Agent({
>             key,
>             cert,
>         })
>     }
>     return;
> }
> ```
> 
> If it would be better for this PR, I can downgrade some dependencies in my project.
> 
> Or, maybe, would be better to extend Dangerjs API to allow sending extra platform-specific options (github, bitbucket etc.) in `dangerfile` script? Like `requester` in gitlab https://github.com/jdalrymple/node-gitlab It would allows me do not fork danger-js only to use it in my case.

Jimimaku:
> Warnings
> ⚠️	Could not get info from npm on [`custom-node-gitlab-requester`](https://www.npmjs.com/package/custom-node-gitlab-requester)
> New dependencies added: [`custom-node-gitlab-requester`](https://www.npmjs.com/package/custom-node-gitlab-requester).
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [5a1a73d](https://github.com/danger/danger-js/commit/5a1a73ddc3ae282834e7c5476836563b4557882e)

Jimimaku:
> Warnings
> ⚠️	Could not get info from npm on [`custom-node-gitlab-requester`](https://www.npmjs.com/package/custom-node-gitlab-requester)
> New dependencies added: [`custom-node-gitlab-requester`](https://www.npmjs.com/package/custom-node-gitlab-requester).
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [5a1a73d](https://github.com/danger/danger-js/commit/5a1a73ddc3ae282834e7c5476836563b4557882e)

fbartho:
@Jimimaku you're duplicating other people's comments. Please stop this.

Jimimaku:
> Before this PR, DangerJS was supporting only NODE_EXTRA_CA_CERTS env-variable to set CA certificates.
> 
> In my case, I need to use also my SSL Key and my SSL Certificate to connect to Gitlab Api.
> 
> When reading node-gitlab docs https://github.com/jdalrymple/node-gitlab I've found only way to do this as using custom requester (requester prop in options)
> 
> So, I've made my own KyRequestor realization with supporting of GITLAB_SSL_KEY and GITLAB_SSL_CERT env-variables.

glensc:
> @Jimimaku you're duplicating other people's comments. Please stop this.

We believe [Jimimaku](https://github.com/Jimimaku) was affected by some malware. his comments should be marked as hidden.

PR #896, opened 2019-07-20
Adds some improvements to knowing if you've accidentally broken the d eclarations file generation

orta:
Figured changing the public API wasn't well documented enough, so I gave that a run-through and fixed up the DTS for TS users.

/cc @NMinhNguyen

### DIFF
--- a/scripts/create-danger-dts.ts
+++ b/scripts/create-danger-dts.ts
@@ -13,4 +13,26 @@ fs.writeFileSync("source/danger.d.ts", output)
 fs.writeFileSync("distribution/danger.d.ts", output)
 fs.writeFileSync("types/index.d.ts", "// TypeScript Version: 2.2\n" + output)
 
-console.log("Awesome - shipped to source/danger.d.ts, distribution/danger.d.ts and types/index.d.ts")
+import * as ts from "typescript"
+
+const program = ts.createProgram(["source/danger.d.ts"], { noEmit: true })
+const emitResult = program.emit()
+const allDiagnostics = ts.getPreEmitDiagnostics(program).concat(emitResult.diagnostics)
+if (allDiagnostics.length) {
+  process.exitCode = 1
+  console.log("Found an error in the generated DTS file, you probably need to edit scripts/danger-dts.ts")
+  console.log(
+    "\nIf you've added something new to the DSL, and the errors are about something missing, you may need to add an interface in `source/dsl/*`."
+  )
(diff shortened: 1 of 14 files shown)


scripts/create-danger-dts.ts
```diff
@@ -13,4 +13,26 @@ fs.writeFileSync("source/danger.d.ts", output)
...
+      let { line, character } = diagnostic.file.getLineAndCharacterOfPosition(diagnostic.start!)
+      let message = ts.flattenDiagnosticMessageText(diagnostic.messageText, "\n")
+      console.error(`${diagnostic.file.fileName}:${line + 1}:${character + 1} - ${message}`)
+    } else {
+      console.error(`${ts.flattenDiagnosticMessageText(diagnostic.messageText, "\n")}`)
+    }
+  })
+} else {
+  console.log("Awesome - shipped to source/danger.d.ts, distribution/danger.d.ts and types/index.d.ts")
+}
```
orta: Verifies the DTS compiles using the TS api

scripts/danger-dts.ts
```diff
@@ -83,6 +83,9 @@ import * as GitHub from "@octokit/rest"
     .filter(line => {
       return !line.startsWith("import") && !line.includes("* @type ")
     })
+    .filter(line => {
+      return !line.includes("Please don't have")
+    })
```
orta: This sentence used to be in a few DSL files, now it's in all of them

source/danger.d.ts
```diff
@@ -200,10 +198,32 @@ interface BitBucketServerJSONDSL {
...
+  /** Make a post call against the bitbucket server API */
+  post(path: string, headers: any, body: any, suppressErrors?: boolean): Promise<any>
+
+  /** Make a put call against the bitbucket server API */
+  put(path: string, headers: any, body: any): Promise<any>
+
+  /** Make a delete call against the bitbucket server API */
+  delete(path: string, headers: any, body: any): Promise<any>
+}
+
```
orta: @NMinhNguyen - this is currently the only "exposed" API in the sense of auto-complete, the entire class instance is there with its real extra functions, you can expand this if there were specific functions you needed

NMinhNguyen:
@orta Thanks so much for this. I tried to resolve conflicts in `CHANGELOG.md` directly on GitHub which created a merge commit which failed in CI. I then thought I would try `git rebase master` but that still failed. It's not obvious to me why Node 8 builds are timing out 😕

PR #898, opened 2019-07-23
Retry API requests in fetch.ts

jtreanor:
This adds retry handling to the api wrapper in `fetch.ts`. It means that 5xx errors and 401 errors will be retried with backoff up to 3 times. As well as providing some resilience to API issues, once integrated into Peril it will allow https://github.com/danger/peril/issues/440 to be fixed.

@orta Would you prefer the retry logic to live in `GithubAPI.ts` so that it only applies to Github?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 
 <!-- Your comment below this -->
 
+- Add retry handling for API requests - [@jtreanor]
+
 # 9.1.1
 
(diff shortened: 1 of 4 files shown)


orta:
Nah, this feels totally fine - I'm sure all of the providers could do with it

jtreanor:
Thanks! If you could release a new Danger version whenever you have a chance, I can handle the update on the Peril side.

orta:
Shipped

PR #899, opened 2019-07-23
Unify some common logic of bitbucket

Soyn:
The code of bitbucket server and bitbucket cloud has the duplicated portion. It's better to unify them.


source/runner/templates/_tests/__snapshots__/_bitbucketCloudTemplate.test.ts.snap
```diff
@@ -31,7 +31,7 @@ exports[`generating messages for BitBucket cloud Mixed icon messages match snaps
   
   
 
-  Generated by 🚫 [dangerJS](https://danger.systems/js) against e70f3d6468f61a4bef68c9e6eaba9166b096e23c
+  Generated by ὪB [dangerJS](https://danger.systems/js) against e70f3d6468f61a4bef68c9e6eaba9166b096e23c
```
f-meloni: Are you sure this is going to be ok in both bitbucket cloud and bitbucket server, including the old bitbucket server versions that a lot of companies are still using?

source/runner/templates/bitbucketServerTemplate.ts
```diff
@@ -76,6 +65,8 @@ export const messageForResultWithIssues = `${warningEmoji} Danger found some iss
  */
 export function template(dangerID: string, results: DangerResults, commitID?: string): string {
   return `
+${messageForResultWithIssues}
```
f-meloni: Wouldn't this be added to the danger comment? If yes, why?

Soyn: In bitbucket cloud implementation --->[Bitbucket Cloud Template](https://github.com/danger/danger-js/blob/4264b34c50f4e386524b441f42b64c7376de4c43/source/runner/templates/bitbucketCloudTemplate.ts#L56). 

```javascript
export function template(dangerID: string, results: DangerResults, commitID?: string): string {
  let summaryMessage: string
  if (!results.fails.length && !results.warnings.length) {
    summaryMessage = `${successEmoji}  All green. ${compliment()}`
  } else {
    summaryMessage = messageForResultWithIssues
  }
  return `
  ${summaryMessage}
   .....
```

The summary message will be inserted to every template. Maybe bitbucket server also need this? If so, i will extract this message hint to a independent function.

f-meloni: I feel we shodn't do that. 
Bitbucket server is used in a lot of repos for years now with this template, and I personally don't see any benefit that this message is giving to the template that could justify change it.
For example I use the bitbucket template, and I have my dangerfile divided in tasks and every taks makes a report... I've sent some times to try to not make the danger comment a little bit smaller, because as I said I had a lot of reports and the comment could become really big given there was also the coverage of the modified files... This would make my comment bigger, but the information I'm getting in exchange would not be really useful IMO

f-meloni: Bitbucket cloud is a new implementation, and I remember there was a reason for that message, and didn't modify any existing template, then I suppose in that case is different and totally ok

Soyn: Okay! I will revert the bitbucket server template change. By the way,  is there a guide or a effective way for developer to debug locally?

source/runner/templates/_tests/__snapshots__/_bitbucketCloudTemplate.test.ts.snap
Soyn: I do not have a bitbucket server env yet. But in current change. I replace the bitbucket cloud emoji with unicode emoji, which is supported by bitbucket server. I think it will be worked, may be not. I test the unicode in my bitbucket cloud account. The emoji can not be rendered. I will change this later.

Soyn:
@f-meloni I worked on this pull request. But the emoji encoding difference in server and cloud is an issue. Some duplicated code caused by the emoji encoding. Do you have some suggestions?

orta:
It might not be worth unifying them? There's simplicity in each one having a single responsibility

Soyn:
@orta I will close this pull request later. But i do not think write the same code twice is a good practice. Maybe we needs a utils function to do some common work which is not associated with platform for better maintainability. Not copy and paste.

PR #903, opened 2019-07-30
fix: #875 Update node-gitlab to 10.0.1 with the Ky exit fix

### LINKED ISSUE #875: GitLab process does not exit
When using the GitLabAPI a [bug in `ky`](https://github.com/sindresorhus/ky/issues/138) prevents danger from exiting. Danger does not call `process.exit` but instead sets `process.exitCode`.

We can add [`process.exit`](https://nodejs.org/api/process.html#process_process_exit_code) in the commands to force an exit after `runRunner` has completed. This will still use the same exitCode and will prevent any future issues caused by leftover timeouts.

awgeorge:
Closes #875


orta:
Perfect, cool - yep, looks good to me 👍

PR #904, opened 2019-07-31
Fix the break link and make some useful links displyed.

Soyn:
When i read the source code of project. I find sone useful link for developer is not displayed. I think it will be helpful for more developers if it's displayed.


orta:
This breaks the links above (check out the markdown preview) but I'm open to having that section at the bottom - yep!

Soyn:
> This breaks the links above (check out the markdown preview) but I'm open to having that section at the bottom - yep!

First time to know the md syntax like that. Hahahah, i have fixed the break link.

orta:
👍 merge on green

PR #905, opened 2019-07-31
 Take commit hash from CI Source if available

f-meloni:
Danger doesn't currently have a way to understand on which commit is running.
To get the current commit it asks to the platform the list of commits on the PR and assumes that is running on the last one.
This creates few problems when you push a new commit before Danger runs on the CI for the previous commit, because it will assume is running on the last one, but that is actually not true, and that means that both the status update and the commit hash in the Danger comment signature will be wrong.
My proposal to solve the problem is to ask the CI the commit hash, and if is available use that one, otherwise fallback on the old behaviour.

### DIFF
--- a/source/ci_source/ci_source.ts
+++ b/source/ci_source/ci_source.ts
@@ -18,6 +18,8 @@ export interface CISource {
   /** What unique id can be found for the code review platform's PR */
   readonly pullRequestID: string
 
+  readonly commitHash?: string
+
   /** allows the source to do some setup */
   setup?(): Promise<any>
 
--- a/source/ci_source/providers/Jenkins.ts
+++ b/source/ci_source/providers/Jenkins.ts
@@ -75,4 +75,8 @@ export class Jenkins implements CISource {
   get ciRunURL() {
     return process.env.RUN_DISPLAY_URL || process.env.BUILD_URL
   }
+
+  get commitHash(): string {
+    return this.env.GIT_COMMIT
+  }
 }
--- a/source/ci_source/providers/Travis.ts
+++ b/source/ci_source/providers/Travis.ts
@@ -83,6 +83,10 @@ export class Travis implements CISource {
     // TODO: This only supports travis.org
     return `https://travis-ci.org/${this.env.TRAVIS_REPO_SLUG}/jobs/${this.env.TRAVIS_JOB_ID}`
   }
+
+  get commitHash(): string {
+    return this.env.TRAVIS_COMMIT
+  }
 }
 
 // See end of https://travis-ci.org/danger/danger-js/jobs/317790046
--- a/source/ci_source/providers/Travis.ts
+++ b/source/ci_source/providers/Travis.ts
@@ -83,10 +83,6 @@ export class Travis implements CISource {
     // TODO: This only supports travis.org
     return `https://travis-ci.org/${this.env.TRAVIS_REPO_SLUG}/jobs/${this.env.TRAVIS_JOB_ID}`
   }
-
-  get commitHash(): string {
-    return this.env.TRAVIS_COMMIT
-  }
 }
 
 // See end of https://travis-ci.org/danger/danger-js/jobs/317790046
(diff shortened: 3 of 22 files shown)


source/ci_source/ci_source.ts
```diff
@@ -18,6 +18,8 @@ export interface CISource {
   /** What unique id can be found for the code review platform's PR */
   readonly pullRequestID: string
 
+  readonly commitHash?: string
```
f-meloni: Not all the CI implement this at the moment, each CI will have to implement its own way of get this, for now I implemented only Jenkins and Travis

source/platforms/_tests/_bitbucket_cloud.test.ts
```diff
@@ -75,4 +75,30 @@ describe("getPlatformReviewDSLRepresentation", () => {
     const dsl = await bbs.getPlatformReviewDSLRepresentation()
     expect(dsl.metadata).toMatchSnapshot()
   })
+
+  it("should update status", async () => {
+    api.postBuildStatus = jest.fn().mockReturnValue({ ok: true })
+    await bbs.updateStatus(true, "message", undefined, "danger", undefined)
+
+    expect(api.postBuildStatus).toHaveBeenCalledWith("007bf2423436", {
```
f-meloni: Given this test was missing in both the platform I added it also if doesn't directly test the change in this PR

source/ci_source/providers/Jenkins.ts
```diff
@@ -75,4 +75,8 @@ export class Jenkins implements CISource {
   get ciRunURL() {
     return process.env.RUN_DISPLAY_URL || process.env.BUILD_URL
   }
+
+  get commitHash(): string {
+    return this.env.GIT_COMMIT
```
f-meloni: `GIT_COMMIT - SHA of the current`
From https://wiki.jenkins.io/display/JENKINS/Git+Plugin

source/ci_source/providers/Travis.ts
```diff
@@ -83,6 +83,10 @@ export class Travis implements CISource {
     // TODO: This only supports travis.org
     return `https://travis-ci.org/${this.env.TRAVIS_REPO_SLUG}/jobs/${this.env.TRAVIS_JOB_ID}`
   }
+
+  get commitHash(): string {
+    return this.env.TRAVIS_COMMIT
```
f-meloni: `TRAVIS_COMMIT: The commit that the current build is testing.`
From: https://docs.travis-ci.com/user/environment-variables/

orta:
Yeah, seems reasonable to me

f-meloni:
Ok it looks working, but there is a problem... Travis doesn't link the correct commit, but a merge commit, like 2d15243b76a1f94891e393130ef7bc5a37ece3bd
Then I will remove the Travis commit hash at the moment and will take more time to have a look at it

source/ci_source/ci_source.ts
Soyn: How about other providers?

f-meloni: As I said in the PR description:

> My proposal to solve the problem is to ask the CI the commit hash, and if is available use that one, otherwise fallback on the old behaviour.

Soyn: I mean it is possible to implement the `commitHash` to some providers which is available.
I know it has a lot providers to inspect. But we need to ensure github, gitlab, bitbucket server and bitbucket cloud work well at least.  According to your description, it could cause a bug. Maybe we can fix some providers we support in `danger ci`. What do you think? @orta

f-meloni: What bug could it cause for the sources where is not implemented? 
If is not available it will work as before, then at the moment the first one to get this is Jenkins, and only if has the git plugin.
I will personally try it there, and then implement it in other providers with other PRs. (I already have Bitrise and Travis ready)

Soyn: >This creates few problems when you push a new commit before Danger runs on the CI for the previous commit, because it will assume is running on the last one, but that is actually not true, and that means that both the status update and the commit hash in the Danger comment signature will be wrong.
I mean some bugs already exist in other providers. So if it is possible. We could fix it now. 
If not, we could fix it later when bug is reported on other providers.

f-meloni: That is how danger currently begin, we can update the other platforms, but we don't need to make all the platforms in one PR.
After we merge this i will slowly add more platforms for this, but everyone can add more platform, is actually a good "first issue" that someone can take to start contributing to the project

orta:
Yeah, agree, this feels fine to build out incrementally @f-meloni  - might be worth making an issue for someone to make more in the future

PR #906, opened 2019-08-01
Use new env variable to replace old

Soyn:
Bitbucket cloud already has the env variable we want. I have test this by my account.
More info in https://confluence.atlassian.com/bitbucket/variables-in-pipelines-794502608.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 
 <!-- Your comment below this -->
 
+- Use new env `BITBUCKET_REPO_FULL_NAME` in bitbucket pipeline. - [@Soyn]
+
 # 9.1.3
 
 - Updates GitLab API to 10.x - [@awgeorge]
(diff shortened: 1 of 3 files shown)


orta:
Looks good - Can you make it support both and add a changelog please? I'm wary of people using an older build of bitbucket getting fails

Soyn:
Updated!

orta:
Thanks @Soyn  - can you also make it still support the old style, and use the new version if it's there so that we keep backwards compatibility too please

Soyn:
It's a necessary work to support old style? The new bb pipeline env should work if you use pipeline on bb cloud.

orta:
I don't know because I don't use it - so I'd rather err on the side of caution, especially with enterprise-y stuff where anyone could have old versions

Soyn:
I use the bb cloud in my work. The new env variable works well. Maybe we do not need to keep the old style.

HelloCore:
It works on my project too.

orta:
I prefer that we do, I'd rather never break backwards compatibility when it's cheap - it's a single ternary.

Soyn:
Okay! I will update the code to keep the backwards compatibility.

orta:
Thanks! Merge on green

PR #909, opened 2019-08-11
Add type notation for better readable.

Soyn:


orta:
Thanks for the PR but I'm fine with the original inference for the return value of that function, and Peril (or other tools using Danger JS) might be using that export so we can't just remove them ad-hoc without good reason.

PR #913, opened 2019-08-25
remove 'Dangerfile.ts' and 'Dangerfile.js'

Soyn:
I do not think the dangerfile named with uppercase is used by user. Cause the error message we hint is all by lower case if user passed with wrong options.

### DIFF
--- a/source/commands/utils/fileUtils.ts
+++ b/source/commands/utils/fileUtils.ts
@@ -19,13 +19,5 @@ export function dangerfilePath(program: any): string {
     return "dangerfile.js"
   }
 
-  if (existsSync("Dangerfile.ts")) {
-    return "Dangerfile.ts"
-  }
-
-  if (existsSync("Dangerfile.js")) {
-    return "Dangerfile.js"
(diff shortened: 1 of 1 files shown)


f-meloni:
I personally don't see a benefit here that could justify a breaking change 🤔
The message suggests to use a lower case `dangerfile`, but ,given we are currently allowing a capitalised `Dangerfile` as well, maybe someone is using it, then why force them to change it?

Soyn:
The capitalized `Dangerfile` is visible in code. Not in doc. Maybe someone uses it. But it's not a common usage. IMO.

f-meloni:
We could change the doc if needed, but if someone is already using it my personal opinion is to not remove the support without a good reason, this change would technically require a major update, and I don't think is worth a major update.

Soyn:
User who use the capitalized `Dangerfile` is not a recommend usage if it's not in doc. So i think it's not our duty to handle this unusual usage. Or we can update the doc to support. But i do not like this way.

PR #917, opened 2019-09-10
Fix Buddy.works isPR because PR ids are not always ints

mmiszy:
I was testing Buddy.works CI and it seems the integration with Danger doesn't work. I figured it's because of an incorrect assumption that PR ids are always integers. In my case, in a public GitHub repository, PR id is eg. `pull/42`.

I tested this and it works fine now 👍 

Pinging @kristof0425 because he implemented the integration.

### DIFF
--- a/source/ci_source/providers/_tests/_buddyWorks.test.ts
+++ b/source/ci_source/providers/_tests/_buddyWorks.test.ts
@@ -3,7 +3,7 @@ import { getCISourceForEnv } from "../../get_ci_source"
 
 const correctEnv = {
   BUDDY_PIPELINE_ID: "170873",
-  BUDDY_EXECUTION_PULL_REQUEST_ID: "1799",
+  BUDDY_EXECUTION_PULL_REQUEST_ID: "pull/1799",
   BUDDY_REPO_SLUG: "danger/dangerjs",
   BUDDY_EXECUTION_URL:
     "https://app.buddy.works/danger/dangerjs/pipelines/pipeline/170873/execution/5d6d49dbaab2cb6fdf975c71",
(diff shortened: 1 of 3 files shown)


source/ci_source/providers/_tests/_buddyWorks.test.ts
```diff
@@ -61,7 +61,7 @@ describe(".isPR", () => {
 describe(".pullRequestID", () => {
   it("pulls it out of the env", () => {
     const buddyWorks = new BuddyWorks(correctEnv)
-    expect(buddyWorks.pullRequestID).toEqual("1799")
+    expect(buddyWorks.pullRequestID).toEqual("pull/1799")
```
orta: This will need you to return the number "42" - so I'd check maybe for "pull" in the string and extract the number out in the BuddyWorks class 👍

mmiszy: I wasn't aware of that, it seemed to be working just fine. I'm almost certain there's another env variable which holds just the number. I'll update the PR in a few hours @orta
Thanks for review!

mmiszy: Updated now @orta

mmiszy: BTW I was wondering if it made sense to change the return type of `get pullRequestID` to `number`? :) This way it would be clear just by looking at the definition that a number is required.

orta: Maybe it was being set from somewhere else. The GitHub API takes strings for the pull ID - likely legacy on their side, but I'm open to making them numbest so long as it's parsed back into a string in the API layers

orta:
Cool, thanks!

Issue #914, opened 2019-08-28
Remove jsome dependency


mathiashoeld:
The npm package `mem` is marked as security vulnerability by Github. The package is a dependency of `os-locale` which is a dependency of `yargs` which is a dependency of `jsome` which has not been updated for 2 years, i.e. it is not likely that the `yargs` dependency will be updated anytime soon. (See [this issue for reference](https://github.com/Javascipt/Jsome/issues/20))

To remove this security vulnerability it would be ideal to remove the dependency on jsome since it is probably only used only for styling the json output.

orta:
https://github.com/rafeca/prettyjson could do it

platan:
I think this issue was resolved by https://github.com/danger/danger-js/pull/920. Can we close it?

orta:
Ace, sure

PR #920, opened 2019-09-24
(danger/danger-js#914) replace jsome

watchinharrison:
## Description

replace `jsome` with `prettyjson`
removes security issue with sub deps of `jsome`

## Motivation
https://github.com/danger/danger-js/issues/914


orta:
Perfect! Less is less ❤️

orta:
merge on green

PR #924, opened 2019-09-26
Fix Bitbucket Server binary files

osmestad:
Thanks for getting my previous fix in! When testing `9.2.1` I noticed PRs with binary files failing, this should fix that:

When getting diffs for binary files with BitBucket Server, Danger fails when trying to map over chunks.
This fix avoids setting the `chunks` property in `structuredDiffForFile` if it is `undefined`.

Not sure if this is the correct location to fix it?
(for me it fails at: `const allLines = structuredDiff.chunks.map...` so I think this should make the code safer at least, and it fixes the issue for me.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@
 
 <!-- Your comment below this -->
 
-<!-- Your comment above this -->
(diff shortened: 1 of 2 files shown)


orta:
This feels like a good spot 👍, yep

PR #928, opened 2019-10-13
Remove @babel/polyfill dependency

### LINKED ISSUE #857: Danger should get rid of deprecated @babel/polyfill
[@babel/polyfill](https://babeljs.io/docs/en/next/babel-polyfill.html) has been deprecated, but it is still being used as a dependency in danger JS, https://github.com/danger/danger-js/blob/master/package.json#L130.

We should remove it and use the replacement.

gpetrioli:
Closes #857


orta:
Yeah, looks like that fail is legit - the runtime wasn't as expected for the async function

Issue #930, opened 2019-10-18
Github Checks API Integration Error GHE


adam-bratin:
When running danger from a GHE repo using the checks API, danger is unable to connect to the checks API. I believe this is because requestAccessTokenForInstallation uses the hard coded https://api.github.com as the API root. I think support for the API ROOT env variable is needed. Will be opening PR with fix.

PR #931, opened 2019-10-18
Bugfix: #930

adam-bratin:
- fixes danger/danger-js#930
- add support for GHE in github checks integration


orta:
Yeah, makes sense to me - thanks @adam-bratin

PR #932, opened 2019-10-20
Move base to danger local

f-meloni:
I think the only one that uses base is `local`, the I've added the option only to `local` instead of having it in the shared args


orta:
Yeah, this makes sense to me 👍 - merge on green

PR #933, opened 2019-10-25
Fix GitHub comments overflow with icon

wenhuanli:
In the GitHub comment template the icon has fixed with of 50 and message has width of 100% of parent. It is possible to make text really take 100% of parent width and overlaps with icon. This happens especially when GitHub comment is forwarded to email and viewed on mobile phones that lines need break. It will take 100% width before it breaks.

**Outdated:** By deleting width=100% browser will make it take up the rest of the space, which is more proper IMO.

**Update:** By giving the first column a min-width of 50 same as fixed width, this will prevent text overlaps with the icon.

You can see the screenshots before and after the change on a phone.

Screenshot 2019-10-26 at 16 51 05
Screenshot 2019-10-26 at 16 51 30

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 ## Master
 
 <!-- Your comment below this -->
-
+- Fix Github comment template that message can overlap with icon when viewed in email notification - [@wenhuanli]
   <!-- Your comment above this -->
 
 # 9.2.2
--- a/source/runner/templates/_tests/__snapshots__/_githubIssueTemplates.test.ts.snap
+++ b/source/runner/templates/_tests/__snapshots__/_githubIssueTemplates.test.ts.snap
@@ -37,7 +37,7 @@ exports[`generating inline messages Shows correct messages for inline/regular vi
 <table>
   <thead>
     <tr>
-      <th width=\\"50\\"></th>
+      <th width=\\"50\\" style=\\"min-width:50\\"></th>
       <th width=\\"100%\\" data-danger-table=\\"true\\">Warnings</th>
     </tr>
   </thead>
@@ -79,7 +79,7 @@ exports[`generating inline messages Shows correct messages with custom icon for 
 <table>
   <thead>
     <tr>
-      <th width=\\"50\\"></th>
+      <th width=\\"50\\" style=\\"min-width:50\\"></th>
       <th width=\\"100%\\" data-danger-table=\\"true\\">Messages</th>
     </tr>
   </thead>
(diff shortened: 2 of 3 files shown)


orta:
I'm not sure I understand how this changes - can you include screenshots?

wenhuanli:
> I'm not sure I understand how this changes - can you include screenshots?

Thanks Orta. Please see updated screenshots.

orta:
Thanks, cool - The problem with this change for me is that is leads to ugly tables on the desktop website because markdowns vs warnings vs errors could have different table widths. Which isn't a trade-off I'm willing to make, given that most of the time people are using that to read the results


<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th  data-danger-table="true" data-kind="Warning">
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Tests were not updated</td>
    </tr>
  </tbody>
</table>

vs

<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Warning">
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Tests were not updated</td>
    </tr>
  </tbody>
</table>

wenhuanli:
<table>
  <thead>
    <tr width="100%">
      <th width="50"></th>
      <th data-danger-table="true">Warning</th>
    </tr>
  </thead>
  <tbody><tr>
      <td><g-emoji class="g-emoji" alias="warning" fallback-src="https://bbgithub.dev.bloomberg.com/images/icons/emoji/unicode/26a0.png">⚠️</g-emoji></td>
      <td>
<p>Short words</p>
  </td>
    </tr></tbody>
</table>

wenhuanli:
@orta Thanks for getting back to me with the concern. I experimented in the last comment and updated the commit. Please see if now it works for web as well? Thanks.

orta:
Maybe that works, but it can't hurt to try. 

GitHub doesn't support letting you set style on HTML elements in it's rendered HTML, so it's possible that this will do nothing in emails. This is the generated HTML from your markdown table:

Screen Shot 2019-10-27 at 8 12 41 AM

I doubt it will break anything, so if you want to give this a shot then you need to make the tests pass (looks like just snapshot) and a changelog entry and we'll get it in.

wenhuanli:
@orta Can you advise me what is the problem with the failed check (retriggred already)? Log says result didn't return in 10 minutes? Thanks.

PR #935, opened 2019-10-25
Support comment editing/deleting for Github Actions

jmenestr:
This PR removes the hardcoded userid for `Github Actions` workflows.

This hard coded value was causing the function

```ts
 getDangerCommentIDs = async (dangerID: string): Promise<string[]> => {
    const userID = await this.getUserID()
...
    return allComments
...
      .filter(comment => comment.user.id === userID) // Does it have the right user ID?
...
  }
````

to always fail because the hard coded userid would NEVER match a comment user id because it was not grabbing the correct userid associated with the given api key.

### DIFF
--- a/source/platforms/github/GitHubAPI.ts
+++ b/source/platforms/github/GitHubAPI.ts
@@ -130,11 +130,6 @@ export class GitHubAPI {
       return parseInt(perilID)
     }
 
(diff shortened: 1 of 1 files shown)


orta:
Ah, I bet during the alpha there was a consistent user id for all actions and then that changed.

PR #937, opened 2019-10-28
fix: use node 12 in Dockerfile

SimenB:
12 is LTS: https://nodejs.org/en/blog/release/v12.13.0/

any chance of a prebuilt image btw? building the entire danger every time is pretty inefficient

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10-slim
+FROM node:12-slim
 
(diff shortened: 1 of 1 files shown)


orta:
Cool, yeah, I'm open to it - ideally some kind of GitHub  action can upload the image to GPR after a tag ( maybe https://github.com/elgohr/Publish-Docker-Github-Action )

Issue #923, opened 2019-09-25
docker image on hub.docker.com is 10 month old


aol-nnov:
https://hub.docker.com/r/dangersystems/danger-js

Am I looking in the wrong place or it is not automatically built any more?

orta:
Yeah, they got hacked and leaked my GitHub creds, so I turned off my account. I don't use it, so I wasn't going to keep up risky infra for little gain.

aol-nnov:
Sad to hear that..

Well, I've reviewed which grants dockerhub requests from github, and permissions are very broad: full access to public and private repos, issues, etc..

So, for now, it seems to me, that keeping separate account for docker automated builds is sufficient. But it is, ahem, inconvenient!..

No docker builds for now, then.. Sad...

aol-nnov:
Anyway, may be, it's time to give them a second chance? Everyone gets hacked sooner or later nowadays..

rohit-smpx:
We could publish to Github instead https://github.com/features/package-registry ?

orta:
Great point, I have access to that

orta:
Still working on this https://github.com/danger/danger-js/actions

PR #938, opened 2019-10-28
Add a docker image

orta:
fixes https://github.com/danger/danger-js/issues/923

/cc @SimenB

Will need to wait for a new tag to verify it's all hooked up right though.

### DIFF
--- a/source/.github/workflows/uploadDockerImage.yml
+++ b/source/.github/workflows/uploadDockerImage.yml
@@ -1,0 +1,19 @@
+name: Upload tag to GitHub
+
+on:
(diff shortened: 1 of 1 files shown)


SimenB:
lovely, thanks! 🤞

SimenB:
did it work? does it need to be in root rather than `source/`?

orta:
Hrm, looks like some access issues now - https://github.com/danger/danger-js/commit/48ab882843f30f9fee8a9cbf48155307f5609f52/checks?check_suite_id=296468751

PR #940, opened 2019-11-03
Handle more events than just pull_request

f-meloni:
With GitHub actions you can have more events in the PR than just the simple `pull_request` one, like for example `pull_request_review` or `pull_request_review_comment`

### DIFF
--- a/source/ci_source/providers/GitHubActions.ts
+++ b/source/ci_source/providers/GitHubActions.ts
@@ -137,11 +137,13 @@ import { readFileSync, existsSync } from "fs"
 export class GitHubActions implements CISource {
   private event: any
 
-  constructor(private readonly env: Env) {
+  constructor(private readonly env: Env, event: any = undefined) {
     const { GITHUB_EVENT_PATH } = env
     const eventFilePath = GITHUB_EVENT_PATH || "/github/workflow/event.json"
-    
-    if (existsSync(eventFilePath)) {
+
+    if (event !== undefined) {
+      this.event = event
+    } else if (existsSync(eventFilePath)) {
       const event = readFileSync(eventFilePath, "utf8")
       this.event = JSON.parse(event)
     }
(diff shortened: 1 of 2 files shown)


source/ci_source/providers/GitHubActions.ts
```diff
@@ -161,21 +163,24 @@ export class GitHubActions implements CISource {
...
-    return this.env.GITHUB_EVENT_NAME !== "pull_request"
+    return this.event.pull_request !== undefined || this.event.issue !== undefined
   }
 
   get pullRequestID(): string {
-    if (this.env.GITHUB_EVENT_NAME === "pull_request") {
+    if (this.event.pull_request !== undefined) {
       return this.event.pull_request.number
+    } else if (this.event.issue !== undefined) {
+      return this.event.issue.number
```
f-meloni: This part is a little bit controversial, I've added it because technically the PR is a particular type of issue, and then I was thinking to someone that wants to run danger when a pr comment is added, and in that case I think he would have to use `issue_comment`, `pull_request_review` and `pull_request_review_comment` triggers.
In issue `comment` there is no pull request JSON, only the issue one https://developer.github.com/v3/activity/events/types/#webhook-payload-example-14

f-meloni: Happy to remove it if is a bad idea

f-meloni:
@orta looks like one of the build failed, but just needs to be restarted.
Unfortunately I can not do it, can you please help me with that?

source/ci_source/providers/GitHubActions.ts
```diff
@@ -137,11 +137,13 @@ import { readFileSync, existsSync } from "fs"
 export class GitHubActions implements CISource {
   private event: any
 
-  constructor(private readonly env: Env) {
+  constructor(private readonly env: Env, event: any = undefined) {
```
orta: I don't see this parameter ever being used?

f-meloni: I use it on tests, to avoid to create an event file

orta: oh yeah, the var gets set below by the file exists - make sense

source/ci_source/providers/GitHubActions.ts
orta: I'm open to seeing where this goes 👍

orta:
cool!

f-meloni:
Thank you @orta :)
I've also updated the CHANGELOG, when you have some time can you please release a new version? :)

orta:
alright, let's go then

PR #952, opened 2019-11-05
Fix github actions check

f-meloni:
Unfortunately after trying the yesterday PR I found some issues with the previous code.
The good news is that after this change I've tested it and works perfectly also with issue triggers

### DIFF
--- a/source/ci_source/providers/GitHubActions.ts
+++ b/source/ci_source/providers/GitHubActions.ts
@@ -163,7 +163,7 @@ export class GitHubActions implements CISource {
   }
 
   get useEventDSL() {
-    return this.event.pull_request !== undefined || this.event.issue !== undefined
+    return this.event.pull_request === undefined && this.event.issue === undefined
   }
 
   get pullRequestID(): string {
@@ -179,8 +179,8 @@ export class GitHubActions implements CISource {
   get repoSlug(): string {
     if (this.event.pull_request !== undefined) {
       return this.event.pull_request.base.repo.full_name
-    } else if (this.event.repo !== undefined) {
-      return this.event.repo.full_name
+    } else if (this.event.repository !== undefined) {
+      return this.event.repository.full_name
     }
 
     throw new Error("repoSlug was called on GitHubActions when it wasn't a PR")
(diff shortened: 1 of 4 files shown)


source/ci_source/providers/GitHubActions.ts
```diff
@@ -163,7 +163,7 @@ export class GitHubActions implements CISource {
   }
 
   get useEventDSL() {
-    return this.event.pull_request !== undefined || this.event.issue !== undefined
```
f-meloni: this before was false when the event was "pull_request", then the right check was the contrary of what I actually wrote

source/ci_source/providers/GitHubActions.ts
```diff
@@ -179,8 +179,8 @@ export class GitHubActions implements CISource {
   get repoSlug(): string {
     if (this.event.pull_request !== undefined) {
       return this.event.pull_request.base.repo.full_name
-    } else if (this.event.repo !== undefined) {
-      return this.event.repo.full_name
+    } else if (this.event.repository !== undefined) {
```
f-meloni: the field on the json is repository

source/platforms/github/GitHubAPI.ts
```diff
@@ -130,6 +130,11 @@ export class GitHubAPI {
       return parseInt(perilID)
     }
 
+    const useGitHubActionsID = process.env["GITHUB_WORKFLOW"]
```
f-meloni: After https://github.com/danger/danger-js/pull/935/commits/27da052541110cae73bb6f897dce0f3160aa69d6 
(on 9.2.3) github actions doesn't actually replace its messages.
This is because when you actually send the user request on action fails with

```
Request failed [403]: https://api.github.com/user
Response: {
 "message": "Resource not accessible by integration",
 "documentation_url": "https://developer.github.com/v3/users/#get-the-authenticated-user"
}
```

f-meloni: This changes was tested and now it works again

orta:
Nice one 👍

Issue #856, opened 2019-04-09
Is Danger always making a new comment?


orta:
I keep seeing notifications in Slack about new comments by Danger, when they should be being edited.

jmenestr:
I am also seeing this. I'm using `9.2.2` and currently see a new comment.

I can inspect each comment's markdown and see the same `dangerid` for each comment.

I was also under the impression that it should search all the comments for ones with the same `dangerid` and update them, rather than create a new one.

tamlenguyen:
Same for me, adding a new comment every time

nick28:
Same for me in Bitbucket Cloud, any updates or any leads how can we solve this?

orta:
Each platform has a `updateOrCreateComment` which must not be doing what we expect

- https://github.com/danger/danger-js/blob/6ae125dca11f88c270da0f15a16c91785d2304a7/source/platforms/BitBucketCloud.ts#L100
- https://github.com/danger/danger-js/blob/6ae125dca11f88c270da0f15a16c91785d2304a7/source/platforms/github/comms/issueCommenter.ts#L133

HelloCore:
> Same for me in Bitbucket Cloud, any updates or any leads how can we solve this?

For `Bitbucket Cloud`, we need `DANGER_BITBUCKETCLOUD_UUID` to detect the owner of the comment.
Could you check your UUID?
It has to be this format `{zzzzzzzz-zzzz-zzzz-zzzz-zzzzzzzzzzzz}`. (38 characters including brackets).

orta:
That doesn't feel right to me - the bot should be able to ask for it's own user ID via the API and it likely does

HelloCore:
Yes, that's right, I've never thought about that before. 
I'm going to work on this.

nick28:
> > Same for me in Bitbucket Cloud, any updates or any leads how can we solve this?
> 
> For `Bitbucket Cloud`, we need `DANGER_BITBUCKETCLOUD_UUID` to detect the owner of the comment.
> Could you check your UUID?
> It has to be this format `{zzzzzzzz-zzzz-zzzz-zzzz-zzzzzzzzzzzz}`. (38 characters including brackets).

I had set it as mentioned in the documentation. And yes it is in the format as mentioned by you

HelloCore:
> > > Same for me in Bitbucket Cloud, any updates or any leads how can we solve this?
> > 
> > 
[...]

PR #964, opened 2019-12-06
Bitbucket Cloud: Getting UUID from API

HelloCore:
From this issue, https://github.com/danger/danger-js/issues/856.
I think it's because it's difficult to get correct `UUID` from `Bitbucket Cloud`, so having Danger CI get it could be better.
However, we need extra permission (Read Account) to get the UUID.

I think we should not recommend UUID because it's easy to make a mistake (which I've done twice), so I've removed anything related to UUID from the document.


orta:
Oh great! This looks like I'd expect - thanks @HelloCore :D

PR #967, opened 2019-12-13
Fix a small bug in the protocol matching regex for DANGER_GITLAB_HOST

beezly:
Not a biggie but I'm about to change this to use the CI_API_V4_URL if DANGER_GITLAB_HOST isn't defined (so that the default behaviour isn't changed) and I spotted this needed fixing first.


orta:
Thanks 👍

PR #968, opened 2019-12-16
Use CI_API_V4_URL in GitLab if it is available

beezly:
I'm sure there is lots of room for improvement in this change! Please suggest whatever you think is appropriate. I'm particularly feeling some hnng around the way I'm testing this change.

Basically, since GitLab 11.7 an environment variable called `CI_API_V4_URL` has been available (per  https://docs.gitlab.com/ce/ci/variables/predefined_variables.html) which contains the information that we currently supply as `DANGER_GITLAB_HOST`.

I've left `DANGER_GITLAB_HOST` dominant because some people may have unusual configs where they have done things with hostnames that we don't expect and changing the behaviour may break things for them, but for new users it should now be possible to configure the CI side of Danger by just supplying a token.

### DIFF
--- a/docs/usage/gitlab.html.md
+++ b/docs/usage/gitlab.html.md
@@ -7,10 +7,13 @@ blurb: An overview of using Danger with GitLab, and some examples
 ---
 
 To use Danger JS with GitLab: you'll need to create a new account for Danger to use, then set the following environment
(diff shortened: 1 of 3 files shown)


orta:
Yes! Perfect, this is great 👍

PR #974, opened 2020-01-15
Fix danger local message when branch is not master

sk-:
The message when there are no changes had a hardcoded value of `master` instead of using the value of `base`, which coudl have been changed with the command line options.


orta:
Thanks!

Issue #969, opened 2019-12-19
Gitlab rRest api doesn't accept Project name anymore


farabi:
Hello.

Danger is trying to call the Gitlab rest api using the project name. while gitlab now only accepts the project ID as you can see in the documentation.

Here is the error, i get.

+ danger-swift ci
Error:  { HTTPError: Not Found
    at fn (/snapshot/danger-js/node_modules/ky/umd.js:242:12)
    at process._tickCallback (internal/process/next_tick.js:68:7)
  name: 'HTTPError',
  response:
   Response {
     size: 0,
     timeout: 0,
     [Symbol(Body internals)]: { body: [PassThrough], disturbed: true, error: null },
     [Symbol(Response internals)]:
      { url:
         'https://gitlab.xxxx.com/api/v4/projects/project%2Fname/merge_requests/15/changes',
        status: 404,
        statusText: 'Not Found',
        headers: [Headers] } },
  description: '404 Project Not Found' }

chebur:
same

orta:
Cool, makes sense - someone want to take a look at a fix?

farabi:
@orta 
I will !

farabi:
@orta  
Dumb question, but how to compile a danger-js to use it as a danger-swift dependency ?

HelloCore:
> @orta
> Dumb question, but how to compile a danger-js to use it as a danger-swift dependency ?

I believe it's this command `yarn run build; yarn run pkg . --output brew-distribution/danger;`.
Copied from https://github.com/danger/danger-js/blob/dab0e63a97705617c0e0017d85a9ff7129cadd5f/package.json#L69

f-meloni:
that could not work given https://github.com/danger/danger-js/issues/958 :(

HelloCore:
@farabi The pr https://github.com/danger/danger-js/pull/970 got merged to master, you could use `yarn run build; yarn run pkg . --output brew-distribution/danger;` to build the binary. Also, don’t forget to run `yarn install`

farabi:
Hello @orta  @chebur 

Actually gitlab does accept project name, the issue is coming from apache configuration where local gitlab is hosted.
Unless AllowEncodedSlashes is set to on.  rest calls with encoded slashes will not work. 
Like in my case : "https://gitlab.xxxx.com/api/v4/projects/project%2Fname/merge_requests/15/changes"

[...]

PR #976, opened 2020-01-21
Use gitlab project ID instead of project name

farabi:
#969  Here i added an environment variable for gitlab project id called `DANGER_GITLAB_PROJECT_ID`
Add added some doc about that. and updated GitlabAPI unit tests.
Project name will be used if project id variable  is not defined (so no breaking changes).


rohit-gohri:
Why was this closed? Did you get this working some other way?

farabi:
@rohit-gohri  here i [reopened the pull request ](https://github.com/danger/danger-js/pull/994)

PR #979, opened 2020-01-28
BitBucket Server: use REST API for raw file requests

minijus:
User centric request is changed to REST API request.
More details https://docs.atlassian.com/bitbucket-server/rest/6.10.0/bitbucket-rest.html#idp329


orta:
OK, CI fail should be fine when integrated on master - this looks good to me 👍

PR #982, opened 2020-02-01
Accept lowercase proxy env variables

f-meloni:
The proxy env variable is often used as lowercase string, then I think makes sense to automatically accept both.


orta:
I've never seen it but isn't offensive to me 👍 

reds are dues to a cache fail, skipping

PR #983, opened 2020-02-03
Upgrade to latest @octokit/rest

gr2m:
Hey friends 👋 

In preparation of [the upcoming `@octokit/rest` v17](https://github.com/octokit/rest.js/issues/1546), I'm testing against some dependents such as Danger. While at it I hope to be helpful with the upgrade.

I already found one problem that I'm addressing here https://github.com/octokit/rest.js/pull/1586

Besides that, I get errors when building:

```
node_modules/memfs-or-file-map-to-github-branch/build/index.d.ts:30:50 - error TS2709: Cannot use namespace 'GitHub' as a type.

30 export declare const memFSToGitHubCommits: (api: GitHub, volume: MemFSVolume, settings: BranchCreationConfig) => Promise<void>;
                                                    ~~~~~~
```

I guess the problem is [the change I did  in `@octokit/rest`'s index.d.ts](https://github.com/octokit/rest.js/pull/1586/files)?

TypeScript is still confusing the hell out of me. Got any pointers? Ideally I'd like to continue supporting both: `import Octokit from "@octokit/rest"` and `import { Octokit } from "@octokit/rest"` for TypeScript users. Starting with v17, it will only be `import { Octokit } from "@octokit/rest"`, so when in doubt, I don't mind breaking the `import Octokit from "@octokit/rest"` usage for TypeScript only. Either works in JS in the latest v16 versions

Note that this PR does relies on the change in https://github.com/octokit/rest.js/pull/1586 in order to reproduce the TypeScript build error.


orta:
Yeah, that error is because I'm doing [this here](https://github.com/orta/memfs-or-file-map-to-github-branch/blob/master/index.ts#L1) - I can fix that to `import { Octokit as GitHub } from "@octokit/rest"` 👍 

To support all forms of the export, you probably want something like:

```ts
export class OctoKit {}

export default OctoKit
```

Getting the syntax without the `* as` should probably still require [`esModuleInterop` being turned on](http://www.typescriptlang.org/v2/en/tsconfig#esModuleInterop)

gr2m:
> Yeah, that error is because I'm doing [this here](https://github.com/orta/memfs-or-file-map-to-github-branch/blob/master/index.ts#L1) - I can fix that to `import { Octokit as GitHub } from "@octokit/rest"` 👍

I was just looking into that, PR incoming

> To support all forms of the export, you probably want something like:
> 
> ```ts
> export class OctoKit {}
> 
> export default OctoKit
> ```

I tried that, but when doing

```ts
import * as OctokitDeprecated from "../index";
const { Octokit } = OctokitDeprecated

// Check constructor
new Octokit();

// Check deprecated constructor
new OctokitDeprecated();
```

I see this error

> This expression is not constructable.
> Type 'typeof import("/Users/gregor/Projects/octokit/rest.js/index")' has no construct signatures.ts(2351)

fbartho:
Yeah, my experience is that default exports explode often under TypeScript :-/

gr2m:
@orta can you publish a new version of `memfs-or-file-map-to-github-branch`?

orta:
Shipped - I don't have an answer about the default issue there. 

Does seem weird to me though.

gr2m:
I think we are all set

orta:
Yep, this works for me - thanks! Good luck with the release

PR #984, opened 2020-02-04
If no case is matched, use platform name.

Soyn:
When i use `danger local`. The platform name is `github` cause default case is `github`. I prefer to use platform name as default case.


orta:
I"m a little unsure what this does - how does this affect your usage?

Soyn:
When i debug on `danger local`. The `LocalGit` platform's name is `github`. The name is weird. It does not affect any usage.

orta:
👍 cool

Issue #972, opened 2020-01-13
danger local should consider uncommitted changes


sk-:
`danger local` only computes the differences between commits and does not consider changes not yet committed.

If I modify a file, I'd like to be able to locally check it before committing it, so that my local commits are compliant.

orta:
Yeah, I use it as a git commit hook - I'm open to a flag or to do some audits to decide if you want staged instead

f-meloni:
I think there was an initial implementation of it, but was not finished https://github.com/danger/danger-js/blob/master/source/commands/danger-local.ts#L14

sk-:
I think that getting staged and unstaged filed is fairly easy, we just need to run `git diff base` instead of `git diff base..HEAD`.

For untracked files, we would need to change the way the list of created files are computed. Currently they are only obtained from the diff. However, given that there's no diff for those files they will never be there, and we would need to use `git status`.

Also, I'm not sure what the implications would be if we add this unstaged or untracked files. as 
some other functions may fail.

WDYT?

orta:
I think it is probably worth faking the diff for the new files before it goes into the DSL generation, so that code can be consistent with `danger  pr` and `danger local` (without `--staging`)

Soyn:
@orta Did you mean when run `danger local`(with no flag from stdin), the uncommitted changes would be as the diff by default?

orta:
yep

Soyn:
I'm a little confused on your comment. You mean we should add the uncommitted diff in current function. https://github.com/danger/danger-js/blob/9ffa89ad76c553e8aff53a02baa37b9b8118e2ce/source/platforms/LocalGit.ts#L47

Soyn:
I have sent a pull request about this issue. I wonder this solution is the right way you think. @orta

PR #990, opened 2020-02-11
Add uncommited changeds to LocalGit

Soyn:
* details in https://github.com/danger/danger-js/issues/972

### DIFF
--- a/source/commands/danger-local.ts
+++ b/source/commands/danger-local.ts
@@ -24,7 +24,7 @@ setSharedArgs(program).parse(process.argv)
 const app = (program as any) as App
 const base = app.base || "master"
 
-const localPlatform = new LocalGit({ base, staged: app.staging })
+const localPlatform = new LocalGit({ base, staging: app.staging })
 localPlatform.validateThereAreChanges().then(changes => {
   if (changes) {
     const fakeSource = new FakeCI(process.env)
--- a/source/commands/danger-local.ts
+++ b/source/commands/danger-local.ts
@@ -16,9 +16,8 @@ interface App extends SharedCLI {
 
 program
   .usage("[options]")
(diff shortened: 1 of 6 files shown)


orta:
Yeah,  there should already be a `--staging` flag in there this can use from when I first built this out

Soyn:
I have updated the pull request.

source/dsl/cli-args.ts
```diff
@@ -15,6 +15,8 @@ export interface CliArgs {
   dangerfile: string
   /** So you can have many danger runs in one code review */
   id: string
+  /** Use staged changes */
+  staged?: boolean
```
orta: Are you sure this works? The option above is `  .option("-s, --staging", "Just use staged changes.")` - so I'd expect this to be `staging`?

Soyn: In my local test. It works well. I will update the pull request later.

Soyn:
Make `staged` to `staging`

source/commands/danger-local.ts
```diff
@@ -16,16 +16,15 @@ interface App extends SharedCLI {
...
-  // .option("-s, --staging", "Just use staged changes.")
   .description("Runs danger without PR metadata, useful for git hooks.")
+  .option("-s, --staging", "Just use staged changes.")
   .option("-b, --base [branch_name]", "Use a different base branch")
 setSharedArgs(program).parse(process.argv)
 
 const app = (program as any) as App
 const base = app.base || "master"
 
-const localPlatform = new LocalGit({ base, staged: app.staging })
```
orta: Oh, this is why it used to work - makes sense

orta:
Cool beans, thanks!

PR #993, opened 2020-02-13
Update `parse-diff` library

417-72KI:
`parse-diff` 0.7.0 has been released.
It fundamentally resolves sergeyt/parse-diff#23, which is based on danger/danger-js#807😄


orta:
Great!

PR #996, opened 2020-02-15
Fix import in template script

### LINKED ISSUE #995: Typescript build fails when using danger for a plugin
The build fails with this error:

```
node_modules/danger/distribution/dsl/GitHubDSL.d.ts:24:10 - error TS2709: Cannot use namespace 'GitHub' as a type.

24     api: GitHub;
            ~~~~~~


Found 1 error.
```

The error appears with `@octokit/rest` v16.43.1 but not before that.

From what I could gather the distribution template in https://github.com/danger/danger-js/blob/d43ae5aa7d5d53008425f13e8059b32757851722/scripts/danger-dts.ts#L15 didn't get updated when  PR https://github.com/danger/danger-js/pull/983 was done.

Also we should probably update the version in package.json to `^16.43.1`

rohit-gohri:
Closes #995


orta:
Nice work 👍

rohit-gohri:
> Nice work +1

Should I merge this?

orta:
You assigned yourself, so I figure you would 👍

PR #1002, opened 2020-02-25
Add structured diff support to Gitlab

rohit-gohri:
Currently gitlab platform doesn't support diffs either through a FullDiff implementation or through structured diffs:
https://github.com/danger/danger-js/blob/6e4371bca1015452375524118aaebb28126c33a6/source/platforms/gitlab/GitLabGit.ts#L14-L21

Gitlab exposes a full `.diff` file for merge requests: eg. https://gitlab.com/gitlab-org/gitlab-foss/merge_requests/27117.diff
But that only works for public repos, can't access it through API (AFAIK) so not viable to use.

But it's API exposes individual file diffs in Compare API: https://docs.gitlab.com/ce/api/repositories.html#compare-branches-tags-or-commits

The format is a bit different from standard git diff, some things are parsed and removed from it (Open Issue: https://gitlab.com/gitlab-org/gitlab/issues/24913). 

So added a function to parse this diff format and create a full diff from it similar to one in danger-ruby:
https://github.com/danger/danger/blob/aa490b4ca59b4b13acaea8f2b11372d06baeec03/lib/danger/request_sources/gitlab.rb#L95-L107


rohit-gohri:
@jamime @notjosh @bigkraig If either of you could take a look that'll be great!

source/platforms/gitlab/GitLabAPI.ts
```diff
@@ -97,18 +102,18 @@ class GitLabAPI {
 
   getMergeRequestCommits = async (): Promise<GitLabMRCommit[]> => {
     this.d("getMergeRequestCommits", this.repoMetadata.repoSlug, this.repoMetadata.pullRequestID)
-    const commits: GitLabMRCommit[] = await this.api.MergeRequests.commits(
```
orta: These changes aren't great - having the type reference on the left means the type system checks to validate, having them on the right means there's no type safety anymore. Are you sure this is what you want?

rohit-gohri: Originally the gitlab npm module's types were not being used:
(currently in master)
https://github.com/danger/danger-js/blob/04c33453b3216b24718df571235064bbc7e659ab/source/platforms/gitlab/GitLabAPI.ts#L55
https://github.com/danger/danger-js/blob/04c33453b3216b24718df571235064bbc7e659ab/source/platforms/gitlab/GitLabAPI.ts#L61
So, all the method results had return type `any` which would get casted to whatever we put on the left side.

I changed it to use the npm module types:
https://github.com/rohit-gohri/danger-js/blob/13fe5ebe38407e2ff4abe491e5340c0fe4d07ee8/source/platforms/gitlab/GitLabAPI.ts#L57
```private api: InstanceType<typeof Gitlab>```

This way we at least we get type checks for the module's methods. But the module's type definitions define all the response types as `Promise<object | object[]`> and that doesn't get cast automatically hence the need for `as GitLabMRCommit[]`. So overall we are gaining more type safety (any deprecated methods would break build instead of at runtime).

I hope this makes sense.

orta: It does - thanks

PR #925, opened 2019-09-29
Azure DevOps support

devandanger:
Azure DevOps provides all capabilities for Danger in theory.
Has source repositories.
Has PR capabilities.
Has CI driving it.

Danger currently supports mostly things around github/bitbucket/circle/travis.  Need to generacize these capabilities to make it easier to add-in Azure Devops.

- [ ] Reposlug functionality is geared only to Azure Devops support
- [ ] CI Type is only for travis / circle. Unclear where this capability is required.
- [ ] PR process only for danger.

PR #1006, opened 2020-03-03
Azure DevOps support

jonasbark:
This is a follow up to https://github.com/danger/danger-js/pull/925
I tested this with Bitbucket Cloud and it worked fine. 

In order to make it compile I had to do changes that I probably shouldn't have done.  These are outlined here: https://github.com/feilfeilundfeil/danger-js/commit/bf207c87ecdc78a0e90f254a276c518f66185ff5

Unfortunately I'm not capable enough on the Typescript side, yet, to fix these odd changes.


source/ci_source/providers/index.ts
```diff
@@ -78,7 +80,8 @@ const realProviders = [
   CodeBuild,
   Codefresh,
   AppCenter,
-  Cirrus
+  Cirrus,
+  AzureDevops,
 ]
```
orta: I think bringing just this and the Providers change would get you most bang for least effort, this means Danger will at least run on the CI

jonasbark: I'm not sure if I understand that correctly. Do you want me to get rid of any of the changes not related to the Azure Devops Provider?

orta: Yeah, that would make it usable, then everything else can be built on top of that PR (instead of of making a mega PR - the `danger init` stuff could then come in as it's own PR too for example

PR #1011, opened 2020-03-05
Fix "@octokit/rest" deprecation warning when using ".issues.addLabels()"

sogame:
When adding a label using `danger.github.utils.createOrAddLabel` the following message appears in the CI logs:

```
Deprecation: [@octokit/rest] "number" parameter is deprecated for ".issues.addLabels()". Use "issue_number" instead
```

This change should prevent it.


orta:
Thanks!

PR #1013, opened 2020-03-06
Fix GitHub Actions condition

### LINKED ISSUE #856: Is Danger always making a new comment?

### LINKED ISSUE #936: DangerJS in GitHub actions does not overwrite or delete previous comments
Like in other CI workflows, I would expect any dangerjs code run in github actions to correctly update/delete previous comments made by previous builds. This currently does not happen, resulting in a *new* comment being created for every run.

While not functionally blocking, it create a rather poor experience because it results in a lot of noise.

I think I've identified the issue at these lines:
https://github.com/danger/danger-js/blob/master/source/platforms/github/GitHubAPI.ts#L133-L136
It seems that the userID is hardcoded for github actions and always returns the same value rather than looking up the user associated with the current GITHUB_TOKEN.

As a result of the hardcoded id, when the API goes to look for previous comments that are created by danger to update/delete them, there's a check on the userid in a `filter` and because the hard coded will probably never match the actual userid, it never gets a list of comments to actually update/delete.

https://github.com/danger/danger-js/blob/master/source/platforms/github/GitHubAPI.ts#L85-L96

I'm mot sure if this behavior was intentional, or if it was just something that was missed.

rohit-gohri:
Closes #856 & closes #936

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,9 @@ x
 
 <!-- Your comment below this -->
 
+- Don't use hardcoded userId to update comments if using personal token in Github Actions - [@rohit-gohri]
 - Add the staged flag to `danger local` command - [@soyn]
-- Disable warning in Github Action if using DANGER_GITHUB_API_TOKEN- [@rohit-gohri]
+- Disable warning in Github Action if using DANGER_GITHUB_API_TOKEN - [@rohit-gohri]
 - Update `parse-diff` library - [@417-72KI]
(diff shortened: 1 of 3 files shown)


rohit-gohri:
Not sure why the build is failing, it gets stuck on cache. All the tests are passing on local

orta:
Cool, let's get this in

Issue #985, opened 2020-02-05
Danger regularly runs out of memory in GitHub Actions


dkundel:
Hi there,

I'm running Danger for a project using GitHub Actions. For most of the time this worked great but recently it started crashing on every run because it's running out of memory. There hasn't been any changes to the GitHub workflow file or to the Dangerfile. 

You can find an example of the error here https://github.com/twilio-labs/open-pixel-art/pull/1946/checks?check_run_id=426717243

<details>
<summary>Full error message</summary>
<pre>
2020-02-05T02:21:23.1759542Z [36;1mnpx danger ci --text-only -f[0m
2020-02-05T02:21:23.1795388Z shell: /bin/bash -e {0}
2020-02-05T02:21:23.1795497Z env:
2020-02-05T02:21:23.1796296Z   GITHUB_TOKEN: ***
2020-02-05T02:21:23.1796411Z ##[endgroup]
2020-02-05T02:22:11.7558013Z 
2020-02-05T02:22:11.7594992Z <--- Last few GCs --->
2020-02-05T02:22:11.7595163Z 
2020-02-05T02:22:11.7596016Z [2949:0x36c5cf0]    45683 ms: Scavenge 1139.9 (1437.5) -> 1139.3 (1442.0) MB, 9.3 / 0.0 ms  (average mu = 0.084, current mu = 0.073) allocation failure 
2020-02-05T02:22:11.7596870Z [2949:0x36c5cf0]    46291 ms: Mark-sweep 1146.3 (1442.0) -> 1133.8 (1437.5) MB, 583.0 / 0.0 ms  (average mu = 0.087, current mu = 0.089) allocation failure scavenge might not succeed
2020-02-05T02:22:11.7597634Z [2949:0x36c5cf0]    46344 ms: Scavenge 1146.9 (1437.5) -> 1146.8 (1440.5) MB, 17.8 / 0.0 ms  (average mu = 0.087, current mu = 0.089) allocation failure 
2020-02-05T02:22:11.7597818Z 
2020-02-05T02:22:11.7597886Z 
2020-02-05T02:22:11.7598487Z <--- JS stacktrace --->
2020-02-05T02:22:11.7598572Z 
2020-02-05T02:22:11.7598786Z ==== JS stack trace =========================================
2020-02-05T02:22:11.7598877Z 
2020-02-05T02:22:11.7599074Z Security context: 0x281c7c21e6e9 <JSObject>
[...]

PR #1019, opened 2020-03-20
fix: change JSON patch implementation for memory improvement

dkundel:
I've done some digging into #985 and here's the outcome.

**Actual Issue**:
The issue seems to be in the algorithm that the `rfc6902` library is using for Arrays. Since it's recursive it will run out of memory when it encounters an array that is too big. 

**Possible Solution**:
I found in this issue https://github.com/chbrown/rfc6902/issues/39 an alternative library (`fast-json-patch`) that could be used. It seems to work against my use case and according to [these benchmarks](https://github.com/Starcounter-Jack/JSON-Patch#performance) it should be significantly more performant.

**Breaking Change**:
As you can see in the updated tests, there are two changes to the output format compared to the other implementation:
1. Order of operations is different
2. `path` for insertions in arrays ends with new index of item instead of `-`. For example: `/pixels/99` instead of `/pixels/-`. I checked the actual standard and this behavior seems to be up to the actual standard: https://tools.ietf.org/html/rfc6902#appendix-A.2

Let me know what you think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ x
 
 <!-- Your comment below this -->
 
+- Changed JSON patch implementation for better memory performance. [@dkundel] **Breaking:** `JSONPatchForFile` will
+  return a different order of operations than previously. It will also return a `path` with the index of the element
+  inserted into an array for `add` operations.
+
 <!-- Your comment above this -->
 
 # 9.3.1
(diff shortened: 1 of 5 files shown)


orta:
I think this is a great call. 

I'll call it a major semver bump (we've had them for trivial things before) - I think it'd be unlikely people are expecting the order to be determinate but at least this warns folks upfront

PR #1020, opened 2020-03-26
Use pagination when fetching existing labels

sogame:
`listLabelsForRepo` has a limit of 100 items, that means that if a repo has more than 100 labels using `createOrAddLabel` can try to create an already existing label.

This PR replaces `listLabelsForRepo` with an endpoint request following all pages.

### DIFF
--- a/source/platforms/github/GitHubUtils.ts
+++ b/source/platforms/github/GitHubUtils.ts
@@ -194,8 +194,8 @@ export const createOrAddLabel = (pr: GitHubPRDSL | undefined, api: GitHub) => as
   d("Checking for existing labels")
   let label: any = null
   try {
-    const existingLabels = await api.issues.listLabelsForRepo({ owner: config.owner, repo: config.repo })
-    label = existingLabels.data.find((l: any) => l.name == labelConfig.name)
+    const existingLabels = await api.paginate('GET /repos/:owner/:repo/labels', { owner: config.owner, repo: config.repo })
+    label = existingLabels.find((l: any) => l.name == labelConfig.name)
   } catch (e) {
     d("api.issues.getLabels gave an error", e)
   }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ x
 <!-- Your comment below this -->
 
 - Pass process arguments back to the original process [@f-meloni]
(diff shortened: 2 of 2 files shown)


source/platforms/github/GitHubUtils.ts
```diff
@@ -194,7 +194,7 @@ export const createOrAddLabel = (pr: GitHubPRDSL | undefined, api: GitHub) => as
   d("Checking for existing labels")
   let label: any = null
   try {
-    const existingLabels = await api.issues.listLabelsForRepo({ owner: config.owner, repo: config.repo })
+    const existingLabels = await api.paginate('GET /repos/:owner/:repo/labels', { owner: config.owner, repo: config.repo })
```
sogame: I haven't been able to test this, but checking the documentation, seems like the right way of doing this 😅 
https://github.com/octokit/rest.js/tree/8373934ed21eb69c8dbacc3e06e1dc1566a6a7d8#pagination

orta: That's a lot of labels! So this doesn't change the response shape?

sogame: I know 😅 But I use a repo that gets automatic dependency bumps and each of them generates a new label 😓 

Looking at what I think is the source code, `listLabelsForRepo` seems to be doing this call internally, so the response should be the same
https://github.com/octokit/plugin-rest-endpoint-methods.js/blob/cb9d3e408a245e8b49971cf61bf76dcab4742669/src/generated/endpoints.ts#L458

orta: The CI fails are probably correct in this case I think

sogame: You're totally right @orta, the build was failing because the response changes slightly and `.data` is not needed (`existingLabels.find` instead of `existingLabels.data.find`).

The build still failing, but only for Node 10.16, because "No output has been received in the last 10m0s". I've tried it a couple of times. Do you think that's related to my changes, or it's some Travis issue?

orta:
Yeah, I don't get why it's always that one. Should probably move that specific run to GitHub Actions to spread it out a bit

sogame:
I've rebased from master and fixed the conflicts in `changelog.md` :-)

orta:
Rock!

PR #1021, opened 2020-03-29
Pass process arguments back to the original process

f-meloni:
With this PR the process that launched danger with `--process`  will receive back all the parameters that were sent to danger (`--dangerfile` for example, plus all the custom ones that the other process handles)
In order to make this work, given the old logic was to have a call with no parameters on it (e.g. `danger-swift`), I'm adding `runner` to the call. (e.g. `danger-swift runner $parameters`)
This also could avoid the need to parse the DSL in the runner to get parameters like the dangerfile path, because you would receive them in the parameters list.

`Danger-swift` supports this already from 1.0.0 https://github.com/danger/swift/blob/master/Sources/Runner/main.swift#L16-L17
`Danger-Python` and `Danger-Kotlin` don't, but add the support for this is fairly trivial and I think they are both not production ready yet, then this could be an acceptable trade off.
The question is: would this break `Peril`?


f-meloni:
cc. @gianluz @turekj

gianluz:
make sense to me!

orta:
Peril doesn't use any of the command APIs - it's totally safe with this change

f-meloni:
Great!

orta:
Can you rebased? Think i fixed the ci 

Other than that and a changelog, this feels reasonable to me

f-meloni:
I saw you change then I merged master into the branch, we can squash and merge though if you want to keep the history clean.
Will update the changelog

f-meloni:
PS. @orta I've noticed that Danger is not commenting anymore on `danger-js` PRs, is it because is using the old danger user token that now is invalid? In that case should we use the new `danger-public` one?

orta:
Aye, aye, it should

PR #1024, opened 2020-03-30
added a new Bug issue template

NotMoni:
We should add a new template for bug reports. This will help others when they want to report a bug.


orta:
Cool, thanks!

NotMoni:
My pleasure :D

PR #1026, opened 2020-04-01
adding bamboo support (working with bitbucket server)

tim3trick:
In the absence of repository systems beside bitbucket i could not validate the regex used for getting the repository slug (looks like it should work with github as well)


orta:
Cool, yeah, this PR looks good - thanks!

PR #1058, opened 2020-07-23
Fix danger local with non master base branch

### LINKED ISSUE #1057: [BUG] renaming master branch causes `danger local` to fail even when base branch is provided
**Describe the bug**
We renamed our `master` branch to `main` and `danger local -b main` fails

**To Reproduce**
1. `git checkout master`
2. `git branch -m main`
3. `git checkout -b testbranch`
4. `yarn run danger local -b main`

**Expected behavior**
`yarn run danger local -b main` runs danger without error

**Screenshots**

    yarn danger local -b main
    yarn run v1.22.4
    $ /opt/docker/react-uswds/node_modules/.bin/danger local -b main
    Could not get diff from git between master and HEAD
    
    Error: fatal: ambiguous argument 'master...HEAD': unknown revision or path not in the working tree.
    Use '--' to separate paths from revisions, like this:
    'git <command> [<revision>...] -- [<file>...]'
    
        at Socket.<anonymous> (/opt/docker/react-uswds/node_modules/danger/distribution/platforms/git/localGetDiff.js:20:19)
        at Socket.emit (events.js:310:20)
        at addChunk (_stream_readable.js:286:12)
        at readableAddChunk (_stream_readable.js:268:9)
        at Socket.Readable.push (_stream_readable.js:209:10)
        at Pipe.onStreamRead (internal/stream_base_commons.js:186:23)
    
    /opt/docker/react-uswds/node_modules/danger/distribution/commands/danger-runner.js:119
        d("Got md " + results.markdowns.length + " w " + results.warnings.length + " f " + results.fails.length + " m " + results.messages.length);
                              ^
    
    TypeError: Cannot read property 'markdowns' of undefined
        at /opt/docker/react-uswds/node_modules/danger/distribution/commands/danger-runner.js:119:27
        at /opt/docker/react-uswds/node_modules/node-cleanup/node-cleanup.js:80:9
        at Array.forEach (<anonymous>)
        at process.exitHandler (/opt/docker/react-uswds/node_modules/node-cleanup/node-cleanup.js:79:21)
        at process.emit (events.js:310:20)
        at process.exit (internal/process/per_thread.js:166:15)
        at process.exceptionHandler (/opt/docker/react-uswds/node_modules/node-cleanup/node-cleanup.js:74:13)
        at process.emit (events.js:310:20)
        at process._fatalException (internal/process/execution.js:164:25)
    
    Danger: ⅹ Failing the build, there is 1 fail.
    ## Failures
    `node` failed.
    ## Markdowns
    ### Log
    
    
    ```sh
    
    ```
    
    
    error Command failed with exit code 7.
    info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

**Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          | 10.2.1
| node             | v12.16.3
| npm              | 6.14.4
| Operating System | Linux

**Additional context**

ahobson:
This fixes #1057

### DIFF
--- a/source/runner/dangerDSLJSON.ts
+++ b/source/runner/dangerDSLJSON.ts
@@ -37,7 +37,11 @@ export class DangerDSLJSON implements DangerDSLJSONType {
     const parsedString = JSON.parse(JSONString)
     Object.assign(this, parsedString.danger)
 
+    // Merge the command line settings with the settings from the
+    // original command invocation.  This is needed because some
+    // commands like danger-local have options that are unknown to
+    // danger-runner
     // @ts-ignore
-    this.settings.cliArgs = cliArgs
+    Object.assign(this.settings.cliArgs, cliArgs)
   }
 }


source/commands/utils/sharedDangerfileArgs.ts
```diff
@@ -7,6 +7,10 @@ process.on("unhandledRejection", function(reason: string, _p: any) {
 })
 
 export interface SharedCLI extends program.CommanderStatic {
+  /** What should we compare against? */
+  base?: string
+  /** Should we run against current staged changes? */
+  staging?: boolean
```
f-meloni, line 13: Wouldn't this add this parameter also to the other commands, and then to their help to?

danger-public:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

Please add your change and name to the master section.
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 28e06a26be464f50f3089795080768932397eecb
</p>

source/commands/utils/sharedDangerfileArgs.ts
ahobson: It would, but I think that's a feature so that the base comparison branch will work for all commands.  This is needed as I mention in #1057 since danger-runner winds up being invoked.

I did suggest another idea in #1057 but there was a comment that the shared argument solution was preferable.

I'm happy to rework to make it more appropriate

f-meloni: The other commands are `pr` and `ci` `runner` etc.
In those context I think `base` and `staging` would not make sense, because you would be running it against a PR, and not a local branch.
The `pr` command has a specific command too, https://github.com/danger/danger-js/blob/master/source/commands/danger-pr.ts#L93 and in that case it works, then why does this need to be on the SharedCLI instead? (just asking to try to get more context)

ahobson: I'm really not familiar with the code base, but I think that's because those options are only needed before invoking `danger runner` but the base branch is needed after invoking `danger runner`.

f-meloni: Isn't the value taken at line 17 and 19, really similarly to what happens in https://github.com/danger/danger-js/blob/master/source/commands/danger-pr.ts#L93 ? (Again is just me trying to understand better)

ahobson: Perhaps my analysis in #1057 is wrong, but I believe what is happening is that `danger-local` winds up invoking `danger-runner` and passes the base branch argument, but because it's not a shared option, that override is lost.

Note that in `danger-pr` the [options are removed](https://github.com/danger/danger-js/blob/master/source/commands/danger-pr.ts#L124-L125) and so `isJSON` is never passed to `danger-runner`.  The base branch option needs to be interpreted by `danger-runner` somehow and the shared arguments is the most obvious way to do that.

ahobson: I'm not positive about this, but it does seem like the cliArgs are used inside the runner [only if the command is not using a platform](https://github.com/danger/danger-js/blob/master/source/runner/jsonToDSL.ts#L41-L42).  If that's the case, then it's only `danger-local.ts` that needs to send the base branch.

But `danger-runner.ts` is invoked by `danger-pr`, `danger-ci`, and `danger-local` (and others), so I think that means we need to merge the cliArgs when generating the json.  I've updated the PR

danger-public:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

Please add your change and name to the master section.
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 5ee6491a7506a82703e95f9f54f84334212d1813
</p>

orta:
Yeah, I think this is the original intent 👍

orta:
Yeah, I think it should be pretty harmless to pass through arbitrary CLI data regardless. Thanks for taking it up @ahobson

orta:
Shipped as 10.3.0 because apparantly I can't remember which is patch or minor it seems 👍

PR #1076, opened 2020-10-09
Bitbucket Cloud: Fix bug when Danger updating inline comment with summary comment

HelloCore:
When updating summary comment, Danger will call `getDangerComments` which will get all Danger's comments (including inline comment) for updating.
We need to filter inline comments out, so Danger will only update the main comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@
 <!-- Your comment below this -->
 
 - Bitbucket Cloud: Fix bug when Danger updating inline comment with summary comment. - [@hellocore]
-  <!-- Your comment above this -->
+
+<!-- Your comment above this -->
 
 # 10.4.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@
 
 <!-- Your comment below this -->
 
-<!-- Your comment above this -->
+- Bitbucket Cloud: Fix bug when Danger updating inline comment with summary comment. - [@hellocore]
+  <!-- Your comment above this -->
 
 # 10.4.2
 
--- a/source/platforms/bitbucket_cloud/BitBucketCloudAPI.ts
+++ b/source/platforms/bitbucket_cloud/BitBucketCloudAPI.ts
@@ -214,11 +214,12 @@ export class BitBucketCloudAPI {
     return await res.text()
   }
 
-  getDangerComments = async (dangerID: string): Promise<BitBucketCloudPRComment[]> => {
+  getDangerMainComments = async (dangerID: string): Promise<BitBucketCloudPRComment[]> => {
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -15,7 +15,8 @@
 
 <!-- Your comment below this -->
 
-<!-- Your comment above this -->
+- Bitbucket Cloud: Fix bug when Danger updating inline comment with summary comment. - [@hellocore]
+  <!-- Your comment above this -->
```
fbartho, line 19:
```suggestion

<!-- Your comment above this -->
```

source/platforms/bitbucket_cloud/BitBucketCloudAPI.ts
```diff
@@ -214,11 +214,12 @@ export class BitBucketCloudAPI {
     return await res.text()
   }
 
-  getDangerComments = async (dangerID: string): Promise<BitBucketCloudPRComment[]> => {
+  getDangerMainComments = async (dangerID: string): Promise<BitBucketCloudPRComment[]> => {
```
fbartho, line 217: `getDangerComments` is a part of the interface for all API providers, so you probably shouldn't modify this implementation & rename it.

Would you mind either putting a helper in the BitBucketCloud.ts file (since that's the only place that is calling this `getDangerMainComments` method and/or duplicating the implementation (with your changes) into a new method?

fbartho (changes requested):
I don't use BitBucket myself, so I cannot test this, but it seems pretty straightforward.

I've provided some minor change requests above, but once those are addressed this seems reasonable to me.

source/platforms/bitbucket_cloud/BitBucketCloudAPI.ts
HelloCore: I don't think `getDangerComments` is a part of the interface because we only have this function in `BitBucketServerAPI` and `BitBucketCloudAPI`. 

I believe the name `getDangerComments` would confused us, since we don't know if the result would be main comment or inline comment.

Moreover, we don't really have the usecase that we need both main comments, and inline comments at the same time, so I renamed it to `getDangerMainComments`

orta: This felt off to me also, but I think I see your reasoning 👍🏻

orta:
👍🏻

orta:
I'll make a release on the next PR

orta:
I'll make a release on the next PR

source/platforms/bitbucket_cloud/BitBucketCloudAPI.ts
fbartho: Sounds good! Sorry that i didn't dive deeper to verify my comments!

PR #1086, opened 2020-11-18
add danger.git.head and danger.git.base

### LINKED ISSUE #1085: danger.git.head && danger.git.base
It'd be super helpful if danger could report the commits that it is comparing against

This would allow you to shell out to `git` with the same information danger itself has

```js
execSync(`git show ${danger.git.head}:package.json`).toString()
execSync(`git show ${danger.git.base}:package.json`).toString()
```

Right now I tried to implement this with `danger.git.commits` and at least within `danger local` its not enough to do this correctly so I'm left to parse danger's argv myself

jamiebuilds:
Resolves #1085

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 
 <!-- Your comment below this -->
 
+- Adds `danger.git.head` and `danger.git.base`
+
 <!-- Your comment above this -->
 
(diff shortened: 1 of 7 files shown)


orta (approved):

orta:
Cool, looks good

PR #1093, opened 2020-12-15
Add env variable to disable transpiling with typescript

OzzieOrca:
- Allows developer to force TS to be compiled with babel

I'm having trouble using `tsc` to compile `dangerfile.ts` and it's currently throwing `SyntaxError: Cannot use import statement outside a module`.

My workflow runs all `.ts` files through Babel only, uses `tsc` as a linter only, and I have `target` set to `esnext` in my `tsconfig.json` file:
```json
{
    "compilerOptions": {
        "target": "esnext",
        ...
    },
}
```

Danger seems to be transpiling with `tsc` but `node` can't execute its output since the target is so high. This PR adds the `DANGER_DISABLE_TSC` flag for a developer to skip `tsc` and use Babel.

Idk if this is desired to be added to this project and I'm open to other implementations. Maybe the logic should be less about disabling tsc and more about forcing Babel. Or maybe a more robust solution to specify a specific transpiler would be better such as `DANGER_TRANSPILATION=off|babel|tsc`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@
 
 <!-- Your comment below this -->
 
-<!-- Your comment above this -->
+- Add `DANGER_DISABLE_TSC` environment variable to disable transpiling with tsc, providing a way to force transpiling
+  with Babel - [@ozzieorca]
+  <!-- Your comment above this -->
 
 # 10.5.4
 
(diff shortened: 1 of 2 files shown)


danger-public:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

Please add your change and name to the master section.
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against ae9bb509fb1853e3bd87424e83d350be6931f7a7
</p>

orta:
Yeah, I think this is fine - I doubt there'll ever be a third transpiler option, so in theory with these two options you can set it to any of the three.

orta:
Shipped in 10.6.0

PR #1113, opened 2021-02-18
Fix Bitrise's `ciRunURL` underlying env var

rogerluan:
Hey 👋 

During some debugging process I realized that this env var is probably wrong, based on what it means and what other similar CI files are using as env var values 😊 So here's the fix!

I also added a handy link at the bottom of the file, copying `GitLabCI.ts` and `Travis.ts` 📜

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 ## Main
 
 <!-- Your comment below this -->
-
+- Fixed Bitrise's `ciRunURL` underlying env var - [@rogerluan]
 <!-- Your comment above this -->
 
 # 10.6.2
--- a/source/ci_source/providers/Bitrise.ts
+++ b/source/ci_source/providers/Bitrise.ts
@@ -71,10 +71,12 @@ export class Bitrise implements CISource {
   }
 
   get ciRunURL() {
-    return this.env.BITRISE_PULL_REQUEST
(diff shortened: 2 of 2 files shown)


danger-public:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

Please add your change and name to the main section.
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 4a656a3fdd97d793ca40fb6cb367bf5517f3bf45
</p>

orta:
Is it possible for people to have older versions of bitrise ( and thus need the old version around?)

rogerluan:
Hey @orta ! Just like [in my other PR](https://github.com/danger/danger-js/pull/1115#issuecomment-781357460), both of these env vars (old and new) are also global env vars and not produced by a step, so they should be the same for everyone.

But in this case I don't even think this was ever correct - as the old env var `BITRISE_PULL_REQUEST` represents just a number (e.g. `1113`), not anything else (like the PR link or anything). This might've been a typo or copy'n'paste oversight. The underlying value of this env var was expected to be a number because that's even in the `isPR()` function a few lines above 😄 

So I'd say the answer is no! :)

PR #1115, opened 2021-02-18
Simplify Bitrise repo slug lookup

### LINKED ISSUE #1047: [BUG] Integration with Bitrise BitBucketServer doesn't parse URL correctly when using ssh
**Describe the bug**
The rest API request always fails (404) because the URL used isn't parsed correctly if the GIT_REPOSITORY_URL uses ssh
I backtracked to Bitrise.ts:59 but I might be wrong

**To Reproduce**
1. Set up danger-swift in Bitrise in a repo hosted in Bitbucket Server
2. Set ENV variables 
3. Set a Git Clone Repository step in your Bitrise workflow
4. Set a script step in your Bitrise workflow that installs danger-swift and runs danger-swift ci

**Expected behavior**
danger-swift ci should run correctly

**Screenshots**
Screenshot 2020-05-29 at 13 12 36

**Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          | 10.2.0
| Operating System | Bitrise Env Xcode 11.3.x, on macOS 10.14.6 (Mojave)

**Additional context**
Workarounding and changing the `GIT_REPOSITORY_URL` manually to the https equivalent solved the problem.

rogerluan:
Hey 👋 

I realized that we had this issue related to Bitrise https://github.com/danger/danger-js/issues/1047 and thought it was easy enough to fix. This PR removes `GIT_REPOSITORY_URL` and uses other env vars that are more user friendly to parse, removing the regex completely - which's also safer and now works for all sorts of URLs.

With my little understanding of how Danger and CI interacts, I _hope_ that simply and blindly dropping the support to `GIT_REPOSITORY_URL` won't break existing users' setups 😬 

Resolves #1047
Docs: https://devcenter.bitrise.io/builds/available-environment-variables/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 <!-- Your comment below this -->
 - Simplified Bitrise repo slug lookup, fixing SSH URL parsing in BitBucketServer - [@rogerluan]
+- Changed `CHANGELOG.md` merge strategy to union to avoid merge conflicts - [@rogerluan]
 - Log failure to update status also when not in verbose mode - [@rogerluan]
 <!-- Your comment above this -->
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 ## Main
 
 <!-- Your comment below this -->
-
+- Simplified Bitrise repo slug lookup, fixing SSH URL parsing in BitBucketServer - [@rogerluan]
 <!-- Your comment above this -->
 
 # 10.6.2
--- a/source/ci_source/providers/Bitrise.ts
+++ b/source/ci_source/providers/Bitrise.ts
@@ -51,15 +51,9 @@ export class Bitrise implements CISource {
   }
 
   get isPR(): boolean {
-    const mustHave = ["GIT_REPOSITORY_URL"]
+    const mustHave = ["BITRISEIO_GIT_REPOSITORY_OWNER", "BITRISEIO_GIT_REPOSITORY_SLUG"]
     const mustBeInts = ["BITRISE_PULL_REQUEST"]
     return ensureEnvKeysExist(this.env, mustHave) && ensureEnvKeysAreInt(this.env, mustBeInts)
-  }
-
-  private _parseRepoURL(): string {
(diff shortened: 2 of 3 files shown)


orta:
I'm not wild on this, because I'm not certain on whether this breaks older installs - 

/cc @BanyikAnna - I think bitrise is a pure PAAS, but you do you have enterprise installs etc, or does everyone have the same "version" of bitrise and thus everyone gets the same env vars on builds?

rogerluan:
This might certainly break users who manually override the designed env vars (in this case, `GIT_REPOSITORY_URL`), but users shouldn't be doing that in the first place I think, so I think it's safe? Both `GIT_REPOSITORY_URL` and the env vars I used are global bitrise env vars so they should be the same for everyone, independent of version - unlike step-specific env vars that can be modified depending on the version of the step you're using.

I'm not sure about Bitrise Enterprise plans though, they might have a different setup. Thanks for tagging someone with knowledge on this @orta ! 🤗

orta:
I'll give them a few days to respond, if we get nothing by next week I'll take it and treat it as a server minor

rogerluan:
Sounds perfect! Thank you @orta 💪

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 
 <!-- Your comment below this -->
 - Simplified Bitrise repo slug lookup, fixing SSH URL parsing in BitBucketServer - [@rogerluan]
+- Changed `CHANGELOG.md` merge strategy to union to avoid merge conflicts - [@rogerluan]
```
rogerluan, line 18: I added a changelog entry for https://github.com/danger/danger-js/pull/1116#issuecomment-781369623 lemme know if this should be reverted either because it's a meta change or doesn't belong to this PR 🤗

orta: I usually give the changelog notes a clean up before a release, it can stay (because it's not worth extra work undoing) but I treat the changelog as user-facing. Conceptually a user of danger shouldn't care too much about our merging strat in PRs.

rogerluan: Ah gotcha! Noted for future events ✅ thanks!! 💪

orta:
OK, let's do it

PR #1123, opened 2021-03-06
Add url to Danger local commits

f-meloni:
👋🏻 ,
Is been a while since my last PR here :)

I've added just a small change to be sure there is always a URL, also on the fake json returned by `danger local`.

Given we have it not nil on Danger-Swift I think this should be populated for all the platforms https://github.com/danger/swift/blob/master/Sources/Danger/GitDSL.swift#L57


f-meloni:
The CI failed, but should be related to the Danger Public token

Issue #1125, opened 2021-03-15
Support falling back to a GitHub App if there is no GITHUB_API_TOKEN  on OSS repos


orta:
**Describe the bug**

Switch the default experience to use the built-in GitHub App account, reducing the docs and setup for Danger and _more importantly_ making Danger + OSS repo make more sense.

orta:
Did a quick look at all of the bits necessary for this as they were already in for checks support (which still looks to suck) - I can make a GH app which they all can use, each person who wants to set it up needs to provide an installation ID for the install

The biggest blocker is whether it can create issue comments without write access to issues

PR #1126, opened 2021-03-17
Support using the Danger shared GitHub App for OSS repos to work around the forks issue

orta:
RE: #1125

### DIFF
--- a/.github/workflows/CI-via-OSS-app.yml
+++ b/.github/workflows/CI-via-OSS-app.yml
@@ -1,0 +1,20 @@
+name: CI
+on: pull_request
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      # Check out, and set up the node/ruby infra
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+
+      # Get local dependencies
+      - run: yarn install
+      - run: yarn build
+
+      # Run with OSS comments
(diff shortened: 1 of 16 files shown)


orta:
OK, confirmed that this isn't _reasonably_ possible as it is. The access to post a comment to the PR requires write access to any PR, and no-one _should_ accept that as a reasonable trade-off for some setup convenience. 

Perhaps using Checks is OK, but I've found every time I've used that system to be underwhelming.

orta:
My key argument against checks is that you lose the middle ground: danger gives message / warn / fail - but you can only really use _fail_ in a check because all feedback is hidden from you until you click though to the checks page.

You'd only click if you were blocked.

orta:
Summary: this idea generally works for most of the route

 - Setting up the auth token mostly invisibly (you'd need to follow the instructions to ge the install id but that's easy)
 - Reading all the meta from the PR looks fine with ^
 - Just leaving feedback blocked by the security model

orta:
Took a walk on this. 

The only way this could realistically work is by making some sort of web service, which I'm really not super wild on. The web service could link a repo/org to installation ID (maybe as a GH repo) which danger can look up at runtime. That would eliminate the need for `DANGER_GITHUB_APP_INSTALL_ID`.

Then I'd have to proxy GH API calls for editing/updating/creating comments to its own account.

Issue #977, opened 2020-01-24
Parse entire commit message not only first line


fuhlig:
Commit-Messages via `danger.git.commits` return only the first line and is parsed

- no body
- special chars (like `!`) are removed

What I am trying to achieve are warnings based on file changes in combination with commit guidelines (e.g. breaking changes with [conventional commits](https://www.conventionalcommits.org/en/)).

message:

danger returns:

I guess is how the message is processed (locally?)
https://github.com/danger/danger-js/blob/4100842c8922fa46e975ed772c1eb6ece677ff7e/source/platforms/git/localGetCommits.ts#L17

orta:
That part of the codebase is only used in `danger local`, for a normal `danger ci` or `danger pr` it would use the GitHub API: https://github.com/danger/danger-js/blob/4100842c8922fa46e975ed772c1eb6ece677ff7e/source/platforms/github/GitHubGit.ts#L19-L29 which comes from https://github.com/danger/danger-js/blob/4100842c8922fa46e975ed772c1eb6ece677ff7e/source/platforms/github/GitHubAPI.ts#L208-L212

A commit looks [like this](https://developer.github.com/v3/repos/commits/#get-a-single-commit) which doesn't seem too have anything a bit more raw for a full message?

fuhlig:
Is this consistent across all VCS?

I don't quite understand why a local commit is handled differently. Is there any difference between local and remote (on any VCS) commits? Honest question because that would be new to me

orta:
Everything in the Danger DSL is normally done via an API, with the exception of `danger local` which replicates one locally. 

Each API (github/gitlab/bitbucket etc) all implements their own Commit API -> `GitCommit` code

PR #1127, opened 2021-03-19
Use full unformatted subject message

denieler:
We want to use danger js locally to check the format of conventional commits for the commit messages. Due to that, we need to have a full unformatted subject line. Similar issue is explained in https://github.com/danger/danger-js/issues/977

### DIFF
--- a/source/runner/Executor.ts
+++ b/source/runner/Executor.ts
@@ -70,7 +70,7 @@ export interface ExecutorOptions {
 const isTests = typeof jest === "object"
 
 interface ExitCodeContainer {
-  exitCode: number
+  exitCode?: number
 }
 
 export class Executor {
--- a/package.json
+++ b/package.json
@@ -131,7 +131,7 @@
     "tslint": "^5.11.0",
     "tslint-config-prettier": "^1.15.0",
     "typedoc": "0.9.0",
-    "typescript": "^3.1.1",
+    "typescript": "^3.9.7",
     "typescript-json-schema": "^0.32.0"
   },
   "dependencies": {
@@ -144,6 +144,7 @@
     "fast-json-patch": "^3.0.0-1",
     "get-stdin": "^6.0.0",
     "gitlab": "^10.0.1",
+    "gitlog": "^4.0.4",
     "http-proxy-agent": "^2.1.0",
     "https-proxy-agent": "^2.2.1",
     "hyperlinker": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -144,7 +144,6 @@
     "fast-json-patch": "^3.0.0-1",
     "get-stdin": "^6.0.0",
     "gitlab": "^10.0.1",
-    "gitlog": "^4.0.4",
     "http-proxy-agent": "^2.1.0",
     "https-proxy-agent": "^2.2.1",
     "hyperlinker": "^1.0.0",
--- a/source/platforms/git/_tests/localGetCommits.test.ts
+++ b/source/platforms/git/_tests/localGetCommits.test.ts
@@ -1,6 +1,5 @@
-import gitlog from "gitlog"
-
 import { localGetCommits } from "../localGetCommits"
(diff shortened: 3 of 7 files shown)


orta:
This will probably need additional work - message is used to _ensure_ it is a single line because it is serialized to JSON right after.

denieler:
hey @orta 👋 

> This will probably need additional work - message is used to ensure it is a single line because it is serialized to JSON right after.

Tested the change and it looks like it works fine. Change `%f` to `%s` only changes formatting but left only a single line of the commit message (https://git-scm.com/docs/pretty-formats#Documentation/pretty-formats.txt-emsem, it takes only `subject`). 

Screenshot 2021-03-19 at 12 13 10

I've tried to find a good place to add tests for this functionality but didn't find, also it looks a bit tricky to test `git log` output 😞  if you have any suggestions - please share.

And travis pr check is failing with some "Bad credentials message". Not sure that this change could cause it 😞 
Screenshot 2021-03-19 at 12 46 44

Please help 😃

denieler:
hey @orta 👋 any news? can we merge this PR?

orta:
Remember I do this in my spare time.

Can you verify that having `"` in a commit message won't break with this change? You can ignore the red CI from travis, I still need to get that fixed

denieler:
@orta sure, checking...

denieler:
hey @orta! It appeared that previous implementation does not respect `"` or any other quotes 😃 After some research I found that it's fairly hard to do that via `git log` which was there before and requires some extra work, so found `gitlog` package - it provides what's necessary. Just had to shape it to the current usage 👍 Looks like now working fine. Please check.

P.S. Had to update typescript, because the new library using new types, which are supported in TS 3.8, but in Danger it was super outdated (3.1.1). And another change was because of that. But looks like it doesn't break anything.

source/runner/Executor.ts
```diff
@@ -70,7 +70,7 @@ export interface ExecutorOptions {
 const isTests = typeof jest === "object"
 
 interface ExitCodeContainer {
-  exitCode: number
+  exitCode?: number
```
denieler, line 73: because TS was updated it started to complain about 

Screenshot 2021-03-22 at 17 51 36

So had to fix this type to fix it

denieler:
hey @orta ! Can you please review this PR? 🙇

orta:
I'm not a fan of adding new dependencies to Danger JS and I don't think this is worth it. 

I'll take reverting back to https://github.com/danger/danger-js/pull/1127/commits/6a179366598260c1796056b6df57a285494040a6 and accept that 👍🏻

denieler:
@orta what do you think if I copy partially the part from the `gitlog` package into dangerjs code? Do you think this will be a better solution?

orta:
Maybe, kinda depends on how much code it is - given the goal is to import a string, switching `"` -> "▍" (some random char) before putting it into the JSON and flipping it on the other side would also solve the problem too.

denieler:
> Maybe, kinda depends on how much code it is - given the goal is to import a string, switching `"` -> "▍" (some random char) before putting it into the JSON and flipping it on the other side would also solve the problem too.

There are few characters, which can affect `JSON.parse`, so I didn't really want to move this way. Ok, let me make a change and I will ping you one more time. If you don't like it - we can go with `string.replace` for a bunch of  special characters

denieler:
@orta please take a look at the new change. I've not rolled back the typescript update, because it looks for me like a healthy dependency update for the library and it looks like not breaking anything

orta:
Looks great to me, let's get this shipped

glensc:
the included [gitlog](https://www.npmjs.com/package/gitlog) is BSD-3-Clause licensed, meaning you must preserve any copyright notices, which is not done here.

read the license here:
- https://opensource.org/licenses/BSD-3-Clause


the file in question:
- https://github.com/denieler/danger-js/blob/040d5a374f74e5cf185d182397570e539adbbc18/source/platforms/git/localLogGitCommits.ts

PR #1132, opened 2021-03-31
Fix publishing Docker image to GitHub

markuspoerschke:
The workflow definition was copied from https://docs.github.com/en/actions/guides/publishing-docker-images#publishing-images-to-github-packages.

No changes were needed to publish the docker image to GitHub's container registry.

You can find a successfull run here: https://github.com/markuspoerschke/danger-js/runs/2240530067?check_suite_focus=true (since you can test it only once you released a new version...)


markuspoerschke:
Fixes #923

f-meloni:
Is this action using the org's package registry or is it deploying the package directly on the danger-js repo's package?
We moved the various docker images all under the Danger org's package registry (e.g., https://github.com/orgs/danger/packages/container/package/danger-swift), that offers a better handling of the Docker Packages, and should be the suggested one.

.github/workflows/uploadDockerImage.yml
```diff
@@ -1,30 +1,19 @@
...
-        USERNAME: ${{ secrets.DOCKER_GITHUB_USERNAME }}
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Push to GitHub Packages
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+          registry: docker.pkg.github.com
+          repository: danger/danger-js/danger-danger-js
```
f-meloni, line 18: I think this should be 

```suggestion
          registry: ghcr.io/danger
          repository: danger-js
```

But I'm not familiar with this action

markuspoerschke:
# Docker Image Build Process at Danger Systems

## Problem

The Danger JS is currently not publishing an up to date Docker Image. To stay compliant with the other Danger projects, the build process for Docker Images should be according to the other projects.

## Objective

Danger JS should provide an up to date Docker Image. This analysis should defined

* to where the Docker Image is published

* how the build process looks like

## Related Danger Projects

The following Danger projects are considered in this analysis:

* [Danger JS](https://github.com/danger/danger-js) (this project)

* [Danger Swift](https://github.com/danger/swift)

* [Danger Ruby](https://github.com/danger/danger) (the original project)

* [Danger Kotlin](https://github.com/danger/kotlin)

* [Danger Python](https://github.com/danger/python)

## Analysis

|            | Published to                                              | Build process                                                                                          | Remarks                                     |
| ---------- |:--------------------------------------------------------- |:------------------------------------------------------------------------------------------------------ | ------------------------------------------- |
| **JS**     | `docker.pkg.github.com/danger/danger-js/danger-danger-js` | [GitHub Action](https://github.com/danger/danger-js/blob/main/.github/workflows/uploadDockerImage.yml) | Build Process is broken                     |
| **Swift**  | `ghcr.io/danger/danger-swift`                             | [GitHub Action](https://github.com/danger/swift/blob/master/.github/workflows/publish_package.yml)     |                                             |
| **Ruby**   | `hub.docker.com/r/dangersystems/danger-rb`                | [Docker Hub](https://hub.docker.com/r/dangersystems/danger-rb)                                         | Only latest development build, no tags      |
| **Kotlin** | `ghcr.io/danger/danger-kotlin`                            | [GitHub Action](https://github.com/danger/kotlin/blob/master/.github/workflows/publish_package.yml)    |                                             |
| **Python** | ?                                                         | ?                                                                                                      | I could not find any published Docker Image |

## Conclusion

The implementations for Swift and Kotlin are providing a working Docker Image. They are using the [recommended ](https://docs.github.com/en/packages/guides/migrating-to-github-container-registry-for-docker-images)GitHub Container Registry for Docker Images.

Danger JS should use the same mechanism to build and publish like Kotlin and Swift.

The Build script for Docker Images can be copied from Kotlin. It was most recently edited by @f-meloni who might be able to help setting up the correct secrets.

orta:
Seems reasonable to me, doesn't look like it changes my deploy flow at all 👍🏻

f-meloni (approved):
Looks good to me.
@orta if you want I can add my secrets here as well, like I did on the other danger.

orta:
Sure, do it 👍🏻

shyim:
What's missing here? I can help when needed :)

f-meloni:
>Sure, do it 👍🏻

I've added the secrets, so we shoudl be fine

PR #1140, opened 2021-05-11
Add thread support for GitLab

shyim:
That would be the way I would solve the thread problem.

After your Okay @orta I can implement also Github too :)

### DIFF
--- a/source/platforms/BitBucketCloud.ts
+++ b/source/platforms/BitBucketCloud.ts
@@ -67,6 +67,14 @@ export class BitBucketCloud implements Platform {
     return true
   }
 
+  supportsThreads() {
+    return false;
+  }
+
+  updateOrCreateThread = async (): Promise<undefined> => {
+    return undefined;
+  }
+
   /**
    * Returns the response for the new comment
    *
--- a/source/platforms/GitLab.ts
+++ b/source/platforms/GitLab.ts
@@ -2,7 +2,7 @@ import GitLabAPI from "./gitlab/GitLabAPI"
 import { Platform, Comment } from "./platform"
 import { GitDSL, GitJSONDSL } from "../dsl/GitDSL"
 import { GitCommit } from "../dsl/Commit"
-import { GitLabDSL, GitLabJSONDSL, GitLabNote } from "../dsl/GitLabDSL"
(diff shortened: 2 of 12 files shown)


JoshuaBehrens:
Any activities here? I could need that one :O

source/platforms/GitLab.ts
```diff
@@ -178,6 +197,17 @@ class GitLab implements Platform {
     )
   }
 
+  getDangerThreads = async (dangerID: string) => {
+    const { id: dangerUserId } = await this.api.getUser()
+    const threads = await this.api.getMergeRequestThreads();
```
unfernandito, line 202: This could be resolved with Promise.all

```suggestion
    const [{ id: dangerUserId }, threads] = await Promise.all([this.api.getUser(), this.api.getMergeRequestThreads()])
```

source/platforms/GitLab.ts
```diff
@@ -92,6 +92,23 @@ class GitLab implements Platform {
...
+  updateOrCreateThread = async (dangerID: string, newComment: string): Promise<string> => {
+    const threads = await this.getDangerThreads(dangerID);
+
+    if (threads.length) {
+      await this.api.updateMergeRequestDiscussion(threads[0].id, threads[0].notes[0].id, newComment);
+      return `${this.api.mergeRequestURL}#note_${threads[0].notes[0].id}`;
+    } else {
+      const thread = await this.api.createMergeRequestDiscussion(newComment);
+
+      return `${this.api.mergeRequestURL}#note_${thread.notes[0].id}`;
```
unfernandito, line 108: Maybe the zero could be a const with a description. Looks like a magic number without any type of context of why the code should the zero index and not any other

source/platforms/BitBucketCloud.ts
```diff
@@ -67,6 +67,14 @@ export class BitBucketCloud implements Platform {
     return true
   }
 
+  supportsThreads() {
```
unfernandito, line 70: the type is infered?

unfernandito (changes requested):

purefan:
pinging @shyim I hope you dont abandon this PR, its good work and I think the review comments are valid

shyim:
As it took me too long to have it. I just built something similar of Danger for our company. 

Free free to create a new PR based on this :)

uncaught:
@orta I suppose this is superseeded by #1331 and could be dropped?

Issue #1134, opened 2021-04-09
[BUG] Multiple inline comments don't work


hlynurstef:
**Describe the bug**
When you try to post multiple inline comments, only the first comment is posted while the others fail (but still appear in the overview comment).

**To Reproduce**
1. Create a danger rule that warn/fail/message on a specific file and line number on several files that are included in the diff on GitHub
2. Run `yarn danger ci` from your CI
3. Notice that only the first inline comment is posted while the others fail

**Expected behavior**
Danger should post all of the inline comments requested.

** Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          | 10.6.3
| node             | 
| npm              |
| Operating System | macOS 10.15.7 (Catalina)

**Additional context**
When running danger with the following command: `DEBUG={*} yarn danger ci` I see that there seem to be two requests made to the same endpoint at almost the same time (maybe related to the error?) and then we can see that the request failed with an error from the GitHub API:

```
2021-04-09T01:14:10.956Z danger:GitHubAPI Sending:  https://api.github.com/repos/driftline-analytics/runmaker/pulls/43/comments {
  'Content-Type': 'application/json',
  Authorization: 'token [REDACTED]'
}
2021-04-09T01:14:10.959Z danger:GitHubAPI Sending:  https://api.github.com/repos/driftline-analytics/runmaker/pulls/43/comments {
  'Content-Type': 'application/json',
  Authorization: 'token [REDACTED]'
}
Request failed [422]: https://api.github.com/repos/driftline-analytics/runmaker/pulls/43/comments
Response: {
  "message": "Validation Failed",
  "errors": [
    {
      "resource": "PullRequestReviewComment",
      "code": "custom",
      "field": "pull_request_review_id",
      "message": "pull_request_review_id must be pending"
    }
  ],
  "documentation_url": "https://docs.github.com/rest/reference/pulls#create-a-review-comment-for-a-pull-request"
}
```

This seems related to https://github.com/danger/swift/issues/301

orta:
Cool, yeah, maybe the API usage rules changed this since this was originally built out. I don't use inline messages, so I'm unlikely to fix it - but I'd accept PRs 👍🏻

hlynurstef:
Seems like the multiple inline comments have started working correctly all of a sudden, I haven't updated any dependencies so I'm guessing something might have changed serverside from GitHub?

Since this has started working correctly I'm going to close this issue.

hlynurstef:
[...]

PR #1176, opened 2021-11-19
Fix: [Github] Multiple Inline Comments on the same file/line should all be posted

Rouby:
This PR should address #1134 by sending all inline comments as a review combined.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 
 <!-- Your comment below this -->
 
+- Fix: [Github] Multiple Inline Comments on the same file/line should all be posted [#1176](https://github.com/danger/danger-js/pull/1176) [@Rouby]
+
 <!-- Your comment above this -->
 
 # 10.7.1
--- a/source/platforms/FakePlatform.ts
+++ b/source/platforms/FakePlatform.ts
@@ -85,4 +85,9 @@ export class FakePlatform implements Platform {
   }
 
   getFileContents = (path: string) => new Promise<string>(res => res(readFileSync(path, "utf8")))
+
+  createInlineReview?: (
+    git: GitDSL,
+    comments: { comment: string; path: string; line: number }[]
+  ) => Promise<any> = undefined
 }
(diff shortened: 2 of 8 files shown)


orta:
Structurally this looks good 👍🏻  - I'd like a test or two though because I don't want this breaking from other contributors changes in the future

Rouby:
> a test or two
Will do :)

fbartho (approved):
This PR looks good to me! Just need to add an entry to the Changelog!

Something like the following inside "Your comment below here":

```markdown
- Fix: [Github] Multiple Inline Comments on the same file/line should all be posted [#1176](https://github.com/danger/danger-js/pull/1176) [@Rouby]
```

Rouby:
> This PR looks good to me! Just need to add an entry to the Changelog!
> 
> Something like the following inside "Your comment below here":
> 
> ```gfm
> - Fix: [Github] Multiple Inline Comments on the same file/line should all be posted [#1176](https://github.com/danger/danger-js/pull/1176) [@Rouby]
> ```

Thanks and done 👍

fbartho:
Note: CI is blocked by #1189 or an alternative solution there (so don’t be alarmed by the test failure)

f-meloni:
Would make sense to have the new comment logic enabled when there is a specific flag (e.g.,` --use-github-reviews`) or something like that, so that is disabled by default and we can then investigate it with calm (but Danger keeps working as expected in the meantime)?

orta:
Yeah, perhaps I misinterpreted this PR, IMO warnings/fails/etc which cannot be registered in the diff should show up in the main comment. I personally find reviews annoying in PRs (because they don't get overwritten by the edits and thus spam the PR on every CI run )

I think this behavior as is could be a flag but we should get it aligned with how danger ruby does is which is ^

f-meloni:
Yeah, mine was more a temporary suggestion.
Failed tests or lint issues might be affected by this change, so I thought we could keep those working while this gets solved

f-meloni:
PS. I just realised I posted this in the PR and not in the issue 😅 sorry

PR #1188, opened 2022-01-06
Release 10.8.0

fbartho:
I don't have permission to push directly to the `main` as specified in the [Contributing Guidelines](https://github.com/danger/danger-js/blob/main/CONTRIBUTING.md#releasing-a-new-version-of-danger)

This means I can neither update the Changelog directly, nor can I tag the release. Sorry for making work for you @orta!

PR #1186, opened 2021-12-26
fix: mark `body` as nullable for Github PRs

G-Rath:
Resolves #1185

PR #1189, opened 2022-01-06
[Maintenance] Switch to node-14 as the oldest tested environment

fbartho:
`create-react-app` requires >= node-14, and our test-CI workflow requires create-react-app to succeed. See the CI failure in #1188 

This PR upgrades all the node-versions that I could find from 10, 10.3, or unspecified, to node 14.

## Alternatives

A. only upgrade the CI for that particular GitHub action workflow
B. upgrade to node-16 (14 has already been in maintenance mode for a long time)
C. upgrade to node-17 
D. upgrade to node-18 (18 won't enter LTS until October)
E. Delete the create-react-app test case

[Node Version Chart](https://nodejs.org/en/about/releases/)

We could also add an entry to our package.json `engines` to express whichever decision we make.

## Blocking

This PR Blocks: #1176, #1186, #1188, #1190, and any future PRs from going green in CI

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,4 @@
 {
-  "presets": [["@babel/preset-env", { "targets": { "node": "6" }, "useBuiltIns": "usage" }]],
+  "presets": [["@babel/preset-env", { "targets": { "node": "14" }, "useBuiltIns": "usage" }]],
   "plugins": ["@babel/plugin-transform-flow-strip-types"]
 }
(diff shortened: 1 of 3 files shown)


.babelrc
```diff
@@ -1,4 +1,4 @@
 {
-  "presets": [["@babel/preset-env", { "targets": { "node": "6" }, "useBuiltIns": "usage" }]],
+  "presets": [["@babel/preset-env", { "targets": { "node": "14" }, "useBuiltIns": "usage" }]],
```
orta, line 2: This is a major semver change in the JS emit, and it's not worth it for that

```suggestion
  "presets": [["@babel/preset-env", { "targets": { "node": "6" }, "useBuiltIns": "usage" }]],
```

orta:
Looks reasonable to me

Issue #1181, opened 2021-12-08
[BUG] Github/Git JSON DSL Parse Error -- when moving Folders/Files


fbartho:
**Describe the bug**
A refactor in a Github-based project that moved some directories around caused DangerJS to stop reporting errors, with the following output:

```log
Reached Bottom of Dangerfile // custom log line we injected

Unable to evaluate the Dangerfile
 TypeError: Cannot read properties of null (reading 'name')
    at Object.get (/home/runner/work/backend/backend/node_modules/jsonpointer/jsonpointer.js:69:14)
    at /home/runner/work/backend/backend/node_modules/danger/distribution/platforms/git/gitJSONToGitDSL.js:143:63
    at Array.reduce (<anonymous>)
    at /home/runner/work/backend/backend/node_modules/danger/distribution/platforms/git/gitJSONToGitDSL.js:135:48
    at step (/home/runner/work/backend/backend/node_modules/danger/distribution/platforms/git/gitJSONToGitDSL.js:52:23)
    at Object.next (/home/runner/work/backend/backend/node_modules/danger/distribution/platforms/git/gitJSONToGitDSL.js:33:53)
    at fulfilled (/home/runner/work/backend/backend/node_modules/danger/distribution/platforms/git/gitJSONToGitDSL.js:24:58)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)


Failing the build, there is 1 fail.
[Additional log output about updating/posting the comment]
```

**To Reproduce**
Unclear, we don't have a minimal reproduction set. I'm happy to inject any debugging code or flags that you deem helpful!

**Expected behavior**
No error thrown.

**Screenshots**
N/A

**Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          | 10.7.1
| node             | 16
| yarn              | 1.22.17
| Operating System | Github Actions `ubuntu-latest`

**Additional context**
N/A

fbartho:
Update: we were able to replicate this issue with another PR that moved files.

The files were moved into a newly created nested directory. Many of them had no other changes, but some of them just had paths updated.

This still isn't a minimal reproduction set, but maybe there's a hint at what JSON file jsonpointer is choking on?

orta:
It's likely related to this https://github.com/danger/danger-js/issues/481 - there's something strange with the git output we get from renames

Issue #481, opened 2018-01-16
Renamed files appear within modified_files, but fail when passed to danger.git.diffForFile


neil-morrison44:
I've got a file within my repo which has had a parent folder renamed 
(e.g. from `/tests/seach/index.spec.js` to `/tests/search/index.spec.js`)

I'm running some tests on the diff of modified files, and this renamed (but not modified) file is turning up within `danger.git.modified_files`.

It appears that when I run `danger.git.diffForFile` on the file's path, danger fires off 1 request for the branch file (which succeeds) and 1 request for the master file using the new path (which fails, 404).

I'd expect Danger to either:
- Know that the file has been renamed but not modified, not include it within `danger.git.modified_files`
- If that's not possible then include it within `danger.git.created_files`
- Handle renames, request the old version using the old name (if possible?)
- More graceful handling / error reporting of 404s during the requests for diffs

orta:
I agree, that logic should be pretty feasible to look up using the existing calls we have (e.g. inside the diff we get) then to use that instead of simply looking for the same named file twice

lawrence-berry:
I've also ran into the issues described above with renamed folders, and agree that it shouldn't be included in `modified_files` and including it in `created_files` would make more sense if it's not possible to recognise these files as having been moved rather than modified.

heatherbooker:
On a related note, when I have a new file added and try to run `danger local`, `diffForFile` fails as it tries to run ` 'git show master:"<my_new_file.ext>"'`. It says:
```
Path '<my_new_file.ext>' exists on disk, but not in 'master'.
```
But if I change it to `JSONPatchForFile`, it reports the diff with no errors. So I think it might extend to `created_files` as well as renamed ones.

heatherbooker:
I spoke too soon - it reports an empty diff if I use `JSONPatchForFile`, but at least it doesn't crash. ><

mAAdhaTTah:
Experiencing this, although I'm not sure it's a major issue. How can we tell that a file has been renamed?

airtonix:
This hasn't been fixed yet, although #1197 seems to think it has.

airtonix:
for now we're having to preempt failing cases :( 

```
const invalidFiles = (
    await Promise.all(
      changedPackageJsons.map(async (filename) => {
        try {
          await execasync(`git show master:"${filename}"`);
        } catch (error) {
          return {};
        }
        const diff = await danger.git.JSONDiffForFile(filename);
        return {
          filename,
[...]

PR #1197, opened 2022-01-14
Fix: [git] JSONDiffForFile when passed the path of a JSON File that was moved, crashes

fbartho:
Don't crash when danger.git.JSONDiffForFile encounters a moved JSON file

Notably affects danger-plugin-yarn; when a package.json is moved, it slips by a test in `JSONPatchForFile` that filters for `danger.git.modified_files`, and I couldn't figure out how to fix `getFileContents` so it could look up the old file's contents using the new file path.

_Note: I also couldn't figure out how to generate the right fixtures, or how to mock `JSONPatchForFile`_

Fixes: #1181
Related: #481

<details><summary>Some useful test data</summary>

```js
const diff = {
  before: null,
  after: {
    name: '@fbartho/something',
    description: 'this is irrelevant',
    version: '1.0.0',
    private: true,
    main: './dist/main.js',
    dependencies: { express: '^4.17.1' }
  },
}

const outerDiff = [
    { op: 'add', path: '/name', value: '@fbartho/something' },
    {
      op: 'add',
      path: '/description',
      value: 'this is irrelevant'
    },
    { op: 'add', path: '/version', value: '1.0.0' },
    { op: 'add', path: '/private', value: true },
    { op: 'add', path: '/main', value: './dist/main.js' },
    { op: 'add', path: '/dependencies', value: [Object] },
  ]
```

</details>


source/platforms/git/gitJSONToGitDSL.ts
```diff
@@ -110,16 +113,19 @@ export const gitJSONToGitDSL = (gitJSONRep: GitJSONDSL, config: GitJSONToGitDSLC
...
 
-    const { diff, before, after } = patchObject
-    return diff.reduce((accumulator, { path }) => {
+    const { diff: outerDiff, before, after } = patchObject
+    return outerDiff.reduce((accumulator, { path }) => {
       // We don't want to show the last root object, as these tend to just go directly
       // to a single value in the patch. This is fine, but not useful when showing a before/after
       const pathSteps = path.split("/")
       const backAStepPath = pathSteps.length <= 2 ? path : pathSteps.slice(0, pathSteps.length - 1).join("/")
+      console.debug("MEEP", { path, pathSteps, backAStepPath, before, after, outerDiff })
```
orta, line 122: Drop this and it's good to merge 👍🏻

orta (approved):

source/platforms/git/gitJSONToGitDSL.ts
fbartho:
```suggestion
```

oops!!

PR #1204, opened 2022-02-01
Upgrade @octokit/rest from v16.43.1 to v18.12.0 - Fixes #1202

### LINKED ISSUE #1202: [Maintenance][Github] @octokit/rest is on version 18, but DangerJS is on 16
**Describe the bug**
@octokit/rest is on version 18.12.0, but DangerJS is referencing Octokit 16

**Expected behavior**
For maintenance reasons, we should upgrade DangerJS to the 18.x API line, but this is a breaking change for DangerJS consumers, and potentially for plugins

@orta is this something we're ready to do, or do we need to hold off for some reason?

fbartho:
Fixes #1202 

This is blocked by: https://github.com/orta/memfs-or-file-map-to-github-branch/pull/5

This PR incidentally upgraded TypeScript, which is also a breaking change (because `.d.ts` files could emit differently), but PR doesn't take advantage of the new TS features because husky's lint-staged step is still configured to use TSLint, and tslint doesn't understand `import…type` and optional-chaining `?.` / null-coalescing `⁇`, so it was blowing up my commits.


fbartho:
Note: the test failure is expected until: https://github.com/orta/memfs-or-file-map-to-github-branch/pull/5 is merged and deployed!

source/runner/jsonToDSL.ts
```diff
@@ -85,7 +85,7 @@ const apiForDSL = (dsl: DangerDSLJSONType): Octokit | BitBucketServerAPI | GitLa
     return new GitLabAPI(gitlab.metadata, getGitLabAPICredentialsFromEnv(process.env))
   }
 
-  const options: Octokit.Options & { debug: boolean } = {
+  const options: ConstructorParameters<typeof Octokit>[0] & { debug: boolean } = {
```
orta, line 88: Nice work

orta:
dts emit seems fine, but I think this is probably a semver major because we expose the octokit API object and if we had to make changes in our API usage there's a very good chance others will too

PR #1205, opened 2022-02-01
chore: Switch from tslint to eslint

fbartho:
tslint is end-of-life, and warns when you install it. Additionally, the version we were on didn't support 'import type' expressions, as well as optional-chaining & null coalescing when I tried to use those in #1204

This PR does update a bunch of tslint-disable comments, but otherwise tries to minimally change runtime source in danger-js. If we want to tighten the eslint rules, I'd be super supportive, but I didn't want to cause a ton of thrash, so mostly the rules that are enabled are the ones that don't trigger on tons of existing code.

This PR also upgrades prettier, because that was needed to handle `readonly string[]` arguments.

**Be aware** this PR drops testing support for node 10 in AppVeyor -- https://github.com/danger/danger-js/pull/1205/commits/b66f86f5e38eb46adbb3b4a8ab945bbc45265816 [`@typescript-eslint` doesn't support node 10](https://ci.appveyor.com/project/orta/danger-js/builds/42418823)

### DIFF
--- a/source/platforms/bitbucket_cloud/_tests_/_bitbucket_cloud_git.test.ts
+++ b/source/platforms/bitbucket_cloud/_tests_/_bitbucket_cloud_git.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable jest/no-export */
 import { bitBucketCloudRawAndDateToGitCommitAuthor } from "../BitBucketCloudGit"
 
 import { BitBucketCloud } from "../../BitBucketCloud"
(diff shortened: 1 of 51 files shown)


source/platforms/bitbucket_cloud/_tests_/_bitbucket_cloud_git.test.ts
```diff
@@ -162,11 +163,14 @@ describe("the dangerfile gitDSL - BitBucket Cloud", () => {
...
-
-    writeFileSync(pathJoin(fixtures, "bbc-dsl-input.json"), JSON.stringify(dataSent, null, "  "), "utf8")
+    expect.assertions(1)
+    expect(async () => {
+      const fakeSource = new FakeCI({})
+      const dataSent = await jsonDSLGenerator(bbc, fakeSource, {} as any)
+      dataSent.settings.github.accessToken = "12345"
+
+      writeFileSync(pathJoin(fixtures, "bbc-dsl-input.json"), JSON.stringify(dataSent, null, "  "), "utf8")
+    }).not.toThrow()
```
orta, line 173: 👍🏻 nice work

orta:
Perfect PR. 

WRT to tightening the linter rules, call me lightly supportive - I mostly disable eslint/tslint on repos I run nowadays because 'I know better' (aka, since I joined TS I mostly only write repos where I'm the core author and so I don't have to share well with others) so you're welcome to do it (in part because you're writing most new code in danger ATM) but know that I'm a sledgehammer who might just turn it off for sections I write :D

orta:
I couldn't find an email for you - but the danger slack seems to have got a bit more active over the last week or two, so you might enjoy having access too ( send me a "hey" email to the mail in my GitHub bio @orta)

PR #1251, opened 2022-03-15
Print default base branch for danger local

glensc:
This uses commander default argument value.


glensc:
I don't know how to test that commandline change, couldn't find anything useful in CONTRIBUTING.md

orta:
Seems like a reasonable change to me 👍🏻 thanks

We don't really have any tests at the CLI level, given that it's usually just a quick jump into app code

orta:
I don't really get on the npm rat race for dependencies unless there is a real CVE which I think is important to update, you're welcome to take a look at updating

PR #1252, opened 2022-03-16
Specify engine required by this package

glensc:
Needed to set baseline when updating dependencies:
- https://github.com/danger/danger-js/pull/1251#issuecomment-1068210929

The version was chosen based on `@types/node` in `package.json`

PR #1254, opened 2022-03-16
Use npm-run-all for danger:prepush to run scripts in sequence

glensc:
Noticed that "danger:prepush" ran commands with ";" rather "&&" meaning the next command is still executed if the previous one failed. This resulted in wery weird errors due to the previous step failure.

This PR uses "npm-run-all" module for platform agnostic and clean API support:
- https://www.npmjs.com/package/npm-run-all

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,8 @@
     "docs:cp_defs": "mkdir docs/docs_generate; cp source/danger.d.ts docs/docs_generate; cp node_modules/@octokit/rest/index.d.ts docs/docs_generate/github.d.ts",
     "docs": "yarn run docs:cp_defs; yarn typedoc --ignoreCompilerErrors --mode modules --json docs/js_ref_dsl_docs.json --includeDeclarations source",
     "dts-lint": "yarn run declarations && yarn dtslint types",
-    "danger:prepush": "yarn build:fast; yarn test:fixtures; node distribution/commands/danger.js local --base main --dangerfile dangerfile.lite.ts"
+    "danger:local": "node distribution/commands/danger.js local --base main --dangerfile dangerfile.lite.ts",
+    "danger:prepush": "npm-run-all build:fast test:fixtures danger:local"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


fbartho (changes requested):
I commented in #1252 not realizing this PR existed.

Sorry, I’d prefer using the `&&` syntax instead of pulling a new dependency for this purpose. 
This change adds extra maintenance burden without enough benefit in my opinion.

glensc:
I find "&&" less readable; also it doesn't work on all platforms. Up to maintainers of course what platforms they want to support for development.

npm-run-all also has "build:*" syntax to automatically run all jobs matching, but then the ordering would be incorrect.

the other way to solve this is with dependencies. i.e each script can have "pre" script, i.e if you run "build:bla" then "prebuild:bla" is checked and ran. 

I'm not sure if this is yarn thing, or npm does it too.

fbartho:
I do understand that some shell syntaxes are maybe foreign for some people. That said, using `&&` is a classic shell rule so shouldn’t be considered “uncommon” or “strange” if you’re writing a CLI command-string. I have no idea what OS doesn’t support this, but we haven’t had any complaints or Feature requests related to such a thing.

This repo has a small-to-medium amount of shell commands, and this PR is only “helping” one of those commands. As a result, I’m just not convinced that the benefit outweighs the maintenance burden of a new dependency.

Additionally, looking at the tool’s pages on npmJS and Github, it hasn’t had a release in 3 years, it “supports” `node ^4.9.1`. Adopting this dependency would immediately cause maintenance problems. That alone is enough to disqualify this tool even if we had a significant need for it.

glensc:
What I've heard is that windows platforms don't do POSIX shell syntax. I have no way to verify this.

but the original problem is that the prepush pipeline doesn't abort on the first error, shall I provide PR with using `&&` then?

fbartho:
I would be happy to review/approve a PR that solves it without adding a dependency if you want to contribute it! Thanks for understanding @glensc!

glensc:
Replacement: https://github.com/danger/danger-js/pull/1258

PR #1257, opened 2022-03-17
Bump node engine to 14

glensc:
This reflects what is running on CI. and also some of the current dependences already require that version.

Using `14.13.1` for `lint-staged ^12`:
- https://github.com/okonet/lint-staged#v12

See:
- https://github.com/danger/danger-js/pull/1260

See https://github.com/danger/danger-js/pull/1252#issuecomment-1071070403

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 # Test against this version of Node.js
 environment:
-  nodejs_version: "12"
+  nodejs_version: "14"
 
 # Install scripts. (runs after repo cloning)
 install:
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
   },
   "homepage": "https://github.com/danger/danger-js#readme",
   "engines": {
-    "node": ">=10"
+    "node": ">=14.13.1"
   },
   "devDependencies": {
     "@babel/cli": "7.1.2",


fbartho (approved):

fbartho:
Looks like appveyor’a matrix might need an update too

fbartho:
https://github.com/danger/danger-js/blob/9337fac93a2ea1855f7dd68eb6e2cab0eff4843b/appveyor.yml#L3

glensc:
Looks like we ran into chicken-egg problem. I wanted to bump node to 14 to be able to finish commander.js update, but updating `@types/node` breaks commander integration:

```
$ tsc -p tsconfig.production.json
node_modules/commander/typings/index.d.ts:26:25 - error TS2689: Cannot extend an interface 'NodeJS.EventEmitter'. Did you mean 'implements'?

26   class Command extends NodeJS.EventEmitter {
                           ~~~~~~

source/platforms/github/GitHubUtils.ts:92:22 - error TS2769: No overload matches this call.
```

I propose we move on here without `@types/node` bump, and do that once dependencies are updated, so the bumping can be done.

glensc:
I've posted `@types/node` pr separately:
- https://github.com/danger/danger-js/pull/1259

glensc:
Who will do the merge now? I see one approval, but it was set before further commits.

image

and @fbartho you didn't remove your approval after further comments.

orta:
PR looks good to me, and there's no CLA on people's spare time on OSS maintenance

orta:
Then no, please do not merge yourself, I + @fbartho  generally merge most things - the ability to self merge is a system designed to allow for folks to worry less about their dependencies and handle absent maintainers (or folks looking to own sections of the codebase after some time)

glensc:
You probably don't need to run danger for all node versions. you are most likely checking the static code committed to repo, so it doesn't matter what node version was used to run danger.js.

But staying on low versions is not a good option either. the dependencies can't be updated and the dependencies will be reported full of security vulnerabilities.

Maintaining multiple version series probably won't going to work either due lack of manpower.

my 2ç

fbartho:
@falkenhawk you're welcome to stay behind on an older version of DangerJS, the older versions are the ones that are still compatible with Node12.

NodeJS itself no longer supports Node12, and new features for DangerJS can't maintain a wider support range than that, it's just not practical. https://github.com/nodejs/release#release-schedule

What's your real concern here @falkenhawk? You're trying to upgrade to latest DangerJS, but not upgrade to latest NodeJS? 

It's common when somebody is testing a special edge case, to use a pattern like [patch-package](https://www.npmjs.com/package/patch-package) to disable the node engine check on this package. We obviously can't provide support if you're patching things like that, but if you think your use-case would still be fine, you're welcome to remove that Engine check in your code-bases.

PR #1259, opened 2022-03-18
Update @types/node to ^14

glensc:
Requires commander update first:
- https://github.com/danger/danger-js/pull/1257#issuecomment-1072126840

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "@types/lodash.memoize": "^4.1.3",
     "@types/micromatch": "^3.1.0",
     "@types/nock": "^10.0.3",
-    "@types/node": "^10.11.3",
+    "@types/node": "^14",
     "@types/node-fetch": "^2.5.12",
     "@types/p-limit": "^2.0.0",
     "@types/prettier": "^1.16.1",
--- a/source/api/_tests/fetch.test.ts
+++ b/source/api/_tests/fetch.test.ts
@@ -50,9 +50,9 @@ class TestProxy {
   private server = http.createServer(this.router)
 
   start = async (): Promise<void> => {
-    return new Promise<void>((resolve, reject) => {
+    return new Promise<void>(() => {
       this.isRunning = true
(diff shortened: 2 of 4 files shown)


glensc:
Resolving this error:

```
source/platforms/github/GitHubUtils.ts:92:22 - error TS2769: No overload matches this call.
  Overload 1 of 4, '(arrayBuffer: WithImplicitCoercion<ArrayBuffer | SharedArrayBuffer>, byteOffset?: number | undefined, length?: number | undefined): Buffer', gave the following error.
    Argument of type 'string' is not assignable to parameter of type 'WithImplicitCoercion<ArrayBuffer | SharedArrayBuffer>'.
  Overload 2 of 4, '(str: WithImplicitCoercion<string> | { [Symbol.toPrimitive](hint: "string"): string; }, encoding?: BufferEncoding | undefined): Buffer', gave the following error.
    Argument of type 'string' is not assignable to parameter of type 'BufferEncoding | undefined'.

92       const buffer = Buffer.from(response.data.content, response.data.encoding)
                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Any ideas what the encoding parameter would be here:
- https://github.com/danger/danger-js/blob/9337fac93a2ea1855f7dd68eb6e2cab0eff4843b/source/platforms/github/GitHubUtils.ts#L91-L93

perhaps the value is always 'utf-8' there,as I would drop it and use the default?

fbartho:
@glensc what you could do is add a parser that converts “string” to BufferEncoding

```typescript
const knownBufferEncodings = [“base64”, TODO]
function verifyEncodingOrThrow(encoding: string): BufferEncoding {
  if (knownBufferEncodings.includes(encoding)) {
    return encoding as BufferEncoding
  }
  throw new Error(`Unsupported response buffer encoding ‘${encoding}’)
}
// Typed in Github, so not tested!
```

fbartho:
Closing this PR as we're on 18 at this point. Thank you for contributing

PR #1262, opened 2022-03-22
Docs: Update dependencies tutorial to use fileMatch

glensc:
I don't remember where I took the new code, probably `index.d.ts`.

This is published here, right?
- https://danger.systems/js/tutorials/dependencies.html


orta:
Yep, that's the right place - thanks!

glensc:
I found also some repos on gitlab.com, are these mirrors or some repos are only there?

glensc:
When (or rather how) is the website updated?
- https://danger.systems/js/tutorials/dependencies.html#lockfiles

PR #1263, opened 2022-03-22
Docs: Use Array.prototype.includes in tutorial

glensc:
Node 6.0.0 has Array.prototype.includes support:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#browser_compatibility

Pardon ahread, for me for mixing argument order. I think I got it correctly, but please verify.


orta:
👍🏻

glensc:
here's another place lodash includes being used:
- https://danger.systems/js/tutorials/node-library.html#keeping-on-top-of-your-library

PR #1266, opened 2022-03-30
docs/tutorials: Fix syntax of JSONDiffForFile dump

glensc:


glensc:
The example has "chalk" in `dependencies.added`, but not in `dependencies.after`. is that a mistake in the example?

glensc:
Also, shouldn't the values of `before` and `after` differ a bit to make it real-life-like? the all versions are the same...

fbartho (approved):

fbartho:
>The example has "chalk" in dependencies.added, but not in dependencies.after. is that a mistake in the example?

@glensc I haven't extensively used that functionality, so I can't tell you what's expected. I suspect we'd have to actually test the code to see.

orta:
Yeah, looking at this test: https://github.com/danger/danger-js/blob/a6b40e953ff318c6230bf0f09152e4ba58cd567f/source/platforms/github/_tests/_github_git.test.ts#L377-L439

The chalk is a typo, the difference is only meant to be the addition of the new dep so I wouldn't expect the change to be too big

Issue #1207, opened 2022-02-08
[BUG] Danger doesn't add a comment if the commented line is not part of the diff


f-meloni:
**Describe the bug**
If your Dangerfile generates an inline comment, which points to a line which not part of your current PR's diff, Danger doesn't post anything on the PR.
The inline comment API doesn't work if the item is not part of the diff, however Danger used to gather all the failed inline comments, and post them as part of the main Danger comment.

**To Reproduce**
1. Create a Dangerfile which comments at line 2 of a random file
2. Change a different file
3. Open a PR

**Expected behavior**
Danger posts a comment with the failure

**Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          | 11.0.2
| danger-swift |  3.12.1
| Operating System | MacOS

**Additional context**
I've investigated it a little bit and saw that a new inline logic is been introduced https://github.com/danger/danger-js/pull/1176 and that might be related

fbartho:
What CI are you testing against?

This definitely used to work on Github, but my team hasn't generated this kind of issue in a while.

orta:
`danger local` posts to the CLI by default, `danger ci` would post to your pull request and not the CLI

carocad:
@orta yes, I know. I was providing the output to show that the output is there but it is not been shown in the PR. Hope it helps.

ThomasGHenry:
Similar thing for us. fail() prints to the github action output, but the status remains successful and there's no comment on the PR. This used to work. I'm not sure when it stopped.

f-meloni:
It might be (just a guess) that before we were trying to post the comments one by one, so if one was failing, that was added to the Danger "global" comment, while now it tries to make only one single review with all the comments, so if it fails to write some of the comments, the API request doesn't fail so the failing comments are not added to the global comment anymore.

fbartho:
@f-meloni I think your guess feels really probable.

Is there any way to resolve this? or is this a gap in the Github API? 

In our repo we had some tools that were regularly adding comments to non-edited files (when a configuration change was needed due to changes elsewhere), so I was able to add a quick user-side workaround for that case, but that still leaves a gap for cases like when you add a new eslint warning that triggers elsewhere in a file that you did actually edit (those comments would still be missing).
[...]

PR #1272, opened 2022-04-21
Fix Danger doesn't add a comment if commented line is not part of diff

Rouby:
This PR should fix the bug mentioned in #1207

### DIFF
--- a/source/runner/Executor.ts
+++ b/source/runner/Executor.ts
@@ -194,7 +194,7 @@ export class Executor {
         output = `Danger: ${tick} passed review, received no feedback.`
       }
 
-      const allMessages = [...fails, ...warnings, ...messages, ...markdowns].map(m => m.message)
+      const allMessages = [...fails, ...warnings, ...messages, ...markdowns].map((m) => m.message)
       const oneMessage = allMessages.join("\n")
       const longMessage = oneMessage.split("\n").length > 30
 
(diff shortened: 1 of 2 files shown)


orta:
This feels pretty good to me (aside from the prettier faff, which I can ignore) thanks!

orta:
Shipped in https://github.com/danger/danger-js/releases/tag/11.0.4

source/runner/Executor.ts
```diff
@@ -399,22 +399,28 @@ export class Executor {
...
-      this.sendInlineReview(git, inlineResultsForReview).catch(_e =>
-        inlineResultsForReview.forEach(inlineResult => inlineResultsIntoResults(inlineResult))
-      ),
+      this.sendInlineReview(git, inlineResultsForReview)
+        .then((_r) => emptyDangerResults)
+        .catch((_e) =>
+          inlineResultsForReview
+            .map((inlineResult) => inlineResultsIntoResults(inlineResult))
+            .reduce(mergeResults, emptyResult)
+        ),
```
fbartho, line 419: Great fix! I had a suspicion the root issue was around here, but the 3 times I traced it, I couldn’t figure it out — indeed it looked like everything was being caught & handled..

fbartho (commented):
Great work!! This has been grinding my gears.

PR #1275, opened 2022-04-25
Refactor Dockerfile to reduce image size

### LINKED ISSUE #1274: [Improvement] container built from official dockerfile is abnormally large
**Describe the bug**
Building a container using the provided `Dockerfile` results in a container that is over 1GB

**To Reproduce**
1. Clone repo
2. run `docker build .`
3. observe container is very large
```
danger-js                                                                      latest                 2ab391c36636   3 minutes ago   1.17GB
```

**Expected behavior**
Container should contain only the build artifacts and dependencies required to run them along with whatever userspace is provided by the base container. In the case of this project I would expect the resultant container to be under 200MB.

**Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          | HEAD
| node             | 14
| npm              | 6.14.16
| Operating System | linux

drawks:
* resolves #1274
* refactors `Dockerfile` to optimize for size
  - converts to use a two stage build with only the minimum in
    the final stage
  - ~~applies `yarn-audit-fix` during build stage setup to automatically
    upgrade to latest patch version of dependencies~~
  - converts final build step to use the `--production` flag to
    prevent development dependencies from being present in final
    image
  - ensures that typescript is installed in resulting container

~~In addition to reducing the overall size of the container (reduced from over 1gb to under 200MB) this change results in container with far fewer known vulnerable node module.~~

```
before   latest                 2ab391c36636   23 minutes ago       1.17GB
after      latest                 91697973ff7b   About a minute ago   245MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,12 +11,14 @@ COPY package.json yarn.lock ./
 RUN yarn install
 COPY . .
 RUN yarn run build:fast
+RUN yarn remove 'typescript' --dev && yarn add 'typescript@^4.5.5'
 RUN yarn install --production --frozen-lockfile
 RUN chmod +x distribution/commands/danger.js
 
 
 FROM node:14-slim
 WORKDIR /usr/src/danger
+ENV PATH="/usr/src/danger/node_modules/.bin:$PATH"
 COPY package.json ./
 COPY --from=build /usr/src/danger/distribution ./dist
 COPY --from=build /usr/src/danger/node_modules ./node_modules
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:14-slim
+FROM node:14-slim as build
 
 LABEL maintainer="Orta Therox"
 LABEL "com.github.actions.name"="Danger JS Action"
@@ -6,12 +6,22 @@ LABEL "com.github.actions.description"="Runs JavaScript/TypeScript Dangerfiles"
 LABEL "com.github.actions.icon"="zap"
 LABEL "com.github.actions.color"="blue"
 
-RUN mkdir -p /usr/src/danger
-COPY . /usr/src/danger
-RUN cd /usr/src/danger && \
-  yarn && \
-  yarn run build:fast && \
-  chmod +x distribution/commands/danger.js && \
-  ln -s $(pwd)/distribution/commands/danger.js /usr/bin/danger
+WORKDIR /usr/src/danger
+RUN yarn global add yarn-audit-fix
+COPY package.json yarn.lock ./
+RUN yarn install && \
+    yarn-audit-fix
+COPY . .
+RUN yarn run build:fast
+RUN rm -rf node_modules
+RUN yarn install --production --frozen-lockfile
+
+FROM node:14-slim
+WORKDIR /usr/src/danger
+COPY package.json ./
+COPY --from=build /usr/src/danger/distribution ./dist
+COPY --from=build /usr/src/danger/node_modules ./node_modules
+RUN chmod +x /usr/src/danger/dist/commands/danger.js && \
+    ln -s /usr/src/danger/dist/commands/danger.js /usr/bin/danger
 
 ENTRYPOINT ["danger", "ci"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ COPY package.json yarn.lock ./
 RUN yarn install
 COPY . .
 RUN yarn run build:fast
-RUN yarn remove 'typescript' --dev && yarn add 'typescript@^4.5.5'
+RUN yarn remove 'typescript' --dev && yarn add 'typescript'
 RUN yarn install --production --frozen-lockfile
 RUN chmod +x distribution/commands/danger.js
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,21 +7,19 @@ LABEL "com.github.actions.icon"="zap"
 LABEL "com.github.actions.color"="blue"
 
 WORKDIR /usr/src/danger
-RUN yarn global add yarn-audit-fix
 COPY package.json yarn.lock ./
-RUN yarn install && \
-    yarn-audit-fix
+RUN yarn install
 COPY . .
 RUN yarn run build:fast
-RUN rm -rf node_modules
 RUN yarn install --production --frozen-lockfile
+RUN chmod +x distribution/commands/danger.js
+
 
 FROM node:14-slim
 WORKDIR /usr/src/danger
 COPY package.json ./
 COPY --from=build /usr/src/danger/distribution ./dist
 COPY --from=build /usr/src/danger/node_modules ./node_modules
-RUN chmod +x /usr/src/danger/dist/commands/danger.js && \
-    ln -s /usr/src/danger/dist/commands/danger.js /usr/bin/danger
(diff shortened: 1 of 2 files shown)


orta:
Yeah, this is good - I expect this might end up dropping TypeScript support for folks - any chance that dependency can be added manually to the build process. Then I'm good to merge

Dockerfile
```diff
@@ -1,17 +1,27 @@
...
 
-RUN mkdir -p /usr/src/danger
-COPY . /usr/src/danger
-RUN cd /usr/src/danger && \
-  yarn && \
-  yarn run build:fast && \
-  chmod +x distribution/commands/danger.js && \
-  ln -s $(pwd)/distribution/commands/danger.js /usr/bin/danger
+WORKDIR /usr/src/danger
+RUN yarn global add yarn-audit-fix
```
airtonix, line 10: This should really be a separate process.

like using dependabot.

Beacuase now, you have something at a gitref (the pkg lock file) but it results in different outcome in the resulting build.

airtonix:
> Yeah, this is good - I expect this might end up dropping TypeScript support for folks - any chance that dependency can be added manually to the build process. Then I'm good to merge

Apart from having a Dockerfile that demonstrates how to build it into someone elses own image, I'm still not even convinced of the benefit of have a dangerjs docker image.

Most things will require people to bring in other packages anyway, the most obvious is Typescript. Which then illustrates where this becomes a slippery slope: how many packages will you start including in the docker image?

orta:
I'm not here to debate whether it makes sense to make a Danger dockerfile, enough people have asked and eventually added it themselves. _I_ don't think it makes sense, but that doesn't mean folks in esoteric build/CI systems have the same constraints as me.

Having this PR break someones docker setup because you can't run a dangerfile.ts isn't a slippery slope argument, for me that's the baseline with respect for danger to run in a project. Which I think is pretty reasonable as a constraint.

ivankatliarchuk:
There is definitely a benefit. Upstream official docker file always nice to have. Everyone could beald an image `FROM` an official one.

probably worth to add `.dockerignore` too

```
node_modules/
.git/
env/
.github/
.vscode/
```

Dockerfile
```diff
@@ -1,17 +1,27 @@
...
+RUN yarn run build:fast
+RUN rm -rf node_modules
+RUN yarn install --production --frozen-lockfile
+
+FROM node:14-slim
+WORKDIR /usr/src/danger
+COPY package.json ./
+COPY --from=build /usr/src/danger/distribution ./dist
+COPY --from=build /usr/src/danger/node_modules ./node_modules
+RUN chmod +x /usr/src/danger/dist/commands/danger.js && \
```
glensc, line 24: do the chmod in build stage

Dockerfile
```diff
@@ -1,17 +1,27 @@
...
-  chmod +x distribution/commands/danger.js && \
-  ln -s $(pwd)/distribution/commands/danger.js /usr/bin/danger
+WORKDIR /usr/src/danger
+RUN yarn global add yarn-audit-fix
+COPY package.json yarn.lock ./
+RUN yarn install && \
+    yarn-audit-fix
+COPY . .
+RUN yarn run build:fast
+RUN rm -rf node_modules
```
glensc, line 16: don't `rm -rf node_modules`, yarn removes superfuous packages (from dev dependencies) already itself

drawks: I haven't touched this for a bit, just looking over your reviews now. But IIRC it wasn't removing them... or it was applying some sort of heuristic that was resulting in some build-time dependencies not being removed. FWIW I'm not really a node/js guy; but wanted to use this tool and it SEEMED like the container was much larger than it needed to be

drawks:
> Yeah, this is good - I expect this might end up dropping TypeScript support for folks - any chance that dependency can be added manually to the build process. Then I'm good to merge

As I mentioned above i'm not really a node/js dev, if it is expected that Typescript is always needed alongside danger would it make sense to add it to `packages.json` so it is always pulled as a dependency?

Dockerfile
glensc: https://github.com/wagoodman/dive might be useful to check what's in the layers

package.json
```diff
@@ -178,7 +178,8 @@
     "readline-sync": "^1.4.9",
     "regenerator-runtime": "^0.13.9",
     "require-from-string": "^2.0.2",
-    "supports-hyperlinks": "^1.0.1"
+    "supports-hyperlinks": "^1.0.1",
+    "typescript": "^4.5.5"
```
glensc, line 182: 1. you need to move it to "dependencies", i.e remove from "devDependenencies" as well, or project might end up with conflicting versions (i.e somebody updates only or dependencies devDependenencies)
2. I'm not familiar how this project parses .ts files runtime, you sure it's not using ts-node?

orta: Yep, we're definitely not adding typescript to the dependencies - it's already the `devDependencies` section

glensc: the opposite, **Move** from devDependencies to dependencies, not just "**Add**", if you follow the discussion.

- https://github.com/danger/danger-js/pull/1275#issuecomment-1428491746

drawks: Done

orta (changes requested):
As I mentioned in the comments, we're not making TypeScript a dependency on Danger (as it's not an explicit dependency) - you'll need to figure out a different way to do this 

Perhaps it can be added inside the dockerfile?

drawks:
> Perhaps it can be added inside the dockerfile?

Gotcha, I can do that

Dockerfile
```diff
@@ -1,17 +1,27 @@
...
-  yarn && \
-  yarn run build:fast && \
-  chmod +x distribution/commands/danger.js && \
-  ln -s $(pwd)/distribution/commands/danger.js /usr/bin/danger
+WORKDIR /usr/src/danger
+COPY package.json yarn.lock ./
+RUN yarn install
+COPY . .
+RUN yarn run build:fast
+RUN yarn remove 'typescript' --dev && yarn add 'typescript@^4.5.5'
```
orta, line 14:
```suggestion
RUN yarn remove 'typescript' --dev && yarn add 'typescript'
```

I'm going to switch this to a versionless version - because the API we use from the lib will never break, and then folks don't need to think about whether they can use specific TS features

orta:
Cool, this seems reasonable to me 👍🏻

Dockerfile
glensc: that is bad. typescript breaks even in minor updates. simple example `catch (e)` go changed to `catch (e: unknown)` somewhere in 4.5 -> 4.9.

how about using this tool specifically for this purpose:
- https://github.com/glensc/yarn-install-dev

glensc:
@drawks your git rebase totally messed up, some dependabot updates pulled in:
- https://github.com/danger/danger-js/pull/1275/commits

@orta why you have merged these? pull request doesn't even mention updating dependencies. should have been separate pr

orta:
I read the lockfile changes, they were harmless and only affected this repo

glensc:
I wonder where the dependabot changes came. from it's not enabled in https://github.com/danger/danger-js repo... at least no open pr's here.

drawks:
That was my fault, I switched computers on my last update and accidentally rebased on my own fork which has dependabot enabled.

Apologies.

PR #1277, opened 2022-05-03
Fix dead links

martinerko:
I noticed some of the links are no longer working due to the fact pointing to repositories that have been archived or they have stopped using danger completely.


orta:
Thanks!

Issue #1276, opened 2022-05-02
[BUG] Bitbucket cloud fails to get DSL in 1 second


OJPARKINSON:
**Describe the bug**
Bitbucket is hitting the 1 secound timeout before it is able to get the DSL, the timeout is there so the CI doesn't hang if broken.

**To Reproduce**
1. Setup for bitbucket cloud and codebuild 
2. run danger ci
3. recieve "Timeout: Failed to get the Danger DSL after 1 second"

**Expected behavior**
After upping the timout to 2000ms it gives enough time to get the DSL.

`source/commands/danger-runner.ts:86`

The easy fix is to just up the timeout but I'm not sure how that will effect other danger setups. Could possibly pass the timeout as config?

orta:
Yeah, 1 second seems actually really low to me - I'd happily take a PR making that something like 10s

OJPARKINSON:
Great, I'll get one in this afternoon.

orta:
This shipped!

PR #1278, opened 2022-05-03
Upped the DSL timeout to 10 seconds

OJPARKINSON:
PR based on #1276

"Bitbucket is hitting the 1 secound timeout before it is able to get the DSL, the timeout is there so the CI doesn't hang if broken."


Acarr44:
I also need this, can it be merged?

orta:
Good timing, I can only deploy on my Mac and  just happen to be on it now

PR #1281, opened 2022-05-30
Allowing slashes in branch names

OJPARKINSON:
I'm currently running into a few issues with branch names that include a slash like `feature/branch-name`. So this means that `CODEBUILD_WEBHOOK_TRIGGER` will look like `branch/feature/branch-name` and danger then splits the string into an array and looks for the second index('feature'). This misses the rest of the name meaning that the PR cannot be found and Danger errors out.

To combat this I've swapped out the split to arrays and seen as we know what the ENVs will be formatted we can just check the start of the strings and then replace the part we don't want.


fbartho (approved):
This seems reasonable to me, however I’ve never used codebuild, so take that with a grain of salt!

orta:
Agree, looks good to me too

PR #1287, opened 2022-07-05
Fixed Danger to find its own inline comment with an exact ID match

nsugiyam:
Fixed a bug where Danger would delete inline comments posted by other Danger with the same prefix ID.

Same logic as `getDangerCommentIDs()`.
https://github.com/danger/danger-js/blob/630f2a49cfa1e7e4b87173ee9754278e00452ac3/source/platforms/github/GitHubAPI.ts#L92


orta:
Ah, I see:

> `export const dangerIDToString = (id: string) => `danger-id-${id};``

Adds a semi at the end which would break and substring matches 👍🏻

PR #1291, opened 2022-07-22
Make BitBucket Server username matching case insensitive

### LINKED ISSUE #1290: BitBucket Server username matching is case sensitive
### Discussed in https://github.com/danger/danger-js/discussions/1289

<div type='discussions-op-text'>

<sup>Originally posted by **DavidBrunow** July 14, 2022</sup>
Yesterday I spent about half my day tracking down an issue where our BitBucket Server was not deleting previous inline comments and therefore each time our build jobs ran it would add the same comments that had already been added. The root cause was that the username we were using to run Danger, for example “sampleusername”, was not capitalized the same way as the username configured in BitBucket Server, for example “sampleuserNAME”.

Since BitBucket Server (and I believe all other hosts although I don’t have experience with GitLab) treats usernames as case-insensitive should we change Danger’s behavior as well?

Here is the line of code that I eventually found at the end of my debugging journey yesterday:

https://github.com/danger/danger-js/blob/main/source/platforms/bitbucket_server/BitBucketServerAPI.ts#L235

Thoughts on `toLowerCase()`ing both the usernames there?
</div>

DavidBrunow:
This fixes #1290.


source/platforms/bitbucket_server/BitBucketServerAPI.ts
```diff
@@ -213,7 +213,7 @@ export class BitBucketServerAPI implements BitBucketServerAPIDSL {
 
     return comments
       .filter(comment => comment!.text.includes(dangerIDMessage))
-      .filter(comment => username || comment!.author.name === username)
```
DavidBrunow, line 216: I'm not sure the history behind the `username || ` part of this code but it does not seem like the proper logic? Happy to revert this if I need to.

source/platforms/bitbucket_server/_tests/_bitbucket_server_api.test.ts
```diff
@@ -241,6 +241,49 @@ describe("API testing - BitBucket Server", () => {
     ])
   })
 
+  it("getDangerCommentsCaseInsensitive", async () => {
```
DavidBrunow, line 244: I added extra tests rather than changing the existing tests to have case differences in the usernames. I went this way to it is more obvious why the usernames don't match perfectly but I'm happy to make changes to go any other direction.

source/platforms/bitbucket_server/BitBucketServerAPI.ts
orta: Yeah, I can't figure it out either, so lets take it and see

orta:
👍🏻

orta:
I have to release on a mac nowadays, so nextx time I'm on one I'll ship a build

Issue #1302, opened 2022-08-18
[BUG] Incompatible with git config diff.noprefix


lode:
**Describe the bug**
When having git configured to hide prefixes on file names in the diff output (https://git-scm.com/docs/git-config#Documentation/git-config.txt-diffnoprefix) fetching a list of changed files fails.

**To Reproduce**
1. Run `danger local` (maybe also works with ci)
2. See nice output
3. Run `git config --global diff.noprefix "true"`
4. Run `danger local` (maybe also works with ci)
5. See an error

```
Error:  TypeError: Cannot read properties of null (reading 'map')
    at parseFile (<...>/node_modules/parse-diff/index.js:159:13)
    ...
```

**Expected behavior**
It should work also when people configure git in a different way. I think it is also possible, depending on how we get the diff output. At least running `git diff --src-prefix="a/" --dst-prefix="b/"` forces needed prefixes even though the global config is set to remove them.

**Your Environment**
 | software         | version
| ---------------- | -------
| danger.js          |10.9.0
| node             |16.17.0
| npm              |8.15.0
| Operating System |Ubuntu 22.04

**Additional context**
If of any interest, I'm using `danger.git.fileMatch`, `danger.git.created_files`, `danger.git.modified_files`, `danger.git.structuredDiffForFile` in my dangerfile. However, it seems to trigger regardless, even on an empty dangerfile.

orta:
Yeah, perhaps the command we run under the hood can be explicit in requiring the prefixes from git, I'm open to PRs there

lode:
:+1: I'm willing to try a PR, though I couldn't find where this was ran. Can you give me a pointer?

lode:
Diving a little deeper.

If I understand correctly this is called in https://github.com/danger/danger-js/blob/main/source/platforms/git/diffToGitJSONDSL.ts. It relies on `parseDiff`, and also does an assumption itself on the prefixes: https://github.com/danger/danger-js/blob/main/source/platforms/git/diffToGitJSONDSL.ts#L20.

I get the feeling this should first be fixed in https://github.com/sergeyt/parse-diff. Is that correct?

lode:
Sorry for the monologue, but I think I found a place in this repo. I'll try a little and will stop spamming here for a while :)

PR #1303, opened 2022-08-18
Fix `danger local` to play nice with git config diff.noprefix

lode:
This fixes https://github.com/danger/danger-js/issues/1302.

As far as I understood a similar fix isn't needed for GitHub/GitLab, as they get their diff from the API.

I didn't wrote any tests as I can imagine it will be hard/impossible to test with git config.


orta:
👍🏻  looks reasonable to me, and yeah, I agree WRT tests

Issue #1304, opened 2022-08-21
[BUG] Staged changes doesn't use just staged, and returns empty diffs


lode:
**Describe the bug**
When calling danger local with `--staging` I also get files which are committed and still in working directory. Also, the `structuredDiffForFile` returns empty diffs for the staged and working directory files.

**To Reproduce**
1. Create a branch, make a change which would trigger a danger check and commit it
2. Make another such change and run `git add .`, and leave it as is, staged but not committed
3. Make another such change and leave it as is, in the working directory, not staged not committed
4. Run `danger local --staging`
5. When using `danger.git.fileMatch(...).getKeyedPaths()` all file paths are in there
6. When using `danger.git.structuredDiffForFile(...)` a diff from the committed file is returned, and `null` is returned for the staged and working directory files

**Expected behavior**
1. I expect only the staged file to be inside `getKeyedPaths()`. (I actually don't mind getting the working directory file as well, however, the help for danger-local states '_Just_ use staged changes'.) This seems fixed by using `--staged` instead of `base` here: https://github.com/danger/danger-js/blob/1e7ee107b3755f208ff1ec3dd393e83045340b35/source/platforms/git/localGetDiff.ts#L6 **Update** here's a PR https://github.com/danger/danger-js/pull/1305.
2. Second, I expect to get the actual diff when calling `structuredDiffForFile()`. It seems that the `--staging` setting isn't passed to `localGetDiff()` from everywhere. I've hard-coded https://github.com/danger/danger-js/blob/1e7ee107b3755f208ff1ec3dd393e83045340b35/source/platforms/git/localGetDiff.ts#L9 to always use `useStagingChanges` regardless of the setting. When running `--staging` with that change, the output is as expected. **Update** here's a PR https://github.com/danger/danger-js/pull/1306.

**Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          | 11.1.1
| node             | 16.17.0
| npm              | 8.15.0
| Operating System | Ubuntu 22.04

**Additional context**
I've setup an example repository showing the problem: https://github.com/lode/danger-js-test. It includes a dangerfile with outputs.

lode:
@orta thanks for reviewing, improving and merging!

This issue is fixed via #1305 and #1306.

PR #1305, opened 2022-08-21
Fix `fileMatch` to not return committed or working-dir files when using `--staging`

lode:
Changes `git diff <base>` to `git diff --staged`. The first one also returns changes from base to working-dir, and thus includes working-dir and committed files. The second only returns changes in the staging area.

Currently, this actually only fixes the file paths, which is one of the issues in https://github.com/danger/danger-js/issues/1304.
It doesn't fix the diffs yet. See https://github.com/danger/danger-js/issues/1304 for more details.


orta:
Thanks!

PR #1306, opened 2022-08-21
Fix `structuredDiffForFile` to return staged diffs when using `--staging`

lode:
The `--staging` argument is used to select files with `fileMatch`, but when fetching diffs from `structuredDiffForFile()`, the `--staging` argument is ignored. This PR passes the `--staging` argument to that method as well and changes `null` diffs into actual diffs from the staging area.

This fixes the second issue from https://github.com/danger/danger-js/issues/1304. It needs https://github.com/danger/danger-js/pull/1305 to fix the correct file paths.

Note I'm not sure about the correctness of the code. I've made `yarn test` happy, but my ts knowledge is not that good to know for sure whether I'm doing the right thing. Also I see the `staging` argument is missing for the config of the other platforms, and I'm unsure what to fill it with there.

### DIFF
--- a/source/platforms/git/gitJSONToGitDSL.ts
+++ b/source/platforms/git/gitJSONToGitDSL.ts
@@ -26,7 +26,7 @@ export interface GitJSONToGitDSLConfig {
   repo?: string
 
   /** Whether to diff only files from the staging area */
-  staging: boolean = false
+  staging: boolean
 
   /** The sha things are going into */
   baseSHA: string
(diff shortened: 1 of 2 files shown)


source/platforms/git/gitJSONToGitDSL.ts
```diff
@@ -35,7 +36,7 @@ export interface GitJSONToGitDSLConfig {
   /** A promise which will return the string content of a file at a sha */
   getFileContents: (path: string, repo: string | undefined, sha: string) => Promise<string>
   /** A promise which will return the diff string content for a file between shas */
-  getFullDiff?: (base: string, head: string) => Promise<string>
+  getFullDiff?: (base: string, head: string, staging: boolean) => Promise<string>
```
orta, line 39:
```suggestion
  getFullDiff?: (base: string, head: string, staging?: boolean) => Promise<string>
```

Is probably what you're looking for IMO

orta:
Cool, looks good thanks!

orta:
All shipped!

source/platforms/git/gitJSONToGitDSL.ts
lode: :+1: thx!

Issue #1311, opened 2022-09-16
[BUG] Unable to run concurrent Danger instances on a single machine.


squarefrog:
**Describe the bug**
We run Danger via Danger Swift, on a self-hosted Mac Mini using GitLab Runners. We have a single runner setup, with `concurrent` enabled.

Danger Swift passes in the DSL via URL, which DangerJS writes to the temporary directory:

```ts
  const sendDSLToSubprocess = () => {
    if (exec.options.passURLForDSL) {
      const resultsPath = join(tmpdir(), "danger-dsl.json")
      writeFileSync(resultsPath, dslJSONString, "utf8")
      const url = `danger://dsl/${resultsPath}`
      d(`Started passing in STDIN via the URL: ${url}`)
      child.stdin.write(url)
      child.stdin.end()
    } else {
      d(`Started passing in STDIN`)
      child.stdin.write(dslJSONString)
      child.stdin.end()
    }
    d(`Passed DSL in via STDIN`)
  }
```

However, this is problematic as this results in a URL like the following, which both processes try to read/write from at the same time:

```
danger://dsl//var/folders/0h/c4gthkj13sz3hb1z4fwpnd3c0000gp/T/danger-dsl.json
```

Instead we should probably suffix a unique directory, for example:

```
danger://dsl//var/folders/0h/c4gthkj13sz3hb1z4fwpnd3c0000gp/T/danger-js/230ebef4-35ce-11ed-a261-0242ac120002/danger-dsl.json
```

**To Reproduce**
1. Setup an environment that lets you test locally
2. In one Terminal window, kick off a Danger run
3. In another Terminal window, kick off another Danger run

**Expected behavior**
The temporary DSL file should be placed in a unique directory, so that one process wont cleanup the DSL while another is using it.

**Your Environment**
 | software         | version
| ---------------- | -------
| danger.js          |  11.1.2
| node             |  N/A
| npm              | N/A
| Operating System | macOS 12.6

**Additional context**
I submitted a [similar fix](https://github.com/danger/swift/pull/530) for danger-swift, where we use a UUID for a folder name. This works well in my testing. If you're happy with this approach, I'm happy to try create a PR for this change. I'm unaware if there are any complications for doing this though. Checking the source code there are two places where `tmpdir()` is used, and this issue could potentially occur:

* [runDangerSubprocess.ts](https://github.com/danger/danger-js/blob/main/source/commands/utils/runDangerSubprocess.ts#L57)
[...]

PR #1312, opened 2022-09-17
Add random chars to files to support concurrency

squarefrog:
As discussed in #1311, we're currently experiencing an issue where we have two or more concurrent processes running Danger. Both processes attempt to read/write to the same file locations on disk, which means we have a race condition where only one process can complete successfully.

This PR attempts to remedy this by appending a random 4 byte token to the end of the filename to prevent clashes. For example:

```
danger://dsl//var/folders/0h/c4gthkj13sz3hb1z4fwpnd3c0000gp/T/danger-dsl-1e170c2b.json
```

I've not written JS before, so I'd love to hear if this could be improved. Additionally, a pre-commit hook seems to have done some lint/formatting. Let me know if this should not be included and I can send a raw diff.

**Note:** I've created a package of this PR and applied to one of our CI servers. We'll let it run concurrently for a few tests to ensure this works manually, as I'm not show how to unit test this enhancement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 ## Main
 
 <!-- Your comment below this -->
-
+- Append random string to danger-results.json and danger-dsl.json files to better support concurrent processes #1311
 
 <!-- Your comment above this -->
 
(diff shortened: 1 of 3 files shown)


squarefrog:
I created a binary for this with `yarn package` and replaced our current brew binary with it. It worked a treat! I pushed the same commit to both branches at the same time, normally one of these pipeline runs would have failed.

orta:
Sorry, been out at nsspain (and I have to deploy danger on my mac nowdays, so it may be next week when I can make a dpeloy) but this is good IMO - the formatting changes are fine

squarefrog:
Sounds good, no rush. We can use the self-built package for now. Thanks for your help with this!

Issue #1283, opened 2022-06-08
[BUG] tsconfig.json "extends" field not parsed by transpiler


zacharyliu:
**Describe the bug**
When using the built-in native TypeScript transpiler, the `extends` field is not supported and only the explicit `compilerOptions` are used. This field is sometimes used in monorepos to extend a shared config, or to extend one of the [official "base" templates](https://github.com/tsconfig/bases).

**To Reproduce**
0. Install a base config: `npm install @tsconfig/node16`
1. Create a `tsconfig.json` extending that config:
```json
{
  "extends": "@tsconfig/node16/tsconfig.json"
}
```
2. Create a Dangerfile which tries to use non-commonjs features:
```ts
import { exec } from "child_process";
```
3. Try to run Danger with `danger ci`. It will report an error like "cannot use import outside a module", as the `module: "commonjs"` option from the base config does not get inherited correctly.

**Expected behavior**
Danger should support all official methods of configuring tsconfig.json.

 | software         | version
| ---------------- | -------
| danger.js          |  11.0.7
| node             | v16.15.0
| npm              | 8.5.5
| Operating System | macOS 12.4

**Additional context**
I believe the bug is because Danger's transpiler directly loads the tsconfig file contents and passes it to `ts.transpileModule`:

https://github.com/danger/danger-js/blob/630f2a49cfa1e7e4b87173ee9754278e00452ac3/source/runner/runners/utils/transpiler.ts#L128-L134

This can be fixed by using the TypeScript APIs for parsing the config options. For example, in `ts-node`:
https://github.com/TypeStrong/ts-node/blob/14323f9d00d5c7051ac09b944c7f423e442145ea/src/configuration.ts#L301-L317

Some of the other functions in this file, like `lookupTSConfig`, can also use the native APIs instead.

fbartho:
Thanks for filing @zacharyliu -- this seems pretty important! Do you feel comfortable filing a PR for this?

@orta -- is there any reason @zacharyliu's proposed change of using `lookupTSConfig` and `parseJsonConfigFileContent` wouldn't work?

https://github.com/microsoft/TypeScript/blob/b8f648832379005fc8c3c9b34cc5e4acd01653e4/src/compiler/commandLineParser.ts#L2679-L2688

orta:
Off-hand, no, I'd follow ts-node - given the amount of traffic they've seen, it should be a reasonable pattern to replicate

PR #1313, opened 2022-09-22
Add support for tsconfig.json files that have an `extends` property

Jtupiter:
### What does the PR do?
Fixes issue https://github.com/danger/danger-js/issues/1283.

Following what was suggested in the issue to mimic what `ts-node `is doing to resolve the `tsconfig.json` [here](https://github.com/TypeStrong/ts-node/blob/14323f9d00d5c7051ac09b944c7f423e442145ea/src/configuration.ts#L303).
Their method is quite a bit more complex since they need to resolve custom `ts-node` compiler options.

For our purposes, reading the config file using `ts.readConfigFile` and then parsing it with `ts.parseJsonConfigFileContent` produces the expected `tsconfig.json` even when it extends another one.

### Tests
Added a test to confirm that danger resolves extended tsconfigs.

### DIFF
--- a/source/runner/runners/utils/_tests/_fixtures/tsconfig.base.json
+++ b/source/runner/runners/utils/_tests/_fixtures/tsconfig.base.json
@@ -1,0 +1,6 @@
+{
+  "compilerOptions": {
+    "target": "ES6",
+    "module": "ES2015"
+  }
+}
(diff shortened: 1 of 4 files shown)


orta:
Hi, I'm a little surprised by the failing tests here - do you have a sense of why it broke?

Jtupiter:
> Hi, I'm a little surprised by the failing tests here - do you have a sense of why it broke?

Yeah, I was confused by this. It's completely unrelated to my change and it passes locally. The failing snapshot also looks correct, which makes me think it has something to do with line endings. Will see if I can resolve this.

PR #1320, opened 2022-10-02
fix: gitlab api `diffForFile`

ivankatliarchuk:
# BUG

## Impact

Gitlab functionality is broken. Construct `await danger.git.diffForFile(file)` throw an unhadled exception when file not found

--- 

found a regression.
 with `gitlab` library
```json
if ((e as any).response.status === 404) {
        return ""
      }
```

`@gitbreaker` library
```json
if ((e as any).response.statusCode === 404) {
        return ""
      }
```

gitbreaker does handle errors in [un-documented way](https://github.com/jdalrymple/gitbeaker/issues/396)

tested locally with local installation
```sh
yarn build && yarn pack
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,7 +149,6 @@
     "fast-json-patch": "^3.0.0-1",
     "get-stdin": "^6.0.0",
     "@gitbeaker/node": "^21.3.0",
-    "gitlab": "^10.0.1",
     "http-proxy-agent": "^5.0.0",
     "https-proxy-agent": "^5.0.1",
     "hyperlinker": "^1.0.0",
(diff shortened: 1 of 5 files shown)


ivankatliarchuk:
Hi, @orta . Would it be possible to review this PR? The bug was introduced, probably some pipelines on Gitlab are failing if they automatically update danger version, ups...

orta:
No need to ping me after a day, no-one gets a SLA on open source I run in my spare time.

That said this looks good, not sure when I'll get around to being on my mac to deploy, but we can merge 👍🏻

PR #1321, opened 2022-10-02
feat: added Github templates

ivankatliarchuk:
Why

- for a new contributor templates could help to 
- some default text whem PR is created
-  useful to make sure to follow a standard process for every pull request


orta:
No thanks, I think we're doing pretty fine without them - we don't need to add extra process

Issue #1301, opened 2022-08-14
[BUG] Gitlab Package Warning


uchihamalolan:
**Describe the bug**
When I do `pnpm install danger`, I get the following warning about "gitlab" package

```bash
WARN  deprecated gitlab@10.2.1: The gitlab package has found a new home in the @gitbeaker organization. For the latest gitlab node library, check out @gitbeaker/node. A full list of the features can be found here: https://github.com/jdalrymple/gitbeaker#readme
```

| software         | version
| ---------------- | -------
| danger.js        | 11.1.1
| node             | 16.15.1
| pnpm             | 7.9.0
| Operating System | MacOS Montrey (12.5)

orta:
Thanks, you're welcome to port us over 👍🏻

glensc:
this is done via https://github.com/danger/danger-js/pull/1318

PR #1325, opened 2022-10-06
issue-1301: update gitbreaker version

ivankatliarchuk:
issue https://github.com/danger/danger-js/issues/1301


ivankatliarchuk:
Looks like will need some assistance e.g. how to fix `yarn declarations` and logic in file `scripts/danger-dts.ts`

Issue #1138, opened 2021-05-11
[Feature Request] Use Thread instead of comment


shyim:
**Describe the bug**
Sometimes you want that people resolve the danger comment. Example on a fail message.

I would like a new command-line option that forces to create a thread instead of a comment.

What do you think about that @orta . I would also provide a PR

orta:
I don't want a CLI flag for it, it's a DSL attribute because it differs per code review service

shyim:
So instead using warn/fail etc. Custom method like danger.github.createDiscussion() ? 🤔

orta:
Hrm, maybe it should be a CLI param, we do that for checks. That would make life easy for downstream languages like swift/kotlin/python etc and doesn't change the JSON contract either.

I'm not sure on the name, because it's a 'review' in GitHub, a 'discussion' in GitLab and I'm sure bitbucket/azure have their own thing too. For now, let's say `--use-thread` but I might change that opinion in the future 👍🏻

fbartho:
I think this feature is now implemented by #1176 for Github (there's no Commandline option, that's just how DangerJS now works). Do you want to do the same for Gitlab @kuba-gaj?

dan-trewin:
@fbartho I would very much be in favor of danger comments being threads (rather than notes) for gitlab.

The reason being: In gitlab you can require all threads be resolved before merging an MR is allowed.

My team's usecase is when danger reports potentially dangerous errors. This should block the MR by default (a thread would effectively do this), but if the author/contributors are sure that the changes are safe, they can resolve the danger comment. Effectively, this is them acknowledging the potentially dangerous errors.

Honestly, it would be nice if notes/threads were both an option, but since danger with github defaults to threads, I would think gitlab should work the same.

fbartho:
I have no stake in the gitlab platform, so I’m not going to say anything other that that your point does seem reasonable! Motivated contributors should feel free to make a clean PR to do this!

That said: each repository vendor has slightly different concepts, so I wouldn’t use GitHub to decide what to do on Gitlab. We want these platforms to feel natural for devs that only use a given provider.

olte-mms:
We also looking for such a implementation

OtaconnSHGR:
The feature would be a most welcome one

orta:
This isn't the sort of OSS project where you can request features and the maintainers will build it for you, so +1s are not useful here. If you'd like to see it, I'd recommend you start looking at implementing it 👍🏻

[...]

PR #1331, opened 2022-10-28
GitLab: Add support to use threads instead of comments

uncaught:
- Add support to use threads instead of comments
- Implements #1138 

# Implementation

- Threads are used if an environment variable `DANGER_GITLAB_USE_THREADS` is set to a value of `1` or `true`.
- Threads are completely removed with all answers (even of other users) if danger removes its main comment.

## GitLab API change

- A new method `getMergeRequestDiscussions` fetches all the threads (discussions) of a merge request
- **BREAKING** The method `createMergeRequestDiscussion` has had its second argument changed from being the position object to be the actual options object for that gitbreaker API call, which contains the position.

# Migration considerations

- You can switch to using threads and back any time.
- Only newly created comments are affected by the current setting.
- Because existing comments can be turned into a thread by simply replying to them, danger will always delete its entire threads, no matter the setting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@
 - GitLab: Added provider tests [#1319](https://github.com/danger/danger-js/pull/1319) [@ivankatliarchuk]
 
 - GitHub: Added `danger.github.pr.draft` field to DSL
-
+- GitLab: Add support for using threads instead of comments [#1331](https://github.com/danger/danger-js/pull/1331) [@uncaught]
 <!-- Your comment above this -->
 
 ## 11.1.2
--- a/docs/usage/gitlab.html.md
+++ b/docs/usage/gitlab.html.md
@@ -41,3 +41,8 @@ danger.gitlab.
   /** The commits associated with the merge request */
   commits: GitLabMRCommit[]
 ```
+
+---
+
+If you want danger to open threads (discussions) instead of just commenting in merge requests, set an environment
+variable `DANGER_GITLAB_USE_THREADS` with value `1` or `true`.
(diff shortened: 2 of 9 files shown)


orta:
Hey, this looks  good to me from a read over it - I'm quite wary of this change being so substantial and only really having one test to ensure it doesn't break in the future. So, I'll give it a week to settle, _if you can_ it'd be good to get some tests - if not, that's fine.

uncaught:
Thanks, @orta, I've added some tests - there were none at all before for the GitLab.ts.

I've also changed one key detail: danger will always delete its entire threads now, no matter the setting. Because I've noticed that you can simply turn danger's comments into threads by replying to them as another user. And that would leave the replies dangling if danger was then to remove its original comment.

[GitLab's discussions API](https://docs.gitlab.com/ee/api/discussions.html#list-project-issue-discussion-items) doesn't distinguish between comments and threads. You will always get all comments, wrapped in a discussion object from that API.

orta:
Great, this looks good to me - I've given it a once over and will give it a second look and then merge soon!

orta (approved):
Nice!

orta:
Thanks for the substantial PR! This has shipped as danger 11.2.0

squarefrog:
This PR has broken thread reporting for us, using GitLab 15.3. Using danger 11.1.4 we still get both inline and global warnings.

Using danger 11.2.0 or later, we only get global warnings, as they replace the body of the inline warnings. Full context available in https://github.com/danger/danger-js/issues/1351

samitha9125:
I second what @squarefrog  says. New feature delete old comments/threads in a MR as well when enabled.

PR #1337, opened 2022-11-11
Moving the GITHUB_WORKFLOW check before actually getting the userInfo

stodirascu:
Because on Github Actions runs, the default `secrets.GITHUB_TOKEN` doesn't have user privileges and the workflow fails with a 403 error.


orta:
Happy PR 1337, looks like a good catch to me!

stodirascu:
Glad I could contribute!
Do you know when's the next release planned, @orta?

PR #1342, opened 2022-12-19
Support arm64 binary generation

pepix:
# What was the problem?

- Currently several `danger-swift` users (especially who use Apple silicon mac) get errors when they use danger in their project.
  - https://github.com/danger/swift/issues/464

## Before fix

error

# What I did

- Separate compile command into two. One is for x86_64 users (Intel Mac) and another one is for arm64 users (ex. M1, M1 Pro, etc...)
- 🆕 Add a new workflow to build macOS binaries for all architectures on GitHub Actions

## After fix

pass

# Benefit

- Apple silicon users can install their suitable `danger` by only `brew install danger/tap/danger-swift` command as before without any concerns.

# Next actions

- 🆕 Configure `Actions Secret` and `Deploy Key`
  1. Generate SSH key with RSA algorithm using `keygen` on your terminal. Leave passphrase empty.
  2. Add public key to **deploy keys** in [`homebrew-tap`](https://github.com/danger/homebrew-tap) repository.
  3. Add secret key to **Actions secrets** in [`danger-js`](https://github.com/danger/danger-js) repository. Key name must be **HOMEBREW_TAP_DEPLOY_SECRET_KEY**.
- ~~Release operator has to run `yarn package:x64` and `yarn package:arm64` when releasing new version from now on. And publish both.~~
  - 🆕 Just run `$ gh workflow run .github/workflows/release.yml version_digit={major | minor | patch}` (select update version type) on your terminal if you have GitHub CLI. Or execute on GitHub Actions dashboard page.


orta:
Nice thinking! 

The deploy process handles the homebrew-tap here: https://github.com/danger/danger-js/blob/main/scripts/create-homebrew-tap-pr.sh

And https://github.com/danger/danger-js/blob/main/.release-it.json#L9 will need to be tweaked to generate both

That should be it though IMO

pepix:
Hello! @orta 
Before updating, I realized it seems only arm64 mac can make package for both architectures. However x64 mac cannot.
I'm not very sure current release operation though, are you sure release operator have arm64 mac or release pipeline can build for both?
If no, we have to prepare workaround before merge.
Updating GitHub Actions could be enough I believe.

pepix:
Hi! @orta 
I've added a new GitHub workflow to build packages for all macOS architectures on cloud.
Once you setup SSH keys, only one command delivers new release even you don't have arm64 mac.

Any feedbacks and opinions are welcome :)

orta:
Alright, well this looks great to me - I was a little wary because ~250 people have write access to the org, but they already have write access to the homebrew parts of it anyway and so changing just this part of the process doesn't actually change anything fundementally. 

So, I've given it all a once over and it looks good to me - later this evening I'll try ship the latest build of danger with it- thanks!

orta:
OK, so there's a lot going on here - I'm going to need to tweak this a bit to let the npm parts of the process work locally on my device, and then the github release + homebrew stuff happen in GH actions

orta:
That said, thanks a lot - this must have been _quite_ time consuming, and probably required learning a bunch of new things!

orta:
Alright, I've got this process up and running:
- https://github.com/danger/danger-js/releases/tag/11.2.1
- https://github.com/danger/homebrew-tap/commit/b22ae31bd49e8fe59881d6d6cd1f8536ff8491e5

It _looks_ correct to me!

orta:
Thanks! I was keeping an intel mac around (I use linux as my main daily machine) specifically for Danger deploys, this means a deploy for me is now trivial and doesn't require pulling out an old computer and doing a bunch of work.

pepix:
I'm glad that I could alleviate your loads by my commit.
Thank you for supporting npm part and others. :)

Now I confirmed `brew install danger/tap/danger-swift` and `danger-swift pr https://github.com/danger/swift/pull/146` works without error on M1 mac.

Issue #1314, opened 2022-09-25
[BUG] addLabels not a function


ivankatliarchuk:
**Describe the bug**

Probably not a bug, I might be missing something.
According to this [PR](https://github.com/danger/danger-js/pull/1139), following should just work
```
// Add labels
danger.gitlab.api.addLabels('Label 1', 'Label 2');
```

However its not a case.

**To Reproduce**
```js
const { danger } = require('danger');
danger.gitlab.api.addLabels('Label 1', 'Label 2');
```

`TypeError: addLabels is not a function at beforeCompletion (dangerfile.js:2:9)`

For ts, same result
```ts
import { danger } from "danger";
danger.gitlab.api.addLabels('Label 1', 'Label 2');
```

`TypeError: addLabels is not a function`

**Expected behavior**
It should update labels.

Current approach
```js
const addLabels = async (prId, input) => {
  const mrLabels = danger.gitlab.mr.labels
  const changes = [...mrLabels, ...input.filter(el => !mrLabels.includes(el, 0))];
  if (mrLabels.length != changes.length) {
    const mr = await danger.gitlab.api.MergeRequests.edit(repo, prId, { labels: [...mrLabels, ...input] })
    console.log(`updated labels '${changes}' for mr '${prId}'`)
  }
}
```

Labels updated.

**Screenshots**
![Screenshot 2022-09-25 at 13 21 41](https://user-images.githubusercontent.com/5395690/192148400-a3bf87c7-df8c-4ce0-a4b7-91a409e89dd3.png)

**Your Environment**

 | software      | version
| -------------| -------
| danger.js      | 11.1.2
| node             | v14.16.1) and v18.9.0 
| npm              | (npm@6.14.12 and (npm v8.19.1)
| Operating System | Macos and Ubuntu(Docker)

orta:
I don't think those are on that object (I think the PR had the wrong text) try without the `.api` (or check the typescript types

ivankatliarchuk:
I went through the code. 
`danger.gitlab.api` actually returns `InstanceType<typeof Gitlab>`. Not sure if was it intentional or not, but as a result, functions `addLabels` and `removeLabels` not used anywhere in the library, would assume not in use outide of it too.

Does it sound allright if I make them available through `danger.gitlab.uits.addLabels` ?

orta:
I think it is available as `danger.gitlab.addLabels` judging on the code

ivankatliarchuk:
```
[...]

PR #1353, opened 2023-02-07
Feature: Expose addLabels, removeLabels via gitlab.utils

glensc:
Fixes https://github.com/danger/danger-js/issues/1314


glensc:
@orta: This is an immediate fix for https://github.com/danger/danger-js/issues/1314
but I really think the way to go is with https://github.com/danger/danger-js/pull/1354, i.e restore the breakage.

There are more notes and musings at:
- https://github.com/danger/danger-js/issues/1314

orta:
Yeah, agree, these should live on the utils section of the gitlab API - as that mirrors the way they are on the github API

PR #1366, opened 2023-03-02
Allow parsing remote files as typescript

snowe2010:
Currently, the transpiler.ts makes an assumption that the filename will always end with .ts, but this is incorrect in the case of remote dangerfiles that have refs attached, for example `abc/123/dangerfile.ts?feature/branch`
Due to this, typescript parsing will always fail for remote files.

This commit adds a new parameter to the `transpiler` function, which parses the filename differently if the file is a remote one. This change also had to cascade up the stack to the `runDangerfileEnvironment` function, where we now accept a set of tuples, where the first is the filename, and the second is if it is remote or not. The only place the remote flag is being set currently is in the `GitHub` runner, where we download the remote dangerfile.

Finally, the pr subcommand failed to use the remote feature correctly, since the file would never exist when the command is first run. The check for the file has been removed, simply leaving behind the same check that is everywhere else.

Fixes: https://github.com/danger/danger-js/discussions/1359


orta:
This makes sense WRT #1359 - when we used remote dangerfiles in Artsy it always was though Peril's runtime and so this flow largely would have been quick verifications from me for others. Great spot!

orta:
Thanks, this is shipped as 11.2.4!

snowe2010:
Wow, I didn't expect a merge first try haha! Thank you very much!

Issue #1380, opened 2023-04-21
[BUG] Inline comments deleted in GitLab pipeline


Xiaobo2020:
**Describe the bug**

I need both normal comment and maybe inline comment.

**To Reproduce**
1. write an inline comment message in dangerfile.js. e.g. `message("inline comment", "package.json", 1)`
2. write a normal comment message in dangerfile.js. e.g. `message("normal comment")`
3. make a merge request in Gitlab and let pipeline run
4. only show inline comment or normal comment.

**Expected behavior**

Both comments should display

**Screenshots**

<img width="495" alt="image" src="https://user-images.githubusercontent.com/16217316/233684672-71358c6f-f87a-4182-a7d9-5d926996f619.png">

**Your Environment**

 | software         | version |
| ---------------- | -------|
| danger.js          | `11.2.6` |
| node             | `16.14.0` |
| npm              | `8.3.1` |
| Operating System | `MacOS` |

orta:
I dont know if gitlab have the same constraints as github - but inline comments there can only be added to files in the diff. So if the package.json line one is not changed, then it would show up in the main body

Xiaobo2020:
Hi @orta . 
Here you can think that there are some changes in package.json.  So theoretically inline comment should show up.
But I found some instructions about Danger in gitlab. It is mentioned that danger will only post one comment, I am not sure if this is related to the problem I am encountering.
https://docs.gitlab.com/ee/development/dangerbot.html#danger-comments-in-merge-requests

orta:
That's Danger in Ruby, so it has different constraints. Without being able to show a repro, there's not much I can tell you - I'd recommend running with `DEBUG=true` or `-verbose` and looking into what its sending then sending us a PR with what you think should fix it

Xiaobo2020:
I try to run `DEBUG='*' npx danger ci -v -f` in my gitlab pipeline. Through the printed information, I did find a strange phenomenon. It seems that danger did create an inline comment through createMergeRequestDiscussion, but then the inline comment was deleted.

| Create | Delete |
|--------|--------|
| <img width="664" alt="image" src="https://user-images.githubusercontent.com/16217316/233771355-c136c0b8-b405-46e3-8297-5a65f997bef7.png"> | <img width="939" alt="image" src="https://user-images.githubusercontent.com/16217316/233771375-46a2d963-2bff-482f-9ad5-929c2e76c8d6.png"> |

This is my dangerfile.js and I'm sure there are some diff in both file.

[...]

PR #1381, opened 2023-04-24
Fix: parameters of createComment in Executor.ts

Xiaobo2020:
Related issue: https://github.com/danger/danger-js/issues/1380


fbartho (changes requested):
This PR is a breaking change to the whole platform API.

Instead how about you fix the Gitlab-only-issue by making changes inside the `GitLab::createComment`?

fbartho:
https://github.com/danger/danger-js/blob/899a8200c0781b16d740c2d68a15e64530f6ab56/source/platforms/GitLab.ts#L126

glensc:
Re-created this PR using review notes:
- https://github.com/danger/danger-js/pull/1395

PR #1388, opened 2023-07-29
add build and push to ghcr similar to danger

wilkyd:
Copy of https://github.com/danger/danger/pull/1445 & 1446 fix for JS

### DIFF
--- a/.github/workflows/publish_package.yml
+++ b/.github/workflows/publish_package.yml
@@ -2,27 +2,46 @@ name: Release Danger-JS package
 
 on:
   push:
+    branches:
+      - 'master'
     tags:
       - '*'
 
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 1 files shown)


orta:
Thanks

orta:
I'm going to revert this as it is a breaking change - we were shipping to docker before and this removes that feature for people who were using it 

I'm open to a version which ships to both, but not breaking folks' existing workflows

wilkyd:
Not sure I understand, both the original workflow and the updated workflow seem to solely push the Docker image to the GitHub Container Registry (ghcr.io). I don't see any push to Docker Hub or other registries. Could you please specify the breaking change? Thanks

.github/workflows/publish_package.yml
```diff
@@ -2,27 +2,46 @@ name: Release Danger-JS package
 
 on:
   push:
+    branches:
+      - 'master'
```
glensc, line 6: the branch is wrong., this repo has "main" as the default branch

glensc: proposed fix: https://github.com/danger/danger-js/pull/1396

orta:
Hah, sorry,  I misread the older build logs!

PR #1392, opened 2023-07-30
Refactor: Put long node error inside html details

glensc:
when node process errors, that huge dump of details is placed as a note. put it into HTML `<details>` so the code block is collapsed by default.

image


glensc:
CI Fix:
- https://github.com/danger/danger-js/pull/1393

orta:
Cool, yeah, this seems reasonable to me 👍🏻

PR #1395, opened 2023-07-30
Add dangerID to gitlab createComment api

glensc:
Alternative fix https://github.com/danger/danger-js/pull/1381 based on review notes:
- https://github.com/danger/danger-js/pull/1381#pullrequestreview-1434996715

Fixes https://github.com/danger/danger-js/issues/1380#issuecomment-1519375625
Replaces https://github.com/danger/danger-js/issues/1380

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 ## Main
 
 <!-- Your comment below this -->
+- Add dangerID to gitlab createComment api #1395 [@glensc]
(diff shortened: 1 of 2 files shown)


orta:
I see, so the issue was that the args passed into it didn't match where it is used somewhere. It doens't match  how the GitHub platform works: `createComment: (comment: string) => api.postPRComment(comment),` but I'll assume that this is right as a few folks agree

PR #1397, opened 2023-08-01
feat: add shareable tsconfig support

p-chan:
Fix https://github.com/danger/danger-js/issues/1283


source/runner/runners/utils/transpiler.ts
```diff
@@ -126,7 +126,7 @@ export const typescriptify = (content: string, dir: string): string => {
   let compilerOptions: any
   const tsConfigPath = lookupTSConfig(dir)
   if (tsConfigPath) {
-    compilerOptions = JSON5.parse(fs.readFileSync(tsConfigPath, "utf8"))
+    compilerOptions = parseTsconfig(tsConfigPath)
```
fbartho, line 129: This project already depends on TypeScript (see line `123`), so it’s not helpful for us to pull in a 3rd-party package whose purpose is to `parseTsconfig` without a dependency on TypeScript!

Instead, I’d accept a PR that uses`ts.parseConfig` or a similar API that supports `extends` from the official dependency — that we’re already pulling in.

fbartho (changes requested):
No objection for this feature in concept, let’s just not pull in a new dependency for it.

Thanks!

fbartho:
This PR looks like it’s a duplicate to #1313 — and that one has tests and an approach that better matches my feedback above.

Not sure why that one stalled out though! So you might be able to pick up from where they started, or at least use it as a hint.

PR #1365, opened 2023-02-27
Feature: Add support for BitBucket Repository Access Token

thawankeane:
BitBucket supports [repositories access tokens](https://support.atlassian.com/bitbucket-cloud/docs/repository-access-tokens/), these tokens can access the repository through scopes, it can be used to authenticate with BitBucket REST API to make changes to repo itself.

This pull request add support for them through a new environment variable called `DANGER_BITBUCKETCLOUD_REPO_ACCESSTOKEN`

PR #1398, opened 2023-08-08
Fix: Use proper env values for Bitbucket Cloud Access Token

shitamori1272:
Thanks to #1365, BitBucket Cloud Access Token has been supported with env `DANGER_BITBUCKETCLOUD_REPO_ACCESSTOKEN`.

However, this feature doesn't work currently because of the following issues.

#### 1. Some logics uses the wrong env name `DANGER_BITBUCKETCLOUD_REPOSITORY_ACCESSTOKEN`. 
Because of this issue, need to set both `DANGER_BITBUCKETCLOUD_REPO_ACCESSTOKEN` and `DANGER_BITBUCKETCLOUD_REPOSITORY_ACCESSTOKEN` to use this feature.
To fix this, changed the env name in the code.

#### 2. Fails to fetch account UUID due to permission limitation.

According to https://support.atlassian.com/bitbucket-cloud/docs/repository-access-token-permissions/, Bitbucket Repository Access Token doesn't have permission to access account information which needs to fetch account UUID.
And this issue causes the following error.
```
Request failed [401]: https://api.bitbucket.org/2.0/user. Attempting retry.
Retry 1 of 3.
Request failed [401]: https://api.bitbucket.org/2.0/user. Attempting retry.
Retry 2 of 3.
Request failed [401]: https://api.bitbucket.org/2.0/user. Attempting retry.
Retry 3 of 3.
Request failed [401]: https://api.bitbucket.org/2.0/user
Response: {
  "type": "error",
  "error": {
    "message": "Token is invalid or not supported for this endpoint."
  }
}
Error:  Error: 401 - Unauthorized (Have you allowed permission 'account' for this credential?)
```

To avoid this error, allowed to use `DANGER_BITBUCKETCLOUD_UUID` env with Access Token.


fbartho (approved):
Good catch, this was a clear typo in the original PR. I have no objections to this change.

orta:
Agree 👍🏻

PR #1404, opened 2023-09-10
Customise GHE actions bot user ID

dimitar-hristov:
# Context
- In GitHub actions, the user ID of the bot is [hardcoded for github.com](https://github.com/danger/danger-js/blob/029fb4f0ad4c1ba9a56cc596057abdeb1df671b2/source/platforms/github/GitHubAPI.ts#L127-L144) without an option to overwrite it for GitHub Enterprise. As a result when called [here](https://github.com/danger/danger-js/blob/029fb4f0ad4c1ba9a56cc596057abdeb1df671b2/source/platforms/github/GitHubAPI.ts#L86) and [here](https://github.com/danger/danger-js/blob/029fb4f0ad4c1ba9a56cc596057abdeb1df671b2/source/platforms/github/GitHubAPI.ts#L315) to update the PR comments, it fails to find the comments.

# Changes
- Adds logic to customise `DANGER_GHE_ACTIONS_BOT_USER_ID `.

### DIFF
--- a/source/platforms/github/GitHubAPI.ts
+++ b/source/platforms/github/GitHubAPI.ts
@@ -150,7 +150,7 @@ export class GitHubAPI {
       return 41898282
     }
 
-    this.d("Danger used ID is undefined.")
+    this.d("Danger user ID is undefined.")
     return undefined
   }
 
--- a/source/platforms/github/GitHubAPI.ts
+++ b/source/platforms/github/GitHubAPI.ts
@@ -140,7 +140,7 @@ export class GitHubAPI {
     const useGitHubActionsID = process.env["GITHUB_WORKFLOW"]
     if (useGitHubActionsID) {
       // Allow to customise the GitHub actions app ID for Github Enterprise
-      const gheActionsID = process.env["GHE_ACTIONS_BOT_USER_ID"]
+      const gheActionsID = process.env["DANGER_GHE_ACTIONS_BOT_USER_ID"]
       if (gheActionsID) {
         return parseInt(gheActionsID)
       }
--- a/source/platforms/github/GitHubAPI.ts
+++ b/source/platforms/github/GitHubAPI.ts
@@ -130,17 +130,22 @@ export class GitHubAPI {
       return parseInt(perilID)
     }
 
-    const info = await this.getUserInfo()
-    if (info.id) {
-      return info.id
-    }
-
     const useGitHubActionsID = process.env["GITHUB_WORKFLOW"]
     if (useGitHubActionsID) {
+      // Allow to customise the GitHub actions app ID for Github Enterprise
+      const gheActionsID = process.env["GHE_ACTIONS_BOT_USER_ID"]
+      if (gheActionsID) {
+        return parseInt(gheActionsID)
+      }
       // This is the user.id of the github-actions app (https://github.com/apps/github-actions)
       // that is used to comment when using danger in a GitHub Action
       // with GITHUB_TOKEN (https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token)
       return 41898282
+    }
+
+    const info = await this.getUserInfo()
+    if (info.id) {
+      return info.id
     }
 
     return undefined
--- a/source/platforms/github/GitHubAPI.ts
+++ b/source/platforms/github/GitHubAPI.ts
@@ -130,11 +130,9 @@ export class GitHubAPI {
       return parseInt(perilID)
     }
 
-    if (process.env["DANGER_GITHUB_API_TOKEN"]) {
-      const info = await this.getUserInfo()
-      if (info.id) {
-        return info.id
-      }
+    const info = await this.getUserInfo()
+    if (info.id) {
+      return info.id
     }
 
     const useGitHubActionsID = process.env["GITHUB_WORKFLOW"]
--- a/source/platforms/github/GitHubAPI.ts
+++ b/source/platforms/github/GitHubAPI.ts
@@ -130,6 +130,13 @@ export class GitHubAPI {
       return parseInt(perilID)
     }
 
+    if (process.env["DANGER_GITHUB_API_TOKEN"]) {
+      const info = await this.getUserInfo()
+      if (info.id) {
+        return info.id
+      }
+    }
+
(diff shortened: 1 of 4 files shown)


dimitar-hristov:
Hey @orta 👋 When you have a moment, could you review this PR, please?

source/platforms/github/GitHubAPI.ts
```diff
@@ -130,19 +130,24 @@ export class GitHubAPI {
...
+      if (gheActionsID) {
+        return parseInt(gheActionsID)
+      }
       // This is the user.id of the github-actions app (https://github.com/apps/github-actions)
       // that is used to comment when using danger in a GitHub Action
       // with GITHUB_TOKEN (https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token)
       return 41898282
     }
 
+    const info = await this.getUserInfo()
```
fbartho, line 146: This reorganization will cause a breaking behavior change even if `GHE_ACTIONS_BOT_USER_ID` is unset.

Originally, if `this.getUserInfo` returns a logged in user then it would immediately return with `info.id`. Now, it will *first* check if `GITHUB_WORKFLOW` is set (and it’s always set on Github, I believe), and so it will always return a hardcoded ID

This would break our codebase, as we set an env-var, I believe `DANGER_GITHUB_TOKEN` or similar (not at my desk) to log in to a bot account with a PAT. This code-change would only allow comments from the GitHub-actions app-id of `41898282`.

source/platforms/github/_tests/_github_api.test.ts
```diff
@@ -363,9 +363,26 @@ describe("Peril", () => {
...
     })
 
-    it("Makes getUserId return undefined", async () => {
+    it("Makes getUserId return PERIL_BOT_USER_ID", async () => {
       const userID = await api.getUserID()
       expect(userID).toBe(1)
     })
   })
+
+  describe("Allows setting GHE_ACTIONS_BOT_USER_ID env variable", () => {
```
fbartho, line 372: `GHE_ACTIONS_BOT_USER_ID` — this is a custom environment variable, right? (not set by github, set by the user).

The naming convention for DangerJS custom environment variables is to prefix it with `DANGER_`, so in this case, I would name it something like `DANGER_GHE_ACTIONS_BOT_USER_ID`

source/platforms/github/_tests/_github_api.test.ts
```diff
@@ -363,9 +363,26 @@ describe("Peril", () => {
       delete process.env.PERIL_BOT_USER_ID
     })
 
-    it("Makes getUserId return undefined", async () => {
+    it("Makes getUserId return PERIL_BOT_USER_ID", async () => {
```
fbartho, line 366: I believe shouldn’t always return `PERIL_BOT_USER` here, instead repeat this test case:

1. A test case that matches the old test name where all vars are unset, `GITHUB_WORKFLOW, GHE_ACTIONS_BOT_USER_ID, PERIL_BOT_USER = undefined`
2. `PERIL_BOT_USER` if nobody is logged in (`GHE_ACTIONS_BOT_USER_ID` is unset, but `GITHUB_WORKFLOW` is set, and `PERIL_BOT_USER` is set)
3. `41898282` if `PERIL_BOT_USER` is unset, `GHE_ACTIONS_BOT_USER_ID` is unset, but `GITHUB_WORKFLOW` is set.
4.  The test case you’ve already added below`"Allows setting GHE_ACTIONS_BOT_USER_ID env variable"`

source/platforms/github/_tests/_github_api.test.ts
fbartho: Also, of course, once we need to document this env-var in appropriate places. Probably with a section for “GitHub Enterprise”

fbartho (changes requested):
No objections to the purpose of this PR — just gotta make sure it doesn’t break :P — Happy to take another look once you’ve ensured it won’t!

source/platforms/github/GitHubAPI.ts
dimitar-hristov: I see, now it makes more sense 😄 I think I found the env var you referred to - it seems to be `DANGER_GITHUB_API_TOKEN`.

I reverted the change, but also added a logic around it to ensure we call the GitHub API for the user only if the `DANGER_GITHUB_API_TOKEN` is set. Otherwise, I am not sure if we want to make the call, when there isn't a configured user. Do you think that makes sense?

source/platforms/github/_tests/_github_api.test.ts
dimitar-hristov: Yep makes sense. I have update the name of the env variable and added some docs. Please let me know if I missed something.

source/platforms/github/_tests/_github_api.test.ts
dimitar-hristov: I think the tests should now be updated to cover all scenarios.

dimitar-hristov:
> No objections to the purpose of this PR — just gotta make sure it doesn’t break :P — Happy to take another look once you’ve ensured it won’t!

Thanks for the quick response, @fbartho! I think all comments should now be addressed. Please have a look at the PR again when you have a moment and let me know if it is better. 🙂

source/platforms/github/GitHubAPI.ts
fbartho: Unfortunately, the regular `GITHUB_TOKEN` might be a valid logged in user so it’s not sufficient to only focuse on `DANGER_GITHUB_API_TOKEN`, I think! — As I understand it, that’s why we call the `getUserInfo` API — so we figure out the permissions on the attached token — figure out if it’s a real user or what exactly.

dimitar-hristov: Hmm, ok, I reverted that part of my change. I still did the tests for completeness, but i don't think they are relevant to my change anymore. 

The test that check if the configured token is a user is a bit weird, but I think it makes sense as it check that we have called the API with the correct creds (the ones used when creating the GitHub instance). And ensure that it correctly returns the mocked user ID.

Let me know what you think.

orta:
I think this is good to go - thanks both of you 👍🏻

PR #1410, opened 2023-11-01
fix(javadoc): fail and warn functions fix misleading documentation

Vacxe:
References:
https://github.com/danger/danger-js/blob/fdc1c49aff5e569580a23d0547ea0442a753826c/source/commands/utils/sharedDangerfileArgs.ts#L52
https://github.com/danger/danger-js/blob/fdc1c49aff5e569580a23d0547ea0442a753826c/source/runner/Executor.ts#L156

Behaviour:
By default `failOnErrors` is `false` and will not fail run in case of any fails will be found. In this case current documentation is a bit misleading because users can expect different behaviour.

### DIFF
--- a/source/runner/Dangerfile.ts
+++ b/source/runner/Dangerfile.ts
@@ -27,7 +27,7 @@ export interface DangerContext {
   schedule(asyncFunction: Scheduleable): void
 
   /**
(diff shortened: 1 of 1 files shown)


orta:
Cool, this seems reasonable to me - thanks!

PR #1411, opened 2023-11-02
Patch 1

normano64:
I noticed that the danger step in CI was always taking more than 10 seconds, the root cause was that nodeCleanup was waiting for the missing DSL timeout before continuing even when the DSL was found

### DIFF
--- a/source/commands/danger-runner.ts
+++ b/source/commands/danger-runner.ts
@@ -51,6 +51,7 @@ const run = (config: SharedCLI) => async (jsonString: string) => {
 
   d("Got STDIN for Danger Run")
   foundDSL = true
+  clearTimeout(missingDSLTimeout)
   const dangerFile = dangerfilePath(program)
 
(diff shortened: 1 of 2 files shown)


orta:
This makes sense to me 👍🏻  - great find

source/commands/danger-runner.ts
```diff
@@ -85,9 +86,9 @@ nodeCleanup((exitCode: number, signal: string) => {
 })
 
 // Add a timeout so that CI doesn't run forever if something has broken.
-setTimeout(() => {
+const missingDSLTimeout = setTimeout(() => {
   if (!foundDSL) {
-    console.error(chalk.red("Timeout: Failed to get the Danger DSL after 1 second"))
+    console.error(chalk.red("Timeout: Failed to get the Danger DSL after 10 second"))
```
glensc, line 91: spelling: "seconds" now

Issue #1412, opened 2023-11-04
[BUG] Danger fails to create inline comments on Gitlab


dimsuz:
**Describe the bug**
I am using self-hosted gitlab and doing this in dangerfile.ts:

```typescript
message("Add something please", "dangerfile.ts", 10)
```
Causes this error in logs (I rerun with debug logs enabled):
```
2023-11-04T12:59:01.986Z danger:GitLabAPI createMergeRequestDiscussion myapp/myapp-android 34
<!--
  0 failure:
  0 warning:
  1 messages

  DangerID: danger-id-Danger;
  File: dangerfile.ts;
  Line: 10;
-->
- :book: Add something please
   {
  position: {
    position_type: 'text',
    base_sha: '4262b65dde38269fbfa91f1ddeacc4791e742414',
    start_sha: '4262b65dde38269fbfa91f1ddeacc4791e742414',
    head_sha: '3d5c791cb2adc7f7e8314e6b77cbf36a339c6796',
    old_path: 'dangerfile.ts',
    old_line: undefined,
    new_path: 'dangerfile.ts',
    new_line: 10
  }
}
2023-11-04T12:59:01.988Z danger:GitLabAPI createMergeRequestDiscussion TypeError: Cannot read properties of undefined (reading 'name')
    at FormData._getContentDisposition (/node_modules/form-data/lib/form_data.js:227:40)
    at FormData._multiPartHeader (/node_modules/form-data/lib/form_data.js:178:33)
    at FormData.append (/node_modules/form-data/lib/form_data.js:71:21)
    at /node_modules/@gitbeaker/core/dist/index.js:168:18
    at Array.forEach (<anonymous>)
    at appendFormFromObject (/node_modules/@gitbeaker/core/dist/index.js:163:28)
    at Object.<anonymous> (/node_modules/@gitbeaker/core/dist/index.js:257:41)
    at step (/node_modules/@gitbeaker/core/dist/index.js:121:23)
    at Object.next (/node_modules/@gitbeaker/core/dist/index.js:102:53)
    at /node_modules/@gitbeaker/core/dist/index.js:95:71
```

At the same time, a "global" table-comment is posted ok to the "Overview" section of the MR.
Access token has `api, read_repository, read_user` access rights. `api` one alone should be enough to post comments IIRC.

I'd be happy to help debug this further if needed.

**Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          | 11.3.0
| node             | v18.13.0
[...]

Issue #1405, opened 2023-09-14
[BUG] [Gitlab] Can't post multiple inline comments


havebeenfitz:
**Describe the bug**

**To Reproduce**
1. Add 2+ messages with file and line to Dangerfile.swift
2. Run danger-swift ci on gitlab instance

**Expected behavior**
The first inline comment is created, then overriden by the second as a general comment

**Screenshots**

**Your Environment**
danger-swift 3.17.1

 | software         | version
| ---------------- | -------
| installation method | homebrew |
 Operating System | macOS

**Additional context**
As I understand there is a connected [issue](https://github.com/danger/danger-js/pull/1176/) for Github. Is it possible to adopt the same fix for Gitlab or is it completely different?

fbartho:
> As I understand there is a connected [issue](https://github.com/danger/danger-js/pull/1176/) for Github. Is it possible to adopt the same fix for Gitlab or is it completely different?

I would certainly examine the linked PR to see if it teaches you anything, but I wouldn’t expect it to necessarily be exactly related. The fundamental of cross-provider DangerJS behavior has been pretty stable for a long time. GitLab is a relatively newer provider so I’d expect the issue to be in one of these places:

- https://github.com/danger/danger-js/blob/main/source/platforms/GitLab.ts
- https://github.com/danger/danger-js/blob/main/source/platforms/gitlab/GitLabAPI.ts

I think more likely, the problem is a misunderstanding about GitLab’s automatic behaviors, for example maybe GitLab has  only one “slot” for a line-level comment, which means the last one always wins. In that case, you might be able to preprocess all the line-level comments, and merge all the ones that target the same/file-line into a single comment. (I even believe this is doable in user-space)

Issue #1386, opened 2023-07-06
[DEPS] Move from @gitbeaker/node to @gitbeaker/rest


nostrorom:
**Describe the bug**

Getting the warning when installing `danger` via `npm`

```bash
  npm WARN deprecated @gitbeaker/node@21.7.0: Please use its successor @gitbeaker/rest
```

**Your Environment**
Gitlab CI

| software         | version
| ---------------- | -------
| danger.js        |  11.2.6
| node             | 18.16.1
| npm              |   9.5.1

**Additional context**
[Similar issue about upgrading](https://github.com/danger/danger-js/pull/1327)

orta:
Sure, you're welcome to give it a look 👍🏻

glensc:
I guess first need problems solved in previous update:
- https://github.com/danger/danger-js/pull/1327

sakgoyal:
any update on this?

heltoft:
@orta We can close this one as we did move to `gitbeaker/rest` with above merges.

glensc:
@nostrorom, @orta, this can be closed

```
CHANGELOG.md:- GitLab: [#1386] Move from `@gitbeaker/node` to `@gitbeaker/rest` [@heltoft]
CHANGELOG.md:- GitLab: Upgrade `@gitbeaker/node` from `^21.3.0` to `^35.8.1` [@buffcode]
```

PR #1424, opened 2024-02-05
Upgrade node engine to >=18 & @types/node to ^18.19.14

heltoft:
The currently used node environment has reached end of life. 
Updating to 18.x which is the version with Maintenance status for Node.js.

- Upgraded node engine to >= 18
- Upgraded @types/node to ^18.19.14

This change is a prerequisite to start solving some of the current GitLab issues, especially in regards to problems with inline comments which needs dependency updates.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -7,10 +7,10 @@ jobs:
 
     steps:
       # Check out, and set up the node/ruby infra
-      - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
         with:
-          node-version: "16"
+          node-version: "18"
 
       # Get local dependencies & test
       - run: yarn install
--- a/.github/workflows/publish_package.yml
+++ b/.github/workflows/publish_package.yml
@@ -24,7 +24,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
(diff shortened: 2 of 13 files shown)


orta:
I'm no onboard with forcing all users of danger to need node 18+ in order to get updates. I have no problem with making that the case for contributing to this repo though!

heltoft:
Hey @orta, do I understand you correctly that you are ok with updating if there is some contributions to go along with the update? If so, then I can assure you there is more changes coming 🚀 

I have been working on getting inline comments to work on GitLab recently in my spare time.
Mainly focused around the following issues:
#1386 
#1412
#1405

I have a version of danger-js running where these issues are solved that we have been using for a few weeks where I work, and it has been working well for us.

However it is quite a lot of changes so what I am trying to do is to split up the changes into smaller chunks 🙏 

How would you suggest for me to proceed in this case?

orta:
Sure, 1 PR is reasonable for that scope

Is the underlaying issue for the engines change that the gitbreaker update is going to force us to make our consumers update also?

heltoft:
Alright sounds good. I have a few more tests to write, but I will create an MR with all the changes when I feel it is ready 🙂 

Yes gitbeaker requires node 18 since it leverages native fetch under the hood in newer versions. The documentation for gitbeaker states that it is possible to make it work with a polyfill for node 16.18+, however I haven't been able to make that work so far, and it would also require to use the `--ignore-engines` flag.
And if we are already bumping node engine I personally think it makes more sense to go to 18 as that is the last maintained version.

PR #1425, opened 2024-02-22
Use node18 for Dockerfile

th-le:
Some of the plugins e.g https://github.com/sogame/danger-plugin-toolbox have dropped support for node12 and its latest version requires node18. 

Updating the Dockerfile to newer node version allows the dependencies to be upgraded as well.


orta:
👍🏻

Issue #1421, opened 2024-01-24
[BUG] prettier.resolveConfig fails due to faulty shouldUseGitHubOverride


mgol:
**Describe the bug**
We have a Danger JS check verifying Prettier formatting of changed files. That check is quite complex and runs through multiple files but at one point it does:
```js
const fileInfoOptions = {
  ...(await prettier.resolveConfig(filePath))
};
```
This worked fine until we upgraded to Prettier `3.1.1` which is published as a mixed ESM/CJS module now, but even the CJS version mostly just re-exports wrapped ESM APIs.

After the upgrade, the Danger JS check fails with an error:
```
TypeError: Cannot read properties of undefined (reading 'filename')
    at shouldUseGitHubOverride (my-project/node_modules/danger/distribution/platforms/github/customGitHubRequire.js:131:16)
    at _module2.default._load (my-project/node_modules/override-require/dist/overrideRequire.js:38:9)
    at cjsLoader (node:internal/modules/esm/translators:356:17)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:305:7)
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:329:24)
    at async loadJs (file://my-project/node_modules/prettier/index.mjs:22668:18)
    at async loadConfig (file://my-project/node_modules/prettier/index.mjs:22726:16)
    at async loadPrettierConfig (file://my-project/node_modules/prettier/index.mjs:23032:18)
    at async Promise.all (index 0)
```
The error points to the `filename` access from `parent` in the `if` in this piece of code:
```js
var shouldUseGitHubOverride = function (request, parent) {
    // Is it a from a file we're handling, and is it relative?
    if (parent.filename.startsWith(exports.dangerPrefix) && request.startsWith(".")) {
        return true;
    }
    // Basically any import that's not a relative import from a Dangerfile
    return false;
};
```
As it turns out, `parent` can be undefined. Going through the Prettier-specific part of the stack, it crashes on the first line of the `loadJs` function:
```js
async function loadJs(file) {
  const module = await import(pathToFileURL4(file).href);
  return module.default;
}
```

If I change `parent.filename` in that `shouldUseGitHubOverride` util to an optional access `parent?.filename`, the check passes.

Going through the stack, the `shouldUseGitHubOverride` util is used by the `override-require` package in its override of the `_load` method of the `'module'` Node.js module as the `isOverride` check:
[...]

PR #1430, opened 2024-03-13
Fix the shouldUseGitHubOverride logic for ESM imports

mgol:
The `shouldUseGitHubOverride` util is called with `Module._load` parameters. So far, it assumed it will get a string `request` and a `NodeModule` `parent` parameters. However, for ESM imports `parent` is actually `undefined`.

Account for this in code by making `parent` an optional argument; add a test.

Fixes gh-1421

cc @fbartho

### DIFF
--- a/source/platforms/github/_tests/_customGitHubRequire.test.ts
+++ b/source/platforms/github/_tests/_customGitHubRequire.test.ts
@@ -31,6 +31,12 @@ describe("shouldUseGitHubOverride", () => {
     const parent: any = { filename: dangerPrefix + "./my-import" }
     expect(shouldUseGitHubOverride(module, parent)).toBeTruthy()
   })
+
+  it("ignores modules without a parent", () => {
+    const module = "./peril"
+    const parent: any = undefined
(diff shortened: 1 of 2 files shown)


fbartho (approved):
Looks good to me! Matches what we talked about!

source/platforms/github/_tests/_customGitHubRequire.test.ts
```diff
@@ -31,6 +31,12 @@ describe("shouldUseGitHubOverride", () => {
     const parent: any = { filename: dangerPrefix + "./my-import" }
     expect(shouldUseGitHubOverride(module, parent)).toBeTruthy()
   })
+
+  it("ignores modules without a parent", () => {
+    const module = "./peril"
+    const parent: any = undefined
```
mgol, line 37: I could avoid the type here - the following would work as well:
```suggestion
    const parent = undefined
```
but I used what's more consistent with the other test cases.

Let me know if you'd like to drop this `any`.

orta: This is fine :+

PR #1437, opened 2024-04-16
Replace DANGER_MANUAL_PR with DANGER_MANUAL_PR_NUM in changelog

7rulnik:
It took me a while to figure out why I was getting 327 instead of the value that I passed :D

https://github.com/danger/danger-js/blob/9d9bd6ead7b24b59de1607bd692141e5ccc76608/source/ci_source/providers/Fake.ts#L10

Also, I think a fake provider is worth documenting it, but it's another issue


orta:
:+1: - yeah, no idea where that number came from

Issue #1432, opened 2024-04-04
[BUG] GitHub API request fails for PR with > 300 files change


narutox900:
**Describe the bug**
We are using `danger@^11.2.7` on CircleCI.
Since a week before, we started noticing that even though `danger` shows a successful execution, the detail logs show that there's an error with GitHub API. Upon reading [GitHub API docs](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files), it states that the limit is 3000 files, however the error shows that the maximum number of files is 300. I haven't been able to find any changelogs related to this problem. Logs from earlier runs on large PRs show that `danger` can still run properly before

**To Reproduce**
1. Make a PR with changes in > 300 files
2. Run `danger` on CI
 

**Expected behavior**

- Returns a fail exit code if `danger` could not run properly.
- Maybe find a way to workaround this new limit

**Screenshots**
```
yarn run v1.19.1
$ /home/circleci/*********/node_modules/.bin/danger ci
Request failed [406]: https://api.github.com/repos/*********/*********/pulls/11777
Response: {
  "message": "Sorry, the diff exceeded the maximum number of files (300). Consider using 'List pull requests files' API or locally cloning the repository instead.",
  "errors": [
    {
      "resource": "PullRequest",
      "field": "diff",
      "code": "too_large"
    }
  ],
  "documentation_url": "https://docs.github.com/rest/pulls/pulls#list-pull-requests-files"
}
------------------------------------

------- Check SP tracking ----------
------- Note: this can log many errors in case of renamed files, but it can still pass the build -----
------------------------------------
------------------------------------
------- Check CI CLASS REMOVED -----
------- Note: this can log many errors in case of renamed files, but it can still pass the build -----
------------------------------------
------------------------------------
------- Check Holistics docs url ---
------------------------------------


Found only warnings, not failing the build.
Feedback: https://github.com/*********/*********/pull/11777#issuecomment-2030941585
Could not add a commit status, the GitHub token for Danger does not have access rights.
If the build fails, then danger will use a failing exit code.
Done in 26.62s.

CircleCI received exit code 0

```

**Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          | 11.2.7
| node             | 18.16.1
| npm              |
| Operating System | Ubuntu 20.04.6 LTS

**Additional context**
Related issues: https://github.com/danger/swift/issues/606, https://github.com/reviewdog/reviewdog/issues/1696

[...]

PR #1440, opened 2024-04-27
GitHub: use files API with pagination

fabianehlert:
Switched from using the GitHub API [Get a pull request](https://docs.github.com/en/rest/pulls#get-a-pull-request) to the [List pull requests files](https://docs.github.com/en/rest/pulls#list-pull-requests-files) API, incorporating pagination to retrieve pull request diffs. This change enables the analysis of pull requests containing more than 300 files.

Although the 300 file limit on the "Get a pull request" API isn't new, GitHub recently began enforcing it with a HTTP 406 status code. Currently, Danger is failing when analyzing pull requests exceeding this limit.

The "List pull requests files" API returns a slightly different git diff format, omitting the initial `diff --git …` part which `parse-diff` expects. As a _hack_, I'm adding it manually now.

Happy to hear your feedback on this approach :)

Fixes https://github.com/danger/danger-js/issues/1432 and https://github.com/danger/swift/issues/606

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 <!-- Your comment below this -->
 
 - Adds a log when you run on GitHub Actions without being a pull_request - [@orta]
+- GitHub: Move to 'List pull request files' API with pagination support [@fabianehlert]
 
 <!-- Your comment above this -->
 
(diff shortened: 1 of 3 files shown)


orta:
This seems reasonable to me 👍🏻

PR #1447, opened 2024-05-09
Add base layer dockerfile for common setup

glensc:
Don't repeat `node:18-slim`. Also fixes problem that labels were set for build layer, not the final image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM node:18-slim as build
+FROM node:18-slim as base
+FROM base as build
 
 LABEL maintainer="Orta Therox"
 LABEL "com.github.actions.name"="Danger JS Action"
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -15,9 +17,7 @@ RUN yarn remove 'typescript' --dev && yarn add 'typescript'
 RUN yarn install --production --frozen-lockfile
 RUN chmod +x distribution/commands/danger.js
 
-
-FROM node:18-slim
-WORKDIR /usr/src/danger
```
fbartho, line 20: Was deleting this `WORKDIR` line intentional?

glensc: it is not deleting, it's setup in the parent layer. see the full `Dockerfile` to see the context.

orta:
Cool, I'm no docker expert, but it seems reasonable enough to me

Issue #1448, opened 2024-05-10
[BUG] Error:  ReferenceError: fetch is not defined


squarefrog:
**Describe the bug**
After updating to Danger JS 12.x I'm unable to run Danger.

**To Reproduce**
1. Install danger through homebrew `brew install danger/tap/danger-js`
2. Try to run danger

**Expected behavior**
I expected Danger to work as I have a node version installed greater than the minimum.

**Screenshots**

```
Starting Danger PR on company/ios-app#5606
Error:  ReferenceError: fetch is not defined
    at defaultRequestHandler (/snapshot/danger-js/node_modules/@gitbeaker/rest/dist/index.js:101:22)
    at Object.requester.<computed> [as get] (/snapshot/danger-js/node_modules/@gitbeaker/requester-utils/dist/index.js:53:16)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at process.runNextTicks [as _tickCallback] (node:internal/process/task_queues:65:3)
    at Function.runMain (pkg/prelude/bootstrap.js:1980:13)
    at node:internal/main/run_main_module:17:47
    at async /snapshot/danger-js/node_modules/@gitbeaker/core/dist/index.js:105:22
```

**Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          | 12.2.0
| node             | v18.20.2, also tried v22.1.0
| npm              | 10.5.0
| Operating System | macOS 14.4.1

**Additional Context**
We run this in an iOS project without any JS dependencies, and I'm not too familiar with JS, so forgive me if this is a trivial issue to solve for an experienced JS engineer!

squarefrog:
If I run this through yarn it works fine, so I guess this could be something to do with installing through brew?

orta:
Yeah, brew must be referencing an old node version somehow

SvenMuc:
I had the same issue with danger-swift installed via brew.
Seems danger-swift uses danger-js under the hood.

https://github.com/danger/swift/issues/608

squarefrog:
Aye that’s right, it does. I do have node installed though. If I do `node —version` it reports v22 what is what I expect. 

Looking at the brew manifest, all it does is download the latest bin from here. Running that same bin works fine. 

I’m really not sure why installing it through brew uses an old version. All I can think is perhaps there’s something in my PATH that is getting intercepted early by the brew installation but not the manual one.

fbartho:
[...]

PR #1449, opened 2024-05-21
Use node 18 targets for yarn pkg

heltoft:
With the move to node 18 in v12.0.0 the targets for the yarn pkg command was not updated as expected. 

These commands are used by the homebrew-tap and fixes the issue of having an outdated node version as part of the compiled binaries when installing with brew as outlined in: https://github.com/danger/danger-js/issues/1448

This should suffice to fix the underlying issue, but I think it could be worth to look into the option of not bundling node in as part of the binary and instead use what is already available on the system it is running on.


fbartho (approved):
Ah! I’m not super familiar with yarn pkg, but this does seem to explain it!

Issue #1446, opened 2024-05-07
[BUG] all changes regardless of type are grouped incorrectly into `modified_files` in `danger.git`


xphir:
**Describe the bug**

Since version `12.0.1`, all changes (including deletions, creations, and modifications) are incorrectly grouped under `modified_files`. The `created_files` and `deleted_files` arrays remain empty, even when there are clearly file creations or deletions.

**To Reproduce**

1. On version `11.3.1`, create a `dangerfile.ts` as follows:
```
import { danger} from "danger"

console.log("Created", danger.git.created_files)
console.log("Deleted", danger.git.deleted_files)
console.log("Modified", danger.git.modified_files)
```
2. Run the command `npx danger pr https://github.com/danger/danger-js/pull/9`.
3. Observe the expected output:
```
Starting Danger PR on danger/danger-js#9
Created [ 'yarn.lock' ]
Deleted [ 'npm-shrinkwrap.json' ]
Modified [ 'README.md', 'package.json' ]

Danger: ✓ passed review, received no feedback.
```
4. Update the version to `12.0.1`
5. Run the same command: `npx danger pr https://github.com/danger/danger-js/pull/9`
6. Notice the unexpected and incorrect output:
```
Starting Danger PR on danger/danger-js#9
Created []
Deleted []
Modified [ 'README.md', 'npm-shrinkwrap.json', 'package.json', 'yarn.lock' ]

Danger: ✓ passed review, received no feedback.
```

**Expected behavior**
I would expect the behavior to remain consistent with version `11.3.1`, where deletions are grouped in `deleted_files`, creations in `created_files`, etc.

**Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          | `12.0.1`, `12.1.0` & `12.2.0`
| node             | `v20.12.1` & `v18.20.2`
| npm              | `10.5.0`
| Operating System | `MacOS Sonoma 14.4.1`

szweier:
I believe the issue stems from https://github.com/danger/danger-js/blob/bb0f4c2c40c9ab70b0f099c26a56f96bfaffeddb/source/platforms/git/diffToGitJSONDSL.ts#L14-L15 because the new response from /pulls/PR/files uses `removed` and `added` instead of `deleted` and `new`. I haven't been able to get danger-js to run locally and test but figure if someone can a PR could be opened fairly quickly. If I'm correct that is.

orta:
[...]

PR #1451, opened 2024-05-30
Add and use status on GitHubFile

szweier:
Attempt to fix https://github.com/danger/danger-js/issues/1446

By adding the file mode to the file patch the output from parseDiff now includes `new` and `deleted` boolean values correctly and give us something to validate against in  `diffToGitJSONDSL`.

New Output:

```
Starting Danger PR on danger/danger-js#9
Created [ 'yarn.lock' ]

Deleted [ 'npm-shrinkwrap.json' ]
Modified [ 'README.md', 'package.json' ]
```


orta:
I'm a little surprised this doesn't change any tests - but seems like the right thing to me

PR #1453, opened 2024-06-04
chore: update not PR condition for GH workflow

sayali10:
https://github.com/danger/danger-js/issues/1454


fbartho (approved):
I forgot about that event name. Necessary for fork-based PRs!

>This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.

From: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

orta:
Thanks!

Issue #1180, opened 2021-11-24
[Feature Request] ESM Support


fbartho:
I got tripped up by a dependency that went "esm-only" today in their 7.x/latest version [strip-ansi](https://github.com/chalk/strip-ansi#readme) [extra info from the maintainer](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).

Do we have a plan for ESM in DangerJS?

I'm not sure what this entails from DangerJS's perspective. I could write this off as an overly eager developer, but it smells like this is going to be an increasing question/stumbling block for people.

orta:
I wasn't planning on moving danger, as you can't `import "danger"` so there's no difference, maybe if it meant it could run in deno it could be worth it.

Though I think maybe supporting `dangerfile.mjs` or `dangerfile.mts` maybe might make sense.

fbartho:
Right, I’m much newer to understanding the practical consequences of `ESM` — I think your direct answers are useful, but slightly different from the crux of what I was getting at.

You said:
> so there's no difference

I probably should have specified: this ESM-only package could not be imported or used in my Dangerfile. It caused an error. It’s fine for now because I was able to downgrade, but that’s not a long-term solution. — I selected this library by scanning my yarn.lock for likely keywords so this library was already a transitive dependency.

_This means that now, I have a ticking time-bomb_, where this dependency has consciously required ESM, and all my other dependencies that rely on it are A) prevented from upgrading it or B) will be forced to switch to a different provider of the feature, or C) it’s going to spread like a virus, forcing them to go ESM-only.

Now maybe I’m totally confused, and there should be a way for DangerJS to import this ESM library — is that the case?

orta:
the dangerfile would be interpreted as either cjs or mjs based on the "type" in your package.json in node, or by the filetype (.cjs/.mjs) - its not on danger to do any different work if you're in an mjs or cjs package as that's still node responsibility, you just hit the timebomb in the same way you would if any of your (cjs) script files would try to import it

BrunnerLivio:
Similar issue here - the latest packages of remark now require ESM.
Running danger locally result in this message:

```
Starting Danger PR on roche-digital-platform/docs-portal/docs-content/docs-docs-portal#37
Unable to evaluate the Dangerfile
<PROJECT_PATH>/node_modules/remark/index.js:1
import {unified} from 'unified'
^^^^^^

SyntaxError: Cannot use import statement outside a module
[...]

PR #1456, opened 2024-06-19
fix(node): #1180 - Adjust tsconfig compiler options

matthewh:
Accidentally applied `es6` when compilerOptions.module was not defined.

I recognize posting an error message to indicate a successful fix isn't great practice but I'm testing locally and this at least proves the code transpiled correctly and executed and failed once danger realized it's not inside a github PR.

# "Success"

```
❯ pwd
/Volumes/Projects/multipart/eslint-plugin-jest
❯ danger local --base main --staging
...

Unable to evaluate the Dangerfile
 TypeError: Cannot read properties of  (reading 'pr')
    at Object.<anonymous> (dangerfile.ts:51:19)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    at requireFromString (/Volumes/Projects/multipart/danger-js/node_modules/require-from-string/index.js:28:4)
    at /Volumes/Projects/multipart/danger-js/distribution/runner/runners/inline.js:183:68
    at step (/Volumes/Projects/multipart/danger-js/distribution/runner/runners/inline.js:56:23)
    at Object.next (/Volumes/Projects/multipart/danger-js/distribution/runner/runners/inline.js:37:53)
    at /Volumes/Projects/multipart/danger-js/distribution/runner/runners/inline.js:31:71
    at new Promise (<anonymous>)
    at __awaiter (/Volumes/Projects/multipart/danger-js/distribution/runner/runners/inline.js:27:12)
    at Object.runDangerfileEnvironment (/Volumes/Projects/multipart/danger-js/distribution/runner/runners/inline.js:123:132)
```

# Addresses original error

```
Cannot use import statement outside a module
dangerfile.ts:46
import { parse } from 'path';
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at internalCompileFunction (node:internal/vm:76:18)
    at wrapSafe (node:internal/modules/cjs/loader:1283:20)
    at Module._compile (node:internal/modules/cjs/loader:1328:27)
    at requireFromString (/usr/src/danger/node_modules/require-from-string/index.js:28:4)
    at /usr/src/danger/dist/runner/runners/inline.js:183:68
    at step (/usr/src/danger/dist/runner/runners/inline.js:56:23)
    at Object.next (/usr/src/danger/dist/runner/runners/inline.js:37:53)
    at /usr/src/danger/dist/runner/runners/inline.js:31:71
    at new Promise (<anonymous>)
    at __awaiter (/usr/src/danger/dist/runner/runners/inline.js:27:12)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 ## Main
 
 <!-- Your comment below this -->
-
+- [#1180] Set module properly when tsconfig does not contain compilerOptions.module [@matthewh]
 <!-- Your comment above this -->
 
 ## 12.3.2
--- a/source/runner/runners/utils/transpiler.ts
+++ b/source/runner/runners/utils/transpiler.ts
@@ -149,10 +149,14 @@ const sanitizeTSConfig = (config: any, esm: boolean = false) => {
   //
   // @see https://github.com/apollographql/react-apollo/pull/1402#issuecomment-351810274
   //
-  if (!esm && safeConfig.compilerOptions.module) {
-    safeConfig.compilerOptions.module = "commonjs"
-  } else {
-    safeConfig.compilerOptions.module = "es6"
+  if (safeConfig.compilerOptions.module) {
+    if (!esm) {
+      // .ts files should fall back to commonjs
+      safeConfig.compilerOptions.module = "commonjs"
+    } else {
+      // .mts files must use `import`/`export` syntax
+      safeConfig.compilerOptions.module = "es6"
+    }
   }
 
   return safeConfig


terrymun:
Thanks for looking into that! I was about to create a MCVE on a dummy repo but you beat me to finding a fix :)

fbartho (approved):
Nice catch. This fix seems consistent with how the issue was introduced.

Not sure how easy this request is, but it feels like it’d be ideal to have a test case for this?

matthewh:
@fbartho I'd love to hear your thoughts on how to test this. I noticed the __Dangerfile fixtures but I don't see anything that sets up a project to trigger this issue given the interplay with existing tsconfig settings. I'm confident I could work something up if desired.

fbartho:
@matthewh I was equally curious re the tests. This feels like it falls in the bucket of a pain to setup a test environment.

(I don't have any obvious/easy advice, hence my PR approval)

Once a year for the last few years I tripped over ESM issues for various projects, and tripped across edge cases with TypeScript, and I wasn't looking forward to needing to solve them here in DangerJS! At least this *works* with the setups we have tested, so that's incremental, good, progress in my opinion.

matthewh:
@fbartho Distilled to the core, I think the variants distill to the following:
* `package.json` - the value of `module: boolean` influences typescript.
* `<dangerfile>` - there are 4 variants now mts, mjs, ts, and js.
* `tsconfig.json` -  danger only directly depends on compileOptions.modules though there are a huge number of settings that could influence the runner. Pick the most common and add new fixtures as bugs are reported. This brings up an question †

† why does danger need to use the current tsconfig at all? Because dangerfile imports live inside the current project. This seems messy when we consider that danger runs on top of a project and _should never be a dependency_. Ideally, I'd run `npx danger` and enforce a stricter set of guidelines for how a dangerfile should be structured. Of course this would likely mean danger would need to handle loading any dependencies at run time. Running danger would generate `.danger_modules` in the current directory adjacent to the dangerfile and keep all dependencies isolated. Overall this would reduce project interdependency on js projects though it sounds like a lift.

fbartho:
No objections to your comments about the variants!
<hr/>
I don't think the way forwards is as obvious as you suggest @matthewh,

>Ideally, I'd run `npx danger` and enforce a stricter set of guidelines for how a dangerfile should be structured. 

For example I strongly disagree with using `npx` without locking danger in as a versioned package.json dependency (forgive me if I jumped to conclusions for what you were saying. I'm objecting to `danger` being an ambient global dependency of indeterminate or unspecified version). 

>Of course this would likely mean danger would need to handle loading any dependencies at run time. Running danger would generate .danger_modules in the current directory adjacent to the dangerfile and keep all dependencies isolated. Overall this would reduce project interdependency on js projects though it sounds like a lift.

I think you're right this would be a hell of a lift. I also don't think it's obvious that `.danger_modules` is easier for people to manage and maintain. I don't think that would necessarily simplify the full problem space. I could be convinced that that might simplify a subset of the problems, but I think it would make other maintenance/debugging/support/usage tasks that much more complicated.

matthewh:
`.danger_modules` is a half baked idea 😁. In my mind you'd define any dependencies in your dangerfile and those would be loaded  by danger. For example:

```
DEPENDENCIES = [
  "danger-plugin-...@<version>",
  "danger-plugin-...@<version>",
  "danger-plugin-...@<version>",
]
```

then when danger runs it would "npm install" these as needed.

It's an interesting problem space.

orta:
Cool, discussion makes sense but lets get this shipped and folks can decide if there's more to do

PR #1476, opened 2025-02-09
Support files being ignored by babel configuration

### LINKED ISSUE #1469: [BUG] babelify in transpiler assumes no ignored babel files
**Describe the bug**

Could be same as https://github.com/danger/danger-js/issues/664...

The code [here](https://github.com/danger/danger-js/blob/491bb0b401148366cbe0e4d2bc9705fea521a647/source/runner/runners/utils/transpiler.ts#L175):
```js
const options = babel.loadOptions ? babel.loadOptions({ filename }) : { plugins: [] }
```
...assumes, as per [babel documentation](https://babeljs.io/docs/babel-core#loadoptions):
>...resulting in an options object where:
>
>`opts.plugins` is a full list of Plugin instances.

However, if a file is [ignored](https://babeljs.io/docs/options#ignore), the options are returned as `null` and [the attempt to spread `options.plugins`](https://github.com/danger/danger-js/blob/491bb0b401148366cbe0e4d2bc9705fea521a647/source/runner/runners/utils/transpiler.ts#L183C5-L183C51):
```js
plugins: [...extraPlugins, ...options.plugins]
```
... causes a failure:
```shell
Error:  TypeError: [BABEL]: Cannot read properties of null (reading 'plugins')
```
Perhaps we can alter the ternary to be:
```js
const options = babel.loadOptions?.({ filename: filename }) ?? { plugins: [] };
```

See https://github.com/danger/danger-js/pull/804 that last touched this.

**To Reproduce**
1. Set an ignored file in `.babelignore` or `ignore` in config
2. Run danger

**Expected behavior**

File is ignored for babel transpilation, danger processes it without issue.

**Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          | 11.3.0
| node             | 22.3.0
| npm              | 10.9.2
| Operating System | macOS 14.7.1

TomStrepsil:
Resolves #1469

Ensure that when `babel.loadOptions` returns `null` (which will be the case if a file is ignored by a babel configuration), this falls through to the guard to return the minimal options (`{ plugins: [] }`), thus preventing the error described in the issue.

For context, this was needed in a project that used [a local babel preset](https://github.com/ASOS/web-toggle-point/blob/main/peripheral/babel-preset-asos/), which had to be ignored to avoid a circular reference in babel.  When using a typescript dangerfile, the shared babel config baulked, due to attempt to spread `null`. 

This could be mitigated via `DANGER_DISABLE_TRANSPILATION`, but this then requires any local danger plugins etc to be in commonJs, which isn't ideal.

N.B. The `babelify` code was previously untested, and I have added bare-minimum test of the new functionality I have added, to simplify the review.  However, I am happy to attempt to "scout rule" the remainder of the function in tests, if that will be welcome.

N.B. The mocked dangerfile (copied from prior-art in the same test file) had to be changed to a _linted version_ - since despite tests passing locally otherwise, [the appveyor build](https://ci.appveyor.com/project/orta/danger-js/builds/51480736) had a failure thus:
```diff
    - Expected  - 1
    + Received  + 2
    - import {a} from 'lodash'; a()
    + import { a } from 'lodash';
    + a();
``` 

p.s. running:
```shell
yarn build; node --inspect distribution/commands/danger-pr.js https://github.com/danger/danger-js/pull/1476
```
...as per [the guide](https://github.com/danger/danger-js/blob/b07053c077cdcd200e22705f450970eb878a804f/README.md?plain=1#L86), has produced:
```shell
Processed 122 files (1.8s) (1 warning)

✔ No circular dependency found!

✨  Done in 5.72s.
Debugger listening on ws://127.0.0.1:9229/20b17684-51a1-4385-85ce-b6c9202e43ab
For help, see: https://nodejs.org/en/docs/inspector
Starting Danger PR on danger/danger-js#1476
You don't have a DANGER_GITHUB_API_TOKEN set up, this is optional, but TBH, you want to do this
Check out: http://danger.systems/js/guides/the_dangerfile.html#working-on-your-dangerfile
Jest tests passed :+1:

Unable to evaluate the Dangerfile
 Error: process.env.GITHUB_STEP_SUMMARY was not set, which is needed for setSummaryMarkdown
    at Object.setSummaryMarkdown (/danger-js/distribution/platforms/GitHub.js:133:23)
    at dangerfile.ts:89:31
    at step (dangerfile.ts:36:23)
    at Object.next (dangerfile.ts:17:53)
    at fulfilled (dangerfile.ts:8:58)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)


Danger: ⅹ Failing the build, there is 1 fail.
## Failures
Danger failed to run `dangerfile.ts`.
## Markdowns
## Error Error

process.env.GITHUB_STEP_SUMMARY was not set, which is needed for setSummaryMarkdown
Error: process.env.GITHUB_STEP_SUMMARY was not set, which is needed for setSummaryMarkdown
    at Object.setSummaryMarkdown (/git/danger-js/distribution/platforms/GitHub.js:133:23)
    at dangerfile.ts:89:31
    at step (dangerfile.ts:36:23)
    at Object.next (dangerfile.ts:17:53)
    at fulfilled (dangerfile.ts:8:58)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

### Dangerfile

--------------------------^
```

...which appears unrelated, and unclear how this can be resolved as a contributor?  

If I run danger locally on this branch, no issues are present:
```shell
node distribution/commands/danger-local.js --base main

Danger: ✓ passed review, received no feedback.
```

### DIFF
--- a/source/platforms/_tests/fixtures/bbc-dsl-input.json
+++ b/source/platforms/_tests/fixtures/bbc-dsl-input.json
@@ -2542,7 +2542,8 @@
   "settings": {
     "github": {
       "accessToken": "12345",
-      "additionalHeaders": {}
+      "additionalHeaders": {},
+      "baseURL": "https://api.github.com"
     },
     "cliArgs": {}
   }
--- a/source/runner/runners/utils/_tests/_transpiler.test.ts
+++ b/source/runner/runners/utils/_tests/_transpiler.test.ts
@@ -92,7 +92,8 @@ describe("dirContains", () => {
 
 describe("babelify", () => {
   it("does not throw when a filepath is ignored due to babel options, and should return the content unchanged", () => {
-    const dangerfile = "import {a} from 'lodash'; a()"
+    const dangerfile = `import { a } from 'lodash';
+a();`
 
     const existsSyncMock = fs.existsSync as jest.Mock
     const actualFs = jest.requireActual("fs") as typeof fs
--- a/source/runner/runners/utils/_tests/_transpiler.test.ts
+++ b/source/runner/runners/utils/_tests/_transpiler.test.ts
@@ -2,13 +2,15 @@ jest.mock("fs", () => ({
   readFileSync: jest.fn(),
   realpathSync: {},
   existsSync: jest.fn(),
+  statSync: jest.fn(),
+  stat: jest.fn(),
 }))
 jest.mock("path", () => {
   const path = jest.requireActual("path")
   return { ...path, resolve: jest.fn(path.resolve) }
 })
 
-import { typescriptify, lookupTSConfig, dirContains } from "../transpiler"
+import { typescriptify, lookupTSConfig, dirContains, babelify } from "../transpiler"
 import * as fs from "fs"
 import * as path from "path"
 
@@ -86,4 +88,25 @@ describe("dirContains", () => {
       expect(dirContains(`c:\\a`, `d:\\a`)).toBe(false)
     })
   }
+})
+
+describe("babelify", () => {
+  it("does not throw when a filepath is ignored due to babel options, and should return the content unchanged", () => {
+    const dangerfile = "import {a} from 'lodash'; a()"
+
+    const existsSyncMock = fs.existsSync as jest.Mock
+    const actualFs = jest.requireActual("fs") as typeof fs
+    existsSyncMock.mockImplementation((path) => path === "/a/b/babel.config.js" || actualFs.existsSync(path))
+    jest.mock(
+      "/a/b/babel.config.js",
+      () => {
+        return { ignore: ["/a/b"] }
+      },
+      { virtual: true }
+    )
+    const act = () => babelify(dangerfile, "a/b/", [])
+
+    expect(act).not.toThrow()
+    expect(act()).toEqual(dangerfile)
+  })
 })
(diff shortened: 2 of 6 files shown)


source/runner/runners/utils/_tests/_transpiler.test.ts
```diff
@@ -2,13 +2,15 @@ jest.mock("fs", () => ({
   readFileSync: jest.fn(),
   realpathSync: {},
   existsSync: jest.fn(),
+  statSync: jest.fn(),
```
TomStrepsil, line 5: N.B. These are needed to prevent an early-return within babel.

source/platforms/_tests/fixtures/bbc-dsl-input.json
```diff
@@ -2542,7 +2542,8 @@
   "settings": {
     "github": {
       "accessToken": "12345",
-      "additionalHeaders": {}
+      "additionalHeaders": {},
+      "baseURL": "https://api.github.com"
```
TomStrepsil, line 2546: N.B. This was produced when running `yarn lint` - unclear why previous PRs did not produce the same?

orta (approved):
Yeah, this makes sense to me 👍🏻 

Thanks!

source/platforms/_tests/fixtures/bbc-dsl-input.json
orta: interesting, weird!

PR #1477, opened 2025-02-20
chore: Upgrade octokit/rest.js for CVE patch

chris-griffin:
Just some cleanup for a transitive CVE: https://github.com/advisories/GHSA-h5c3-5r3r-rr8q

- v19 dropped support for node 10/12 (https://github.com/octokit/rest.js/releases/tag/v19.0.0)
- v20 dropped support for node 14/16, removed preview support for the REST API, and removed the agent option (https://github.com/octokit/rest.js/releases/tag/v20.0.0)
- v21 updated the package to ESM (https://github.com/octokit/rest.js/releases/tag/v21.0.0)

None of these breaking changes should impact v12 of danger-js as it requires node >= 18.

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,7 +145,7 @@
   },
   "dependencies": {
     "@gitbeaker/rest": "^38.0.0",
-    "@octokit/rest": "^18.12.0",
+    "@octokit/rest": "^21.1.1",
     "async-retry": "1.2.3",
     "chalk": "^2.3.0",
     "commander": "^2.18.0",


mrginglymus:
danger-js also pulls in `memfs-or-file-map-to-github-branch`, which still relies on the old version of `@octokit/rest`:

https://github.com/orta/memfs-or-file-map-to-github-branch/blob/c293897e6b7cf80b23f16956ee2dc236953323e7/package.json#L19

This package will need updating in order to prevent danger-js pulling in the CVE.

fbartho:
@mrginglymus @chris-griffin: I've provided a PR for `memfs-or-file-map-to-github-branch` here https://github.com/orta/memfs-or-file-map-to-github-branch/pull/6

Now I think we're stuck until @orta gets a chance to look at the upstream library.

fbartho:
@chris-griffin I went to test this PR locally (and update `devDependency` `release-it` to a newer version, since that also depended on ancient `@octokit/rest`, and I ran into two unexpected issues:

1. `yarn.lock` wasn't updated in this PR, did you use npm or something else?
2. The above suggested you hadn't hooked in to run the pre-push scripts.

When I ran `yarn build` I got the following errors:

```shell
$ yarn build
…
source/platforms/GitHub.ts:6:39 - error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("@octokit/rest")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mts', or add the field `"type": "module"` to '~/Code/danger-js/package.json'.

6 import { Octokit as NodeGitHub } from "@octokit/rest"
                                        ~~~~~~~~~~~~~~~
# Many of the same type of error repeated for different files.
…
```

<hr/>

Looks like we need to do some work to import from ESM dependencies. ***sigh***
Any support would be appreciated!

fbartho:
I made an alternative PR here: #1481 (including @chris-griffin's initial changes)

PR #1483, opened 2025-04-10
feat: add pre-commit hook configuration

adam-moss:
This adds support for using danger directly with `.pre-commit-hooks.yaml` configuration file widely used with the python (and other) ecosystems.

More info https://pre-commit.com

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,13 @@
+- id: danger
+  name: danger
+  description: formalize your team etiquette
+  language: node
+  stages:
+    - pre-commit
+  entry: danger local
+  args:
+    - --base=main
+    - --dangerfile=dangerfile.local.js
+    - --staging
+  pass_filenames: false
+  always_run: true


orta:
Is this something that people would expect to find via searching through `node_modules`?

adam-moss:
> Is this something that people would expect to find via searching through `node_modules`?

No, in their repo the would define a `.pre-commit-config.yaml` file which would look something like:

<details><summary>Example .pre-commit-config.yaml</summary>
<p>

```yaml
%YAML 1.2
# yaml-language-server: $schema=https://json.schemastore.org/pre-commit-config.json
---
default_install_hook_types:
  - pre-commit
  - prepare-commit-msg
  - commit-msg
  - post-commit
  - pre-push

default_stages:
  - pre-commit

fail_fast: true

repos:
  - repo: meta
    hooks:
      - id: check-hooks-apply
        stages:
          - manual
      - id: check-useless-excludes
        stages:
          - manual
      - id: identity
        stages:
          - manual

  - repo: https://github.com/danger/danger-js
    rev: v12.3.4
    hooks:
      - id: danger
        additional_dependencies:
          - "@gitlab/codeowners"
```

</p>
</details> 

You can think of `pre-commit` as being similar to `husky`, widely used outside the node ecosystem.

orta:
Eh, doesn't seem like too much faff for us, sure 👍🏻

glensc:
btw, this change was just removed in 974713052d101ad1874e7c4c1f3ec71f09e65b1c. Why?

PR #1488, opened 2025-05-15
refactor: replace get-stdin with native stream/consumers module

baevm:
Hi! Since danger-js targets node >= 18, get-stdin can be replaced with native stream/consumers module
reference: https://github.com/sindresorhus/get-stdin?tab=readme-ov-file#tip


fbartho (approved):
Love it! Thanks @baevm

baevm:
@orta @fbartho just a friendly reminder about this PR - looks like its good to merge?

orta:
Lets at least get this in, and then when something else arrives it will be a part of the process :+1:

PR #1497, opened 2025-11-14
chore(deps): bump husky to v9

yoshi-taka:
https://dev.to/typicode/what-s-new-in-husky-5-32g5

This PR upgrades Husky from v1 to v9, moving from the legacy JS-based hook runner to the modern core.hooksPath + shell-based architecture.

Why this change
	•	Husky v9 is significantly lighter and faster.
	•	The old v1 flow relied on JS runners and extra scaffolding, which increased overhead and required more maintenance.
	•	v9 reduces dependencies to zero and improves hook startup time.

What changed
	•	Updated Husky to v9 and regenerated the .husky/ directory.

Benefits
	•	Smaller install footprint.
	•	Faster pre-commit / pre-push execution.
	•	Simpler, more intuitive hooks.
	•	Modern and actively maintained architecture.


orta (approved):
Seems good to me

Issue #1489, opened 2025-10-06
[BUG] npm package contains extra files


rubiesonthesky:
**Describe the bug**

When looking at the code in npmjs.org (https://www.npmjs.com/package/danger?activeTab=code), you can see that there are a lot extra files that do not need to be included to the package. The size of `danger` is already big, so it would be nice to reduce the install size even a little bit.

Paths that could be added to `.npmignore` 
- `.github`
- `.eslintignore`
- `.eslintrc.json`
- `.gitattributes`
- `.pre-commit-config.yaml`
- `.prettierignore`
- `.release-it.json`
- `CHANGELOG.md`
- `CONTRIBUTING.md`
- `Dockerfile`
- `VISION.md`
- `tsconfig.spec.json`

Total would be around 90 kB saved. It's not that much of the total size of 1.43 MB, but at least something.

glensc:
But for now, I just took the list and inserted it (easier)
- https://github.com/danger/danger-js/pull/1499

glensc:
I don't think anybody is doing the whitelist based gitignores:
- https://github.com/danger/danger-js/issues/1489#issuecomment-3648149266

so, this can be just closed

glensc:
@rubiesonthesky, @orta close?

glensc:
First this looked like it's not released, but there seems to be some mess with git tags:
- https://github.com/danger/danger-js/issues/1507

Also, the [changelog](https://github.com/danger/danger-js/blob/d38955b019daeb5ecfabcae8d0d0524632537b4c/CHANGELOG.md) is messed up (I'm working on a pr).

PR #1499, opened 2025-12-12
Build: Update .npmignore to exclude additional files

glensc:
Added various configuration and documentation files to `.npmignore`:
- https://github.com/danger/danger-js/issues/1489


fbartho (approved):
No strong objections, just a very light opinion!

PR #1501, opened 2026-02-27
refactor: replace p-limit with inline concurrency limiter

roli-lpci:
## Summary

Replace the `p-limit` dependency (and its `@types/p-limit` dev dependency) with a lightweight inline concurrency limiter (~30 lines) in `GitHubAPI.ts`.

### Motivation

- `p-limit` v3+ is ESM-only, which creates friction in CJS codebases — the project currently pins `^2.1.0` to avoid this
- The usage is a single call site: a fixed concurrency of 25 for GitHub API requests
- Inlining removes 2 dependencies (`p-limit` + `@types/p-limit`) with zero behavior change

### Changes

- **`source/platforms/github/GitHubAPI.ts`** — Remove `p-limit` import, add inline `pLimit()` function with identical queue-based semantics
- **`package.json`** — Remove `p-limit` from `dependencies` and `@types/p-limit` from `devDependencies`
- **`yarn.lock`** — Updated automatically

### Testing

- `tsc --noEmit` passes cleanly
- Production build (`tsc -p tsconfig.production.json`) compiles successfully
- No circular dependencies (`madge` check passes)
- All 47 GitHub-related unit tests pass


orta (approved):
Yeah, nice idea

PR #1258, opened 2022-03-18
Fix: Enable exception mode to skip scripts on first error

glensc:
Replaces https://github.com/danger/danger-js/pull/1254

Noticed that "danger:prepush" ran commands with ";" rather "&&" meaning the next command is still executed if the previous one failed. This resulted in very weird errors due to the previous step failure.

Now, aborts of first error:
```
➜ yarn danger:prepush
yarn run v1.22.17
$ set -e; yarn build:fast; yarn test:fixtures; node distribution/commands/danger.js local --base main --dangerfile dangerfile.lite.ts
$ tsc -p tsconfig.production.json
source/commands/ci/runner.ts:70:30 - error TS2339: Property 'publishCheck' does not exist on type 'SharedCLI'.
...
➜ 
```

PR #1508, opened 2026-04-12
Fix: Use && to fail on scripts on error

glensc:
Replaces with  https://github.com/danger/danger-js/pull/1258 using `&&`, to be windows compatible. Other scripts already do [this way](https://github.com/danger/danger-js/pull/1258#issuecomment-4231344497):
- https://github.com/danger/danger-js/commit/20d5cbb29882278bde702d7fdbc9626ec9784f38
- https://github.com/danger/danger-js/commit/d8c94281fa566ea43481f033fd7ff3f416c01498


orta:
👍🏻

PR #1514, opened 2026-06-20
feat: swapping our node-fetch for unidici / native

nicholasgriffintn:
- Replaces direct `node-fetch` imports with `undici` and removes dependencies
- Moves proxy handling from `agent` to `dispatcher`
- Removes GitLab fetch shim and upgrades nock
- Updates tests

Note:

Unidici v8 drops support for node 20 and v7 dropped support for node 18.

It would be best to be on the latest version but that would require Danger JS to presumably also drop support for 18 and 20.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,6 @@
     "@types/lodash.memoize": "^4.1.3",
     "@types/micromatch": "^3.1.0",
     "@types/node": "18.19.18",
-    "@types/node-fetch": "^2.5.12",
     "@types/parse-github-url": "^1.0.3",
     "@types/prettier": "^1.16.1",
     "@types/readline-sync": "^1.4.3",
@@ -132,7 +131,7 @@
     "jest-json-reporter": "^1.2.2",
     "lint-staged": "^12.3.7",
     "madge": "^6.0.1",
-    "nock": "^13.2.0",
+    "nock": "^14.0.15",
     "pkg": "^5.8.1",
     "prettier": "^2.5.1",
     "release-it": "^18.1.2",
@@ -152,8 +151,6 @@
     "core-js": "^3.8.2",
     "debug": "^4.1.1",
     "fast-json-patch": "^3.0.0-1",
-    "http-proxy-agent": "^7.0.2",
-    "https-proxy-agent": "^7.0.2",
     "hyperlinker": "^1.0.0",
     "ini": "^5.0.0",
     "json5": "^2.2.3",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -176,7 +172,8 @@
     "readline-sync": "^1.4.9",
     "regenerator-runtime": "^0.13.9",
     "require-from-string": "^2.0.2",
-    "supports-hyperlinks": "^4.3.0"
+    "supports-hyperlinks": "^4.3.0",
+    "undici": "6.21.1"
```
fbartho, line 176: https://www.npmjs.com/package/undici This seems to suggest that the version you selected is several major versions behind. (unless I’m confused?).

I’m not super familiar with undici, but if we’re swapping out dependencies, I would have expected to be on the latest version.

fbartho (changes requested):
Since this is a pretty high sensitivity spot for code change I look forward to further reviews from participants.

Thanks for starting the process!

package.json
nicholasgriffintn: Sorry, I should have wrote the reason for that, I forgot to. I'll update the description.

Unidici v8 drops support for node 20 and v7 dropped support for node 18.

It would be best to be on the latest version but that would require Danger JS to presumably also drop support for 18 and 20.

fbartho: Huh! Not my decision to make, but this might be enough for me to suggest moving to node v24 as our minimum?

nicholasgriffintn: Yeah happy to update it, just working from the current support that's documented in the package.

orta: I think this is worth keeping lower, backwards compat is pretty important to this project 👍🏻

orta:
TBH, I'm pretty down for a 'cleanup for node 24 style thing for Danger JS. Personally, I've been moving all my projects to use [risk instead](https://github.com/danger/risk).

orta:
LEts do this

package.json
glensc: but it shouldn't be pinned to certain version, but use semver range: `"undici": "^6.21.1"`

nicholasgriffintn: Personally, I always use pinned to avoid supply chain attacks and then allow automations to upgrade via PRs like dependabot.

glensc: You shouldn't pin in **libraries**, as then downstream apps can't upgrade, but rather require library to update pin again.

However, this concrete dependency was already upgraded due same reason, that the pinned version needs security update: #1518, but it has no effect downstream, until another danger-js npm release is made.

nicholasgriffintn: not strictly true, sure thats a usability concern, but doing so is causing security concerns its just which one you see as more of a problem.

but sure, seems to have been changed, my default is always to pin nowadays, outside of peer deps.

glensc: pinning in libraries moves burden of updating dependency to library maintainer. if library uses version range, you can update dependency in your app.

surely, certain package mangers have "overrides" to overcome pinned version in library, but that goes more to danger zone, as there's no validation what version you pick. could be totally incompatible version.

PR #1516, opened 2026-06-23
fix: avoid spurious ERR_STREAM_PREMATURE_CLOSE on Node 24.17.0+ by declining gzip for API requests

### LINKED ISSUE #1515: danger ci fails with FetchError: Premature close (ERR_STREAM_PREMATURE_CLOSE) on Node 24.17.0+
## Summary

On **Node 24.17.0+**, every `danger ci` run against the GitHub API fails during DSL assembly while fetching the PR file list / diff / commits — *before the Dangerfile is evaluated*:

```
Failed to fetch GitHub pull request files: FetchError: Invalid response body while trying to fetch https://<host>/api/v3/repos/<owner>/<repo>/pulls/<n>/files?page=1&per_page=100: Premature close
    at Gunzip.<anonymous> (node_modules/node-fetch/lib/index.js:400:12)
  type: 'system',
  errno: 'ERR_STREAM_PREMATURE_CLOSE',
  code: 'ERR_STREAM_PREMATURE_CLOSE'
```

It is deterministic: **green on Node 24.16.0, red on Node 24.17.0** for the same repo and the same Danger version.

## Root cause

This is **not** a real decode/stream failure — it's a false positive from a Node change interacting with `node-fetch@2`:

- Node 24.17.0 (a security release) changed `http.Agent` so that **idle keep-alive sockets** in the free socket pool retain a `'data'` listener. See nodejs/node#63989 (*"node-fetch throws ERR_STREAM_PREMATURE_CLOSE on keep-alive socket closures after latest security releases"*) and the fix nodejs/node#64004 (*"http: avoid stream listeners on idle agent sockets"*).
- `node-fetch@2`'s `fixResponseChunkedTransferBadEnding` (`lib/index.js:1739`) only runs for responses sent as `Transfer-Encoding: chunked` with **no** `Content-Length` — which is exactly how the GitHub API returns gzip-encoded JSON. On socket `close` it decides the connection ended uncleanly via `socket.listenerCount('data') > 0` (`lib/index.js:1745`) and throws a synthetic `ERR_STREAM_PREMATURE_CLOSE` (`:1748`). With the new idle-socket listener, that check now fires on perfectly healthy responses.

The `Gunzip` in the stack trace is incidental — it's just the response body stream the synthetic error gets `.destroy()`'d onto.

## Why this needs addressing in Danger

- The crashing requests (`getPullRequestFiles`, `getPullRequestDiff`, `getPullRequestCommits`) are issued by Danger core during DSL assembly, **before the Dangerfile runs**, so consumers have no seam to work around it (no request-option hook; the only fetch DI point is reserved for Peril).
- `node-fetch@2` is effectively frozen at 2.7.0 and won't be patched; `node-fetch@3` is ESM-only and can't be `require()`'d by Danger.
- The real fix (nodejs/node#64004) is merged but **not yet in any released Node line** (latest at the time of writing: v24.17.0 / v22.23.0), so "upgrade Node" isn't actionable for affected users today.

## Suggested fix

Have Danger's fetch wrapper decline gzip negotiation (`compress: false`) for its API requests. The server then returns identity-encoded bodies **with a `Content-Length`**, so the `chunked && !content-length` branch never arms. Minimal, backwards-compatible, no dependency change, and easily reverted once a patched Node line ships or the undici migration (#1514) lands.

I'll open a PR with this change plus a regression test.

## Environment

| | |
|---|---|
| Danger | 13.x (also affects 11.x / 12.x — all depend on `node-fetch@^2`) |
| Node | 24.17.0+ (and the corresponding 22.x security release line) |
| Platform | GitHub / GitHub Enterprise |

sakkadak:
Fixes #1515

## Problem

On **Node 24.17.0+**, every `danger ci` run fails during DSL assembly while fetching the PR files / diff / commits, before the Dangerfile is evaluated:

```
FetchError: Invalid response body ... /pulls/<n>/files: Premature close
  at Gunzip.<anonymous> (node_modules/node-fetch/lib/index.js:400)
  errno: 'ERR_STREAM_PREMATURE_CLOSE', code: 'ERR_STREAM_PREMATURE_CLOSE'
```

Deterministic: green on Node 24.16.0, red on Node 24.17.0.

## Root cause

Node 24.17.0's `http.Agent` change leaves a `'data'` listener on idle keep-alive sockets (nodejs/node#63989; fixed in nodejs/node#64004 *"http: avoid stream listeners on idle agent sockets"*). `node-fetch@2`'s `fixResponseChunkedTransferBadEnding` (`lib/index.js:1739`/`1745`) inspects `socket.listenerCount('data')` to detect an unclean end, and now false-positives for `Transfer-Encoding: chunked` responses with no `Content-Length` — i.e. gzip-encoded GitHub API responses — throwing a synthetic `ERR_STREAM_PREMATURE_CLOSE`. It is not a real decode error; the `Gunzip` in the trace is just the body stream the synthetic error is destroyed onto.

`node-fetch@2` is frozen and `node-fetch@3` is ESM-only, and the Node fix isn't in a released line yet, so the practical place to address this is Danger's own fetch wrapper.

## Fix

Set `compress: false` (guarded with `=== undefined`, so callers can still opt in) in the shared `api()` wrapper, alongside the existing proxy-agent handling. The server then returns identity-encoded bodies **with a `Content-Length`**, so node-fetch's bad-ending check never arms. This mirrors how the wrapper already centralizes cross-cutting request config (the proxy `agent`).

## Backwards compatibility

- No signature / dependency / Node-floor change; valid on all supported Node versions.
- The `=== undefined` guard preserves any explicit caller `compress` setting.
- Only observable change: requests no longer advertise gzip; responses come back identity + `Content-Length`. Danger's payloads are small JSON, so the bandwidth difference is negligible.

## Tests

Extends `source/api/_tests/fetch.test.ts` (same `TestServer` pattern as the proxy-agent tests):
- defaults `compress` to `false`
- does not override an explicitly provided `compress`

`yarn jest source/api/_tests/fetch.test.ts` → 9/9 pass. `tsc -p tsconfig.production.json`, `eslint`, and `prettier --check` are all clean.

## Notes / open to feedback

- This applies to every platform that goes through `api()` (GitHub / GitLab / BitBucket), not just GitHub. The bug is platform-agnostic, so the broad fix protects them all — but I'm happy to narrow the scope to the GitHub path if you'd prefer.
- This is an interim shim; it can be reverted once a patched Node line ships or #1514 (undici migration) lands.

Opened as a **draft** for your review of the approach before marking ready.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 <!-- Your comment below this -->
 
+- Fix spurious `Premature close` (`ERR_STREAM_PREMATURE_CLOSE`) failures when running on Node 24.17.0+ by no longer requesting gzip-compressed GitHub API responses - [@sakkadak]
 
 <!-- Your comment above this -->
 
@@ -2137,6 +2138,7 @@ Not usable for others, only stubs of classes etc. - [@orta]
 [@rouby]: https://github.com/rouby
 [@rzgry]: https://github.com/rzgry
 [@sajjadzamani]: https://github.com/sajjadzamani
+[@sakkadak]: https://github.com/sakkadak
 [@sandratatarevicova]: https://github.com/sandratatarevicova
 [@sebinsua]: https://github.com/sebinsua
 [@sgtcoolguy]: https://github.com/sgtcoolguy
--- a/source/api/_tests/fetch.test.ts
+++ b/source/api/_tests/fetch.test.ts
@@ -172,4 +172,20 @@ describe("fetch", () => {
     let agent = options.agent as HttpProxyAgent<string>
     expect(agent["proxy"].href).toBe(proxyUrl)
   })
+
+  it("disables compression by default to avoid the Node 24.17+ node-fetch premature-close", async () => {
+    await server.start({})
+
+    let options: node_fetch.RequestInit = {}
+    await api(url, options, true)
+    expect(options.compress).toBe(false)
+  })
+
+  it("does not override an explicitly provided compress option", async () => {
+    await server.start({})
+
+    let options: node_fetch.RequestInit = { compress: true }
+    await api(url, options, true)
+    expect(options.compress).toBe(true)
+  })
 })
--- a/source/api/fetch.ts
+++ b/source/api/fetch.ts
@@ -116,6 +116,19 @@ export function api(
     init.agent = secure ? new HttpsProxyAgent(proxy) : new HttpProxyAgent(proxy)
   }
 
+  // Node >= 24.17.0 attaches a 'data' listener to idle keep-alive sockets in the http.Agent
+  // free pool (the CVE-2026-48931 "response queue poisoning" hardening). node-fetch@2's
+  // chunked-bad-ending detector (fixResponseChunkedTransferBadEnding) reads
+  // `socket.listenerCount('data')` and misreads that listener as an unclean connection close,
+  // throwing a bogus `ERR_STREAM_PREMATURE_CLOSE` ("Premature close") for responses sent as
+  // `Transfer-Encoding: chunked` with no `Content-Length` — i.e. gzip-encoded GitHub API
+  // responses. Declining gzip makes the server return an identity-encoded body with a
+  // `Content-Length`, so the detector never arms. Guarded with `=== undefined` so callers can
+  // still opt back into compression. See nodejs/node#63989 (report) and nodejs/node#64004 (fix).
+  if (init.compress === undefined) {
+    init.compress = false
+  }
+
   return retryableFetch(url, init).then(async (response: node_fetch.Response) => {
     // Handle failing errors
     if (!suppressErrorReporting && !response.ok) {


sakkadak:
**Upstream Node status (for context):** the Node-side fix ([nodejs/node#64004](https://github.com/nodejs/node/pull/64004)) is merged to `main` and backported to the `v24.x-staging` / `v22.x-staging` branches — but it is **not in any released tag**, not on the `v24.x` release branch, and not in the current v24.18.0 release proposal ([#64062](https://github.com/nodejs/node/pull/64062)) yet. So it hasn't been promoted off staging into a release.

Net: anyone on the current secure releases (24.17.0 / 22.23.0) — which are effectively mandatory due to CVE-2026-48931 — has **no Node-level fix available today**, which is the gap this change covers. It can be reverted once a tagged release actually carries the backport.

sakkadak:
Superseded by #1514 (released in v13.0.10), which removes `node-fetch` in favor of `undici` and fixes this at the root — so this `compress:false` shim no longer applies (the file it patched is now undici-based). Closing in favor of that. Thanks!

Issue #1517, opened 2026-07-06
[BUG]: Release 13.0.10 introduces security vulnerabilities due to package undici@6.21.1.


traysa:
**Describe the bug**
Release 13.0.10 introduces security vulnerabilities due to package undici@6.21.1.

_PS: I do not have a Bluesky account. It would be valuable for others to see if there is an security issue. Therefore posted as bug._

**To Reproduce**
Run actions/dependency-review-action in Github workflow

**Expected behavior**
No introduction of known security issues

**Screenshots**

<img width="1094" height="204" alt="Image" src="https://github.com/user-attachments/assets/2a03b956-a127-4961-b68a-04b4b5f86cbe" />

**Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          | 13.0.10
| node             |
| npm              |
| Operating System |

**Additional context**

fbartho:
I'm pretty sure danger doesn't connect to untrusted hosts via WebSocket, so assuming this is true, this is not a critical security issue for use. (In other words, this functionality is unused).

We should still fix it. Either by reverting to fetch, or upgrading to a version of undici that doesn't have the issue.

I believe we will need to raise our minimum supported version of node to get a later version of undici.

Thank you for reporting this issue!

glensc:
next release must be made to full fix this issue (or re-open until then):
- https://github.com/danger/danger-js/pull/1518#issuecomment-5049002019

PR #1518, opened 2026-07-06
Upgrade to undici 6.27.0 to resolve transitive CVEs - fixes #1517

rjatkins:
Unpinning undici so that further fixes to undici@6.x will be fixable in consumers, through their lockfiles

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 
 <!-- Your comment below this -->
 
+- Upgrade to undici 6.27.0 to resolve transitive CVEs - fixes [#1517](https://github.com/danger/danger-js/issues/1517) [@rjatkins]
+
 <!-- Your comment above this -->
 
 ## 13.0.9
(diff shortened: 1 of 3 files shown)


fbartho (approved):
🍻

orta:
thanks, agree!

orta:
hrm, we spent some time thinking about the version - unlocking this becomes a server major because it drops a lot of node versions

rjatkins:
Upgrading to `undici@7` or higher will drop node 18 and 19. `undici@6.21.1` and `6.27.0` both still support `>=18.17` and up. This is why I've used `undici@^6.27.0`, and not `undici@>=6.27.0`, since `^` ensures that `7.x.x` or `8.x.x` versions are still not used.

glensc:
This has no effect, until release is made:
- https://github.com/danger/danger-js/pull/1514/changes#r3632137488

so https://github.com/danger/danger-js/issues/1517 shouldn't be closed yet

PR #1522, opened 2026-07-08
[Dependency] Lock TS to 6.0.3

clates:
In support of https://github.com/danger/danger-js/issues/1521

PR #1524, opened 2026-07-22
Security: Unpin typescript in Dockerfile from 6.0.3

glensc:
The intention is to avoid installing 7.x, rather actually requiring the concrete 6.0.3 version
- https://github.com/danger/danger-js/pull/1522#issuecomment-5048922762


orta:
Yeah, this makes sense

Issue #1443, opened 2024-05-02
[BUG] GitHubReview['state'] type does not match current GitHub review states


clayton-duarte:
**Describe the bug**

The current `GitHubReview.state` type definition is `"APPROVED" | "REQUEST_CHANGES" | "COMMENT" | "PENDING" | undefined.`

But while trying to catch if a PR had any requested changes I found out that the state returned by GitHub is actually `CHANGES_REQUESTED`.

**To Reproduce**

1. Open a new PR;
2. Ask someone to request changes on that PR;
3.  Inside your `dangerfile` check the return of `reviews.find((review) => review.state === 'REQUEST_CHANGES')`;
4. This will not return the review you are looking for;
5. Now try `reviews.find((review) => review.state === 'CHANGES_REQUESTED')`;
6. This will return yourreview;

**Expected behavior**

`GitHubReview.state` type should correctly reflect the expected strings from GitHub;

**Screenshots**

**Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          | 11.2.8
| node             | v18.19.0
| npm              | 10.2.3
| Operating System | MacOS

**Additional context**
I've checked the most recent `danger` releases and the latest `12.1.0` release still presents this issue.

orta:
Sure, you're welcome to add it

clayton-duarte:
Thanks for the quick reply @orta !

I'll be very happy to contribute, I just have a question before start working on a PR.

I was checking the codebase and found out Danger uses github's REST API. With that in mind, I was checking their documentation and I have not-so-good news. They do not have a proper enum for that field:

https://docs.github.com/en/rest/pulls/reviews#get-a-review-for-a-pull-request

So, my suggestion is to simply change `GitHubReview.state` to `string` and add some examples as a comment.

What do you suggest?

clayton-duarte:
from their response schema:
```
    "state": {
      "type": "string",
      "examples": [
        "CHANGES_REQUESTED"
      ]
    },
```

orta:
Maybe change it to be the existing literal array with `| string` at the end so that people can get a sense but not be complacent?

daichunghy:
Still true on current `main`: `source/dsl/GitHubDSL.ts` line 456 has `state?: "APPROVED" | "REQUEST_CHANGES" | "COMMENT" | "PENDING"`.

[...]

PR #1525, opened 2026-07-30
Fix GitHubReview state values to match the GitHub API

Socialpranker:
`GitHubReview.state` was typed as `"APPROVED" | "REQUEST_CHANGES" | "COMMENT" | "PENDING"`, but the reviews API doesn't send `REQUEST_CHANGES` or `COMMENT`. It sends `CHANGES_REQUESTED` and `COMMENTED`, which is why the check in #1443 silently never matches:

```ts
reviews.find(review => review.state === 'REQUEST_CHANGES')
```

`DISMISSED` was missing too, so a dismissed review had no valid type either.

On the shape of the fix: @orta suggested keeping the literal array and adding `| string`, and checking GitHub's own OpenAPI description makes that the right call rather than just a lenient one. `pull-request-review.state` there is a plain `type: string` with `CHANGES_REQUESTED` as the example, with no enum, so GitHub deliberately doesn't commit to a closed set. A fully exact union would still be wrong even with the names corrected. The literals give you autocomplete, and `| string` stops the type from claiming a guarantee GitHub doesn't make.

One side effect worth flagging: `source/danger-incoming-process-schema.json` is generated by `typescript-json-schema`, and adding `| string` drops the `enum` from that property, leaving `"type": "string"`. That's the honest description now, since the old enum listed two values GitHub never sends and would reject the two it does. Nothing validates against that schema at runtime; it's there for other `danger-process` implementations.

The regenerated `danger.d.ts` and schema are included. `source/dsl/GitHubDSL.ts` also picked up two unrelated formatting hunks (`author_association` indentation and a trailing space) because lint-staged runs prettier over the whole staged file on commit.

Ran the full suite: 682 passing. Note that the tests don't run on Node 25 at all on main, since `buffer-equal-constant-time` reads `SlowBuffer.prototype` which Node 25 removed, so 32 suites fail to load before any test executes. On Node 22 everything is green.

Written with Claude Code (Claude Opus 5); I reviewed and tested everything before submitting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 <!-- Your comment below this -->
 
+- `github.reviews` now reports the review states GitHub actually sends, so checking for a review that requested changes works - fixes [#1443](https://github.com/danger/danger-js/issues/1443) [@Socialpranker]
 - Upgrade to undici 6.27.0 to resolve transitive CVEs - fixes [#1517](https://github.com/danger/danger-js/issues/1517) [@rjatkins]
 
 <!-- Your comment above this -->
@@ -2146,6 +2147,7 @@ Not usable for others, only stubs of classes etc. - [@orta]
 [@sharkysharks]: https://github.com/sharkysharks
 [@shyim]: https://github.com/shyim
 [@snowe2010]: https://github.com/snowe2010
+[@socialpranker]: https://github.com/Socialpranker
 [@sogame]: https://github.com/sogame
 [@soyn]: https://github.com/Soyn
 [@stefanbuck]: https://github.com/stefanbuck
--- a/source/danger-incoming-process-schema.json
+++ b/source/danger-incoming-process-schema.json
@@ -1888,13 +1888,7 @@
                     "type": "number"
                 },
                 "state": {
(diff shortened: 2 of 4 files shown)


Socialpranker:
The red `test` job here isn't from this PR. It fails in the "Make a small babel app" step while installing a fresh `create-react-app`:

```
error @testing-library/jest-dom@6.10.0: The engine "node" is incompatible with this module. Expected version ">=22". Got "20.20.2"
```

`.github/workflows/CI.yml` pins `node-version: "20"`, and that transitive dependency now requires Node 22, so the step fails before Danger runs. #1524 hits the identical error, so it's independent of what either PR changes. Happy to bump the workflow to Node 22 in a separate PR if that's useful.

fbartho (approved):

Socialpranker:
Now that #1526 is merged, I updated this branch onto current main and the `test` job is green (083bdc2) — the Node 20 engine error that was failing it is gone.

Approved, green and mergeable, so this should be ready whenever one of you has a moment. @orta @fbartho

Socialpranker:
I know I already flagged this as ready — just a last nudge since it's been sitting green and approved for a couple weeks now. Not trying to be a pest, just don't want it to slip through the cracks.

PR #1528, opened 2026-08-23
fix(github): paginate pull request reviews

### LINKED ISSUE #1383: [BUG] danger.github.reviews only providing the first 30 reviews
**Describe the bug**

When using `danger.github.reviews` it only includes the first 30 reviews on the PR. 

**To Reproduce**
1. Have a PR with more than 30 reviews.
2. Use   
  ```javascript
  console.log(danger.github.reviews.length);
  ```

**Expected behavior**

Expect to have the correct number of reviews.

**Screenshots**

**Your Environment**

 | software         | version
| ---------------- | -------
| danger.js          |  11.2.6
| node             | v16.20.0
| npm              | 8.19.4
| Operating System | Linux and MacOS

**Additional context**

LeonMAG:
## Summary

- paginate `danger.github.reviews` using the existing GitHub API helper
- preserve the review request's `Accept` header on every page
- add regression coverage for a second page of reviews and the single-page case
- add a changelog entry for #1383

Fixes #1383.

## Test plan

- [x] Added the pagination test first: it failed on the previous implementation with 30 reviews returned instead of 31.
- [x] `yarn test source/platforms/github/_tests/_github_api.test.ts --runInBand`
- [x] `yarn type-check`
- [x] `yarn lint`
- [x] `yarn build`
- [x] `yarn declarations`
- [x] `yarn tsc --esModuleInterop distribution/danger.d.ts`

The full Jest run reports all 62 suites passing (684 passing tests, 1 skipped), but exits with code 1 locally. The unchanged baseline branch has the same non-zero exit after all of its 62 suites pass, so this is not introduced by the patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 <!-- Your comment below this -->
 
+- GitHub: Fetch all pages of pull request reviews [#1383](https://github.com/danger/danger-js/issues/1383) [@LeonMAG]
 - Upgrade to undici 6.27.0 to resolve transitive CVEs - fixes [#1517](https://github.com/danger/danger-js/issues/1517) [@rjatkins]
 
 <!-- Your comment above this -->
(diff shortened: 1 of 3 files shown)


orta:
Yeah, this seems reasonable to me 👍🏻

PR #1532, opened 2026-08-23
Add support for deploying via github actions

orta:
Moves the npm publish off maintainer laptops and onto a tag-triggered Actions workflow, so the set of people who can ship a release is controlled by a repository ruleset rather than by who happens to hold write access (currently 241 people, via org membership).

### How releases work now

- `npm run release -- patch` bumps, commits, tags and pushes — it no longer publishes to npm, so there's no OTP to type.
- Pushing the version tag runs `npm_publish.yml`, which publishes to npm and then kicks off the existing `release.yml` for the macOS binaries and the homebrew tap.
- Tag creation is restricted by the [Release tags](https://github.com/danger/danger-js/rules/21251190) ruleset (already active), bypassable only by org admins — so that's the deploy gate.

### Notes

- npm auth uses [trusted publishing](https://docs.npmjs.com/trusted-publishers) via OIDC — there is no `NPM_TOKEN` secret to leak, and releases get a provenance attestation.
- **The `danger` package needs a trusted publisher configured on npmjs.com before the first tag push**, or the publish step will fail: org `danger`, repo `danger-js`, workflow `npm_publish.yml`, environment blank.
- The workflow asserts the tag matches `package.json` version before publishing.
- Tags here are unprefixed (`14.0.4`, not `v14.0.4`); `git.tagName` is now pinned in `.release-it.json` so that can't drift away from the workflow's trigger pattern.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/npm_publish.yml
+++ b/.github/workflows/npm_publish.yml
@@ -1,0 +1,48 @@
+name: Publish danger to npm
+
+# Pushing a version tag is the deploy trigger. Who can create tags is
+# restricted by the "Release tags" repository ruleset, so that ruleset's
+# bypass list is the list of people who can ship a release.
+on:
+  push:
+    tags:
+      - "[0-9]+.[0-9]+.[0-9]+"
+      - "[0-9]+.[0-9]+.[0-9]+-*"
+
+permissions:
+  contents: read
+  id-token: write # npm trusted publishing
+  actions: write # to kick off release.yml below
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/npm_publish.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22"
```
glensc, line 25: LLM's so love node-22, even the latest LTS is 24

PR #1535, opened 2026-08-27
Make node 24 builds on releases

### LINKED ISSUE #1527: [BUG]: 13.0.10 prebuilt binaries embed Node 18.5.0 and cannot run their own bundled undici 6 - every GitHub API fetch fails
### Describe the bug

The prebuilt `pkg` release binaries for 13.0.10 (`danger-macos-arm64.zip`, `danger-macos-x64.zip`, …) which is what `brew install danger/tap/danger-js` installs - embed **Node 18.5.0**.

Since #1514 (`feat: swapping our node-fetch for undici / native`, first shipped in 13.0.10) Danger depends on `undici@^6.27.0`. undici 6 calls `stream.finished()` on a WHATWG `ReadableStream` in `fetchFinale`, and Node's `stream.finished()` only accepts web streams from **v19.5.0** onward. (undici 6 itself declares `engines.node: >=18.17`.)

So on the prebuilt binaries every GitHub API response that has a body throws, and Danger can't fetch the PR at all:

```
Failed to fetch GitHub pull request files: TypeError: terminated
    at Fetch.onAborted (/snapshot/danger-js/node_modules/undici/lib/web/fetch/index.js:2046:49)
    at Fetch.emit (node:events:537:28)
    at Fetch.terminate (/snapshot/danger-js/node_modules/undici/lib/web/fetch/index.js:93:10)
    at /snapshot/danger-js/node_modules/undici/lib/web/fetch/index.js:511:30
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  [cause]: TypeError [ERR_INVALID_ARG_TYPE]: The "stream" argument must be an instance of Stream. Received an instance of ReadableStream
      at new NodeError (node:internal/errors:388:5)
      at eos (node:internal/streams/end-of-stream:65:11)
      at fetchFinale (/snapshot/danger-js/node_modules/undici/lib/web/fetch/index.js:1101:5)
      at mainFetch (/snapshot/danger-js/node_modules/undici/lib/web/fetch/index.js:765:5)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
    code: 'ERR_INVALID_ARG_TYPE'
  }
}
Failed to fetch pull request diff: TypeError: terminated
    …
```

`package.json` declares no `pkg.targets`, so the build falls back to `pkg`'s default base, and `pkg` never shipped a base newer than Node 18.5. `engines` is also still `{"node": ">=18"}`, which understates the actual requirement now that undici 6 is a dependency.

This is deterministic, not flaky — it fails on 100% of runs. It's currently breaking every Danger job on our CI as each machine picks up the Homebrew upgrade.

### Reproduction

```sh
curl -sSLO https://github.com/danger/danger-js/releases/download/13.0.10/danger-macos-arm64.zip
unzip -q danger-macos-arm64.zip && chmod +x danger-arm64
echo 'message("files: " + danger.git.modified_files.length)' > dangerfile.js
export DANGER_GITHUB_API_TOKEN=<token>

./danger-arm64 --version          # 13.0.10
./danger-arm64 pr https://github.com/<org>/<repo>/pull/<n>
#   -> Failed to fetch GitHub pull request files: TypeError: terminated
#   (the binary reports "Node.js v18.5.0" on any uncaught throw)
```

The same version installed from npm and run under Node 20 works fine:

```sh
npm i danger@13.0.10
./node_modules/.bin/danger pr https://github.com/<org>/<repo>/pull/<n>
#   -> Danger: ✓ passed, found only messages.
```

13.0.5 (the last `node-fetch` release) also works, so this is specific to the undici swap meeting `pkg`'s Node 18.5 base.

### Expected behaviour

The prebuilt binaries should run on a Node new enough for the dependencies they bundle. `pkg` is archived and can't produce a Node >= 19.5 base, so this probably needs one of:

- switching the release build to [`@yao-pkg/pkg`](https://github.com/yao-pkg/pkg) (maintained fork, supports Node 20/22/24) or Node SEA;
- pinning `undici` back to a release that works on Node 18.5;
- or, if neither is feasible soon, pulling the 13.0.10 macOS/Linux binaries so Homebrew users don't auto-upgrade into a broken install.

Bumping `engines.node` to `>=19.5` would also make the constraint explicit.

### Workaround for anyone hitting this

Don't use the prebuilt binary; install from npm and run it under Node >= 19.5. With danger-swift:

```sh
npm i danger@13.0.10
danger-swift ci --danger-js-path "$PWD/node_modules/.bin/danger"
```

Note it must point at the `danger` bin, not `danger-js` — invoking the `danger-js` bin name breaks commander's external subcommand lookup (`error: danger-js-ci(1) does not exist`), which is arguably a second small bug since danger-swift probes for `danger-js` first.

### Your Environment

| software | version |
| ---------------- | ------- |
| danger-js        | 13.0.10 (Homebrew `danger/tap/danger-js`, macOS arm64) |
| node (embedded)  | 18.5.0 — the binary's own runtime; the host's Node is not used |
| danger-swift     | 3.22.1 |
| OS               | macOS 15 / Darwin 25.6.0, Apple Silicon |

orta:
Swap the archived pkg for @yao-pkg/pkg so the prebuilt macOS binaries embed node 24 instead of node 18.5, which was too old for undici.

Fixes #1527


orta:
hrm, are you guys getting ping from the code owners? (not including your names because that could trigger it )

It should only happen on workflow changes, which is very rare

